PR #69, opened 2019-09-18
WIP: Introduce record and verify modes

### LINKED ISSUE #34: Support for updating and checking golden files
So far we’ve been thinking mostly about interactive development, but there are really 3 modes that Paparazzi can run in:

1. Development – emit screenshots for manual checking
2. Updating/creating golden files – generate screenshots intended for checking in
3. Checking golden files – confirm screenshots match golden files

I think the programmatic API should be the same in each mode: call snapshot().

But we should have mechanisms to create and check golden files to defend against regressions. Maybe a system property?
`-DpaparazziMode=dev`
`-DpaparazziMode=update`
`-DpaparazziMode=check`

The update mode should be opinionated about where golden files go. Perhaps something like this:

```
src/test/paparazzi/com/squareup/paparazzi/sample/KeypadViewTest.testViews.zero_dollars.png
src/test/paparazzi/com/squareup/paparazzi/sample/KeypadViewTest.testViews.five_bucks.png
src/test/paparazzi/com/squareup/paparazzi/sample/KeypadViewTest.testViews.grey.png
src/test/paparazzi/com/squareup/paparazzi/sample/KeypadViewTest.testViews.bolt.png
```

I’d love it if we didn’t actually need to write files to the filesystem in `check` mode. We already have a bitmap in memory.

When the check mode fails we should write a bitmap with actual image.

I’d also like it if we defer failure until after the test completes. That way we get all the failing screenshots.

jrodbx:
Closes #34


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -249,3 +254,16 @@ class Paparazzi(
...
+internal fun detectHandler(mode: Mode) =
+  when (mode) {
+    DEVELOPMENT -> HtmlReportWriter()
+    RECORD -> SnapshotWriter()
+    VERIFY -> SnapshotVerifier()
+  }
+
+enum class Mode {
+  DEVELOPMENT,
+  RECORD,
```
swankjesse: Craxy idea: can we combine DEVELOPMENT and RECORD modes? I think we can! All we need to do is emit the HTML files that you view to a different directory from the PNG files you commit

jrodbx:
I'll update based on feedback and push a new PR.

PR #73, opened 2019-09-20
Add modern device types

jrodbx:


paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
```diff
@@ -189,6 +189,150 @@ data class DeviceConfig(
...
+        ratio = ScreenRatio.LONG,
+        size = ScreenSize.NORMAL,
+        keyboard = Keyboard.NOKEY,
+        touchScreen = TouchScreen.FINGER,
+        keyboardState = KeyboardState.SOFT,
+        softButtons = true,
+        navigation = Navigation.NONAV
+    )
+
+    val PIXEL_3a = DeviceConfig(
```
swankjesse: `PIXEL_3A`

jrodbx:
> This is rad.
> 
> Would love a comment with a link to the source of truth for these specs.
> Also would be really handy to add the release date for everything, either in code or comments.

Added link to spec and release date member (as String).

PR #75, opened 2019-09-22
Update README with new URLs

fawaad:
Change from `https://square.github.io/paparazzi` to `https://cashapp.github.io/paparazzi`


jrodbx:
Thanks!

PR #84, opened 2019-09-29
#83 look at ANDROID_SDK_ROOT for Android platforms 

### LINKED ISSUE #83: Look at ANDROID_SDK_ROOT for Android SDK, then ANDROID_HOME as backup
`ANDROID_HOME` is [deprecated](https://developer.android.com/studio/command-line/variables), even though it's still widely used. Google now recommends using `ANDROID_SDK_ROOT` so it would be safe for us to handle that.

kirillzh:
Closes #83


jrodbx:
Thanks!

Issue #82, opened 2019-09-29
FileNotFoundException: /usr/local/share/android-sdk/platforms/android-Q/build.prop


kirillzh:
When running paparazzi getting this error:
```
   java.io.FileNotFoundException: /usr/local/share/android-sdk/platforms/android-Q/build.prop (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:195)
        at java.io.FileInputStream.<init>(FileInputStream.java:138)
        at app.cash.paparazzi.DeviceConfig$Companion.loadProperties(DeviceConfig.kt:202)
        at app.cash.paparazzi.internal.Renderer.prepare(Renderer.kt:76)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:106)
        at app.cash.paparazzi.Paparazzi$apply$1.evaluate(Paparazzi.kt:89)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:305)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:365)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:330)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:78)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:328)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:65)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:292)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:305)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:412)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
        at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
[...]

PR #85, opened 2019-09-29
#82 ignore empty Android platform directories

kirillzh:


paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -41,16 +41,14 @@ fun detectEnvironment(): Environment {
...
 
   return Environment(platformDir, userDir, resDir, packageName)
 }
 
 private fun checkInstalledJvm() {
-  val jvmVendor = System.getProperty("java.vendor")
-  val jvmVersion = System.getProperty("java.version")
-  if (jvmVendor == null || jvmVersion == null) return // we tried...
+  val jvmVendor = System.getProperty("java.vendor") ?: return
+  val jvmVersion = System.getProperty("java.version") ?: return
```
kirillzh: Unrelated to this PR but should we throw exception here? If there's no `java.vendor` or `java.version` are we still guaranteed that JVM is available?

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -34,12 +32,13 @@ fun detectEnvironment(): Environment {
...
-      .filter { Files.isDirectory(it) }
+  val platformDir = File("$androidHome/platforms")
+      .listFiles()
+      .orEmpty()
+      .filter { it.isDirectory }
       .map { it.toString() }
       .sorted()
-      .reduce { _, next -> next }
-      .orElse(null)
+      .last()
```
kirillzh: `sorted().last()` is basically `max()` but I don't think it's any more readable... 

Also, that's a very edge casy but, sorting platforms by string value doesn't seem reliable way of getting latest platform (I assume that's what is wanted here). If I had `android-9` installed (lol) it would use that as a latest platform... I'm wondering if we could use `sdkmanager` cli for a more deterministic approach but that would just slow things down. Alternatively we can maintain a map of known platform names and specify the order there but that would require maintaining it every year or so. As I said, that's an edge case and as long as there are no 9- platforms installed should work fine.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -34,12 +32,13 @@ fun detectEnvironment(): Environment {
   val userDir = System.getProperty("user.dir")
   val userHome = System.getProperty("user.home")
   val androidHome = System.getenv("ANDROID_HOME") ?: "$userHome/Library/Android/sdk"
-  val platformDir = Files.list(Paths.get("$androidHome/platforms"))
```
kirillzh: This returned Java Streams, I could replace them with Kotlin Sequences but I don't think we would gain anything from it considering the size of collection, so I just used plain collections API.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -16,8 +16,6 @@
 package app.cash.paparazzi
 
 import java.io.File
-import java.nio.file.Files
-import java.nio.file.Paths
```
kirillzh: 👋

kirillzh:
Resolved merge conflict, no additional changes.

jrodbx:
Closed in #82

PR #103, opened 2019-10-26
Update snapshot builds link in README

ericdiazjr21:


jrodbx:
CI build failure appears to be unrelated.  Thanks!

gabrielittner:
I don't think this change is a good idea. The old link was the one you put into your build file to be able to use snapshots. Now browsing them is a bit easier but anyone not knowing how repositories work will struggle using snapshots because they don't know they need to remove part of the link after copying it to their build file.

jrodbx:
Fair point, but I don't think the old way served as good Gradle documentation either as nothing mentioned "this is how you point to snapshot repos in your Gradle".  Plus, it was a clickable link, so people will naturally click and see "Directory listing forbidden", which is confusing.

How about https://github.com/cashapp/paparazzi/pull/104?

PR #114, opened 2019-12-21
Explicitly specify site-url for mkdocs

saket:
More details here: https://github.com/cashapp/sqldelight/pull/1533


jrodbx:
thanks!

Issue #109, opened 2019-11-15
Support AppCompat subclasses created in code


jrodbx:
Libraries like Contour promote creating views in XML vs code.  We've encountered a couple of issues around vector drawable resources and custom fonts with this approach, due to how AppCompat handles these during layout inflation vs direct view instantiation.

jrodbx:
Remaining work superseded by this more descriptive issue: https://github.com/cashapp/paparazzi/issues/119

PR #115, opened 2020-01-10
Force platform SDK version to test project compileSdkVersion

### LINKED ISSUE #91: Build.VERSION.SDK_INT takes a value of zero
Some logic that depends on `Build.VERSION.SDK_INT` will become abnormal because of it takes a value of zero. For example, `ContextCompat.getDrawable()` is a compatible method for getting Drawables in all Android versions, provided by Google in `support-compat` package.

872A672F-EE87-4C5F-8410-D0AB6D219F8A

As shown in the figure above, the program will enter the wrong branch when calling this method, which will cause `Drawable` resource to be parsed into a `String `resource, the `resolvedId` variable will eventually be 0, unable to get the correct `Drawable` object.

Is it possible to provide a way to set the runtime SDK version? In addition, it is better if project parameters such as `minSdkVersion` and `targetSdkVersion` can be obtained from the Gradle Plugin.

jrodbx:
Closes #91.  Part of #109.


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -56,4 +58,15 @@ open class PrepareResourcesTask : DefaultTask() {
         }
     throw IllegalStateException("No source sets available")
   }
+
+  private fun Project.targetSdkVersion(): String {
+    val androidExtension = extensions.getByType(BaseExtension::class.java)
+    return androidExtension.compileSdkVersion.substringAfter(
+        "android-", DEFAULT_COMPILE_SDK_VERSION.toString()
```
JakeWharton: This is not guaranteed to be the format and it's not guaranteed to contain a number. It can be any arbitrary string which is then looked up in the `platforms/` dir as a folder name.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -56,4 +58,15 @@ open class PrepareResourcesTask : DefaultTask() {
         }
     throw IllegalStateException("No source sets available")
   }
+
+  private fun Project.targetSdkVersion(): String {
```
JakeWharton: compileSdk. targetSdk is a different concept.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -46,8 +47,9 @@ fun detectEnvironment(): Environment {
   val configLines = File("build/intermediates/paparazzi/resources.txt").readLines()
   val packageName = configLines[0]
   val resDir = configLines[1]
+  val targetSdkVersion = configLines[2].toInt()
```
JakeWharton: Continuing from above, even ignoring custom SDKs and ignoring old SDKs from Google that contained Google APIs which had a different format, in the case of only official Android SDKs you will not always get an integer value. When the preview SDK for the next version of Android is made available it will be named `android-R`.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -256,6 +263,27 @@ class Paparazzi(
...
+    val modifiersField: Field = Field::class.java
+        .getDeclaredField("modifiers")
+        .apply {
+          isAccessible = true
+        }
+
+    val versionClass = try {
+      Paparazzi::class.java.classLoader.loadClass("android.os.Build\$VERSION")
+    } catch (e: ClassNotFoundException) {
+      return
```
JakeWharton: Rethrow as `AssertionError`? This should never happen.

jrodbx: happens in PaparazziTest

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
jrodbx: good catch.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
jrodbx: Originally, I was attempting to use manifest.targetSdk but when that returned null, i started using BaseExtension.getCompileSdkVersion whose underlying target member is set via one of the following methods:

```
    /** @see #getCompileSdkVersion() */
    public void compileSdkVersion(String version) {
        checkWritability();
        this.target = version;
    }

    /** @see #getCompileSdkVersion() */
    public void compileSdkVersion(int apiLevel) {
        compileSdkVersion("android-" + apiLevel);
    }

    public void setCompileSdkVersion(int apiLevel) {
        compileSdkVersion(apiLevel);
    }

    public void setCompileSdkVersion(String target) {
        compileSdkVersion(target);
    }
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: > Based on this, I think you're better off pulling the _actual_ target SDK version from the Gradle model which is always going to be an integer value.

how to grab this value?

jrodbx:
> Based on this, I think you're better off pulling the actual target SDK version from the Gradle model which is always going to be an integer value.

gonna tackle this under a separate issue.  right now, using:

```
project.plugins.withId("com.android.library") {
   android.libraryVariants.all {
      println(project.path + "/" + it.name + ": " + it.mergedFlavor.targetSdkVersion.apiLevel)
   }
}
```

returns null for `targetSdkVersion`.  I probably need to hook into a separate AGP task for this to be non-null.

Issue #120, opened 2020-03-05
Compile error with AGP 3.6


guelo:
With AGP 3.6 we get a build error with this stacktrace.

```
Caused by: java.lang.NoSuchMethodError: com.android.build.gradle.tasks.MergeResources.getOutputDir()Ljava/io/File;
	at app.cash.paparazzi.gradle.PrepareResourcesTask.writeResourcesFile(PrepareResourcesTask.kt:45)
```

jrodbx:
Closed in #121

PR #121, opened 2020-03-05
Fix AGP 3.6 compatibility

guelo:
This is the change necessary to fix issue #120. 

Google changed the MergeResource getOutputDir public API here https://android.googlesource.com/platform/tools/base/+/44c4c5a4aed32ba7403d1279ee7fec4b3ebe8b3f%5E%21/#F14

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'com.github.ben-manes.versions'
 buildscript {
   ext.versions = [
     kotlin: '1.3.61',
-    agp: '3.6.0',
+    agp: '3.6.1',
     layoutlib: '3.4.0', // "should" be similar to versions.agp
     androidTools: '26.5.0', // agp + 23.0.0 = androidTools
     jcodec: '0.2.5',
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'com.github.ben-manes.versions'
 buildscript {
   ext.versions = [
     kotlin: '1.3.61',
-    agp: '3.5.1',
+    agp: '3.6.0',
     layoutlib: '3.4.0', // "should" be similar to versions.agp
     androidTools: '26.5.0', // agp + 23.0.0 = androidTools
     jcodec: '0.2.5',
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -23,18 +23,17 @@ import org.gradle.api.Project
 import org.gradle.api.file.Directory
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.InputDirectory
-import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.Nested
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.TaskProvider
 
 open class PrepareResourcesTask : DefaultTask() {
   // Replace with @InputDirectory once mergeResourcesProvider.outputDir is of type Provider<File>.
+  @Internal
   internal lateinit var mergeResourcesProvider: TaskProvider<MergeResources>
-    @InputDirectory get
 
+  @Nested
   internal var outputDir: Provider<Directory> = project.objects.directoryProperty()
-    @OutputDirectory get
 
   @TaskAction
   fun writeResourcesFile() {
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -29,11 +29,12 @@ import org.gradle.api.tasks.TaskProvider
 
 open class PrepareResourcesTask : DefaultTask() {
   // Replace with @InputDirectory once mergeResourcesProvider.outputDir is of type Provider<File>.
-  @Internal
-  internal lateinit var mergeResourcesProvider: TaskProvider<MergeResources>
 
-  @Nested
+  internal lateinit var mergeResourcesProvider: TaskProvider<MergeResources>
+    @Internal get
+
   internal var outputDir: Provider<Directory> = project.objects.directoryProperty()
+    @Internal get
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: 'com.github.ben-manes.versions'
 buildscript {
   ext.versions = [
     kotlin: '1.3.61',
-    agp: '3.5.1',
+    agp: '3.6.0',
```
jrodbx: 3.6.1?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -42,7 +42,7 @@ open class PrepareResourcesTask : DefaultTask() {
         .use {
           it.write(project.packageName())
           it.newLine()
-          it.write(mergeResourcesProvider.get().outputDir.path)
+          it.write(mergeResourcesProvider.get().outputDir.get().asFile.path)
```
jrodbx: would hate to force all users to a minimum of AGP 3.6.  will need to think about this.  @AlecStrong any learnings from how SqlDelight manages breaking API changes across AGP versions?

autonomousapps: @jrodbx asked me to take a look, since I've also had to deal with breaking changes in API in my own plugin, which supports AGP 3.5.3, 3.6, and the latest 4.0 beta. While I have not solved the problem in a general way,* fortunately the changes have been small enough I've been able to create little workarounds in the two cases it has come up. For example, when AGP 3.6 changed the type of the `compile<variant>JavaWithJavac` task from `AndroidJavaCompile` to `JavaCompile` (from core Gradle), that was an issue. I dealt with that [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/master/src/main/kotlin/com/autonomousapps/internal/dependencyAnalysis.kt#L149-L155) by grabbing the task by name alone rather than name and type, and used the generalized `Task#getOutputs()` (`TaskOutputs`) API provided by Gradle to get access to its file outputs.

In another case, both the name of a task changed, _and_ one of its output types, and I chose to use reflection to get the strongly typed property (instead of using `TaskOutputs`). You can see that [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/master/src/main/kotlin/com/autonomousapps/internal/dependencyAnalysis.kt#L189) and [here](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/blob/master/src/main/kotlin/com/autonomousapps/internal/agpUtils.kt#L21).

You can get the current AGP version available at runtime by importing `com.android.builder.model.ANDROID_GRADLE_PLUGIN_VERSION`.

*I have spoken to a few people on the Gradle community slack who have solved this problem in a general way by having a complex build with many subprojects. There is an interface / adapter between the main plugin code and AGP, and the various subprojects each implement a different version of the interface, one per AGP version. The correct subproject is used at runtime. This is a completely general solution, but obviously a bit on the complex side. They did it because it was proprietary and they sold it as a product, and had customers on various versions of AGP.

guelo:
Trying to fix the tests on Travis I've run into two issues. AGP 3.6 is more strict about requiring `android.useAndroidX=true` in gradle.properties  and it's more strict about task property annotations. I've tried to come up with fixes but so far haven't been able to make it work.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
guelo: I added a reflection based solution in the last commit which should work for agp<3.5. Please take a look.

guelo:
Travis is stuck in a chicken and egg problem because :sample is referring to `apply plugin: 'app.cash.paparazzi'` which is in the maven repo instead of the local plugin. But the maven version doesn't work with AGP 3.6 which we need to compile my changes. Any ideas?

autonomousapps:
You should probably compile against the lowest version of AGP you want to
support. Also, I noticed that the plugin has AGP on the `implementation`
configuration, but it should probably be `compileOnly`. You can expect the
project where your plugin is applied to supply the dep at runtime.

jrodbx:
> but it should probably be `compileOnly`

shouldn't it be `implementation` since it depends on the `MergeResources` task?

jrodbx:
> AGP 3.6 is more strict about requiring android.useAndroidX=true in gradle.properties

I wonder if this is a bug rather than unexpected behavior, per this: https://issuetracker.google.com/issues/150655444.

jrodbx:
I'll take a closer look at the build issue in the morning.

autonomousapps:
>> but it should probably be compileOnly

> shouldn't it be `implementation` since it depends on the `MergeResources` task?

I usually make AGP `compileOnly` in my plugins because I expect it to be present at runtime -- provided by the user in their build. I've had too many issues with classpath conflicts. It could happen, for example, that someone has AGP in `buildSrc/build.gradle` as an `implementation` dependency, which will put it on the build classpath, and then they apply your plugin in the main build. Which version of AGP is actually there at runtime? It's the one supplied by buildSrc, not what you've bundled. So it just seems safer to not assume that what you're compiling against is what you actually run against.

(nb: this can also be important if, for a given plugin, AGP is actually optional. You wouldn't want to force your users to have it)

paparazzi-gradle-plugin/src/test/projects/verify-resources/gradle.properties
```diff
@@ -0,0 +1,2 @@
+VERSION_NAME=0.5.0-SNAPSHOT
+android.useAndroidX=true
```
jrodbx: this test suite doesn't require androidx dependencies, so we can omit this one.

jrodbx:
> Travis is stuck in a chicken and egg problem because :sample is referring to `apply plugin: 'app.cash.paparazzi'` which is in the maven repo instead of the local plugin. But the maven version doesn't work with AGP 3.6 which we need to compile my changes. Any ideas?

I'll disable the sample for now.  Long-term, I'm thinking we should probably use Gradle composite builds to keep :sample isolated to avoid this type of issue.

Can you update and rebase on top of https://github.com/cashapp/paparazzi/pull/123?

jrodbx:
thanks!

guelo:
great! Any idea when you're going to do the next release?

PR #125, opened 2020-05-27
Hook into testCompile tasks earlier to enable IDE runs

jrodbx:
Currently, the plugin hooks into `test${TEST_VARIANT}` to generate `build/intermediates/paparazzi/resources.txt` via `PrepareResourcesTask`.

When running tests from Android Studio, the Gradle task sequence appears to only invoke:
`compile${VARIANT}Sources` for all transitive dependency modules, and
`compile${TEST_VARIANT}Sources` for the module under test.

Since `test${TEST_VARIANT}` depends on the `compile*Sources` tasks and not the other way around, IDE test runs will avoid this task and result in test failures.

This fixes that behavior by hooking the plugin's `PrepareResourcesTask` into the earlier test build tasks.


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -47,10 +49,17 @@ class PaparazziPlugin : Plugin<Project> {
         it.dependsOn(variant.mergeResourcesProvider)
       }
 
-      project.tasks.named("test${variant.unitTestVariant.name.capitalize()}")
-          .configure {
-            it.dependsOn(writeResourcesTask)
-          }
+      val testVariantSlug = variant.unitTestVariant.name.capitalize()
```
JakeWharton, line 52: Pass `Locale.US`, although I believe the method is still experimental in 1.3.x

paparazzi-gradle-plugin/src/test/projects/verify-resources-java/gradle.properties
```diff
@@ -0,0 +1,2 @@
+VERSION_NAME=0.5.0-SNAPSHOT
```
JakeWharton, line 1: Probably not needed, right? Accidental copy/pasta?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: https://github.com/cashapp/paparazzi/pull/126

paparazzi-gradle-plugin/src/test/projects/verify-resources-java/gradle.properties
jrodbx: It "needs" it because I copied the test fixture stuff from Wire.  Can clean up later.

jrodbx:
Note to anyone who has been working with the snapshot releases.  In order for this change to take effect, you'll have to delete all metadata descriptors and artifacts from your local Gradle build caches, i.e., 

```
$ find $HOME/.gradle/caches -name app.cash.paparazzi | xargs rm -rf
```

PR #130, opened 2020-06-03
Hook AppCompatViewInflater into Paparazzi

jrodbx:
Enables the automatic swapping of FooView for AppCompatFooView in Paparazzi world and with that, recognition of attributes like `app:srcCompat`.

```
<LinearLayout
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_height="match_parent"
    android:background="@color/launchBackground"
    android:gravity="center"
    android:orientation="vertical"
    >

  <ImageView
      android:layout_width="200dp"
      android:layout_height="200dp"
      app:srcCompat="@drawable/arrow_up"
      />

</LinearLayout>
```

image


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -52,6 +53,7 @@ class Paparazzi(
   private val environment: Environment = detectEnvironment(),
   private val deviceConfig: DeviceConfig = DeviceConfig.NEXUS_5,
   private val theme: String = "android:Theme.Material.NoActionBar.Fullscreen",
+  private val disableAppCompat: Boolean = false,
```
JakeWharton, line 56: would prefer this was in the positive `appCompatEnabled: Boolean = true`

jrodbx: updated

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -284,6 +292,77 @@ class Paparazzi(
         }
   }
 
+  private fun initializeAppCompatIfPresent() {
+    lateinit var appCompatDelegateClass: Class<*>
+    try {
+      // See androidx.appcompat.widget.AppCompatDrawableManager#preload()
+      val appCompatDrawableManagerClass =
+        Class.forName("androidx.appcompat.widget.AppCompatDrawableManager")
```
gamepro65, line 300: Should it also check for support library too for projects that have yet to jetify / migrate to androidx?

PR #132, opened 2020-06-04
Use ByteBuddy to intercept usages of ResourceCompat.getFont

### LINKED ISSUE #119: Support custom fonts when constructing views in code
```
  // fails
  @Test
  fun textViewInCode() {
    val textView = AppCompatTextView(paparazzi.context).apply {
      typeface = ResourcesCompat.getFont(context, R.font.cashmarket_medium)
      text = "Hello, world!"
    }

    paparazzi.snapshot(textView, "text in code")
  }

  // passes
  @Test
  fun textViewInXml() {
    val textView = paparazzi.inflate<TextView>(R.layout.textview)
    paparazzi.snapshot(textView, "text in xml")
  }
```

related xml for the latter:
```
<TextView xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:fontFamily="@font/cashmarket_medium"
    android:text="Hello, world!"
    tools:ignore="HardcodedText"
    />
```

jrodbx:
Closes #119

Until https://issuetracker.google.com/issues/156065472 is fixed, specifically, font lookups via `ResourcesCompat.getFont` and, more generally, any calls to non-platform libraries like Androidx will fail, since they're not mocked out by LayoutLib.

This PR attempts to bridge that gap for the specific case by introducing a bytecode interceptor, routing calls from `androidx.core.content.res.ResourcesCompat.getFont(Context, int)` to `android.content.res.Resources.getFont(int id)`.  There may be a few edge cases missed, but for now, this appears to look fine.

Comparison of `ResourcesCompat.getFont` (androidx.core:core:1.1.0, left) to `Resources.getFont` (API 29, right):
Screen Shot 2020-06-04 at 12 57 20 AM

Same layout rendered in code (left) vs xml (right):
Screen Shot 2020-06-04 at 12 44 26 AM


jrodbx:
Seems like we're seeing another case where CI is generating a slightly different "actual" test image, which is failing the test.  Confirmed by copying over the raw image bytes from the failed test assertions, writing the following scratch file code:

```
import java.io.File
import java.nio.file.Files
import java.nio.file.Paths

val root = File("/Users/jrod/Development/paparazzi/paparazzi-gradle-plugin")
val expectedBytes = File(root, "expected_bytes")
val expectedByteArray = expectedBytes.readText().split(", ").map { it.toByte() }.toByteArray()
val actualBytes = File(root, "actual_bytes")
val actualByteArray = actualBytes.readText().split(", ").map { it.toByte() }.toByteArray()

Files.write(Paths.get("/Users/jrod/Development/paparazzi/paparazzi-gradle-plugin/expected.png"), expectedByteArray)
Files.write(Paths.get("/Users/jrod/Development/paparazzi/paparazzi-gradle-plugin/actual.png"), actualByteArray)
```

and comparing the two outputs using an image editor (Beyond Compare):

Screen Shot 2020-06-04 at 3 13 38 AM

Note the slightly faded font rendering in the bottom dark image...

swankjesse:
Ugh, different font rasterizers are super annoying.

paparazzi-agent/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'java-library'
+apply plugin: 'org.jetbrains.kotlin.jvm'
+
+dependencies {
+  implementation deps.kotlin.stdlib.jdk8
+
+  implementation 'net.bytebuddy:byte-buddy:1.10.10'
+  implementation 'net.bytebuddy:byte-buddy-agent:1.9.0'
```
JakeWharton, line 8: Move to dep map?

Also why different versions?

paparazzi-agent/src/main/java/app/cash/paparazzi/agent/AgentTestRule.kt
```diff
@@ -0,0 +1,23 @@
...
+    description: Description
+  ) = object : Statement() {
+    override fun evaluate() {
+      ByteBuddyAgent.install()
+      InterceptorRegistrar.registerMethodInterceptors()
+
+      try {
+        base.evaluate()
+      } finally {
+      }
```
JakeWharton, line 20: Remove

jrodbx: Not sure if you mean only the `finally` or the whole try/finally?

Someone could wrap this in a RuleChain and a preceding rule could throw, so I figure keeping both would make sense?

paparazzi-agent/build.gradle
jrodbx: Ah, forgot about this.  Good catch.

No good reason.  I think I copy-pasted from different sites and forgot to align 'em.

Both now fixed.

jrodbx:
Added ImageAssertions to allow image comparisons to succeed within a tolerance of <= 0.01%

paparazzi-agent/src/main/java/app/cash/paparazzi/agent/AgentTestRule.kt
jrodbx: Merging, will follow up on this.

JakeWharton: The code as-is does nothing. The whole try/finally can be removed.

jrodbx: removing the try-catch would result in never running the test….
```
class MyTest {
    @get:Rule
    val paparazzi =Paparazzi()  // magically adds AgentRule before itself

    @Test fun sometest() {  throw IllegalStateException("test ran!") }
}
```

with `base.evaluate`:
paparazzi.evaluate -> agentrule.evaluate -> invokemethod.evaluate which runs the test method (boom)

without `base.evaluate`:
paparazzi.evaluate -> agentrule.evaluate -> never runs test.  False success.

JakeWharton: Keep `base.evaluate()`. Just remove the try/finally. It catches no exceptions and runs no code on an exception being thrown.

jrodbx: https://github.com/cashapp/paparazzi/pull/133

PR #134, opened 2020-08-06
Publish repository details from properties

menny:
These changes will allow overriding all publishing-related properties from the command-line. In our case, we have an internal maven repository that we publish to.

### DIFF
--- a/gradle/gradle-mvn-push.gradle
+++ b/gradle/gradle-mvn-push.gradle
@@ -24,24 +24,6 @@ def isReleaseBuild() {
   return VERSION_NAME.contains("SNAPSHOT") == false
 }
 
-def getReleaseRepositoryUrl() {
-  return hasProperty('RELEASE_REPOSITORY_URL') ? RELEASE_REPOSITORY_URL :
-      "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
-}
-
-def getSnapshotRepositoryUrl() {
-  return hasProperty('SNAPSHOT_REPOSITORY_URL') ? SNAPSHOT_REPOSITORY_URL :
-      "https://oss.sonatype.org/content/repositories/snapshots/"
-}
-
-def getRepositoryUsername() {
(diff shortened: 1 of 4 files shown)


gradle/gradle-mvn-push.gradle
```diff
@@ -24,24 +24,6 @@ def isReleaseBuild() {
...
-  return hasProperty('SNAPSHOT_REPOSITORY_URL') ? SNAPSHOT_REPOSITORY_URL :
-      "https://oss.sonatype.org/content/repositories/snapshots/"
-}
-
-def getRepositoryUsername() {
-  return hasProperty('SONATYPE_NEXUS_USERNAME') ? SONATYPE_NEXUS_USERNAME : ""
-}
-
-def getRepositoryPassword() {
-  return hasProperty('SONATYPE_NEXUS_PASSWORD') ? SONATYPE_NEXUS_PASSWORD : ""
```
jrodbx, line 42: I've been using the following to publish to Square's internal repository for testing purposes: `./gradlew uploadArchives -PRELEASE_REPOSITORY_URL=(your repo) -PSONATYPE_NEXUS_USERNAME=(your username) -PSONATYPE_NEXUS_PASSWORD=(your password)`.  Is that not working for you?

jrodbx: If so, let's keep this as is and maybe add something to our README?

menny: these are the command-line arguments that work for us, too. But, with this change, we can also set which repository to use to fetch the internally built plugin and do that by applying a single patch on-top of HEAD (a commit that changes the `gradle.properties ` file.

jrodbx: Got it.  Yea, I see that ultimately, the main difference in this change is enabling the plugin to be pulled from a different repository.  We tend to publish and pull all dependencies from the same repository for consistency, so I think we'll leave this as is for now.

PR #135, opened 2020-08-06
Agnostic Test Inputs

menny:


.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
menny, line 1: Our default java environment is JDK 11. Adding this marker will automatically switch javac to 1.8 (which is required by AGP 3.6) is the developer is using `jenv`.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -44,12 +44,14 @@ fun detectEnvironment(): Environment {
       .reduce { _, next -> next }
       .orElse(null)
 
-  val configLines = File("build/intermediates/paparazzi/resources.txt").readLines()
-  val packageName = configLines[0]
-  val resDir = configLines[1]
-  val compileSdkVersion = configLines[2].toInt()
+  File(System.getProperty("PAPARAZZI_RESOURCES_DETAILS_FILE_KEY")).readLines()
```
menny, line 47: by taking the path from the outside, it allows us to use other `build-dir` values within Gradle and also different build systems.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -21,36 +21,36 @@ import com.android.Version
...
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.util.VersionNumber
 import java.io.File
 
 open class PrepareResourcesTask : DefaultTask() {
-  // Replace with @InputDirectory once mergeResourcesProvider.outputDir is of type Provider<File>.
 
   internal lateinit var mergeResourcesProvider: TaskProvider<MergeResources>
-    @Internal get
+    @Input get
```
menny, line 37: Marking this as input for caching and up-to-date (I did not add [CacheableTask](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/CacheableTask.html) annotation, though. I wasn't sure if this is something I wanted to do)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -62,6 +63,10 @@ class PaparazziPlugin : Plugin<Project> {
         project.tasks.named("compile${testVariantSlug}Kotlin")
             .configure { it.dependsOn(writeResourcesTask) }
       }
+
+      project.tasks.named("test${testVariantSlug}", Test::class.java).configure {
+        it.systemProperties["PAPARAZZI_RESOURCES_DETAILS_FILE_KEY"] = "${project.buildDir}/${paparazziResourcesDetailsFile}"
```
menny, line 68: Another option I was considering, is to just generate that `resources.txt` file in, well, as resources and have the Paparazzi rule know exactly where to look for that file. That option will remove the need for this `systemProperties` addition and will still work on other build-system.

paparazzi/build.gradle
```diff
@@ -53,6 +53,10 @@ def generateTestConfig = tasks.register("generateTestConfig") {
       writer.writeLine("app.cash.paparazzi")
       writer.writeLine(".")
       writer.writeLine("28")
+      def androidSdkFolder = System.getenv("ANDROID_SDK_ROOT")
```
menny, line 56: Since the project here does not have the Android plugin, I'm using the env variable. I think it should work, but it might not on some machines. Another option would be to try and read the value from `local.properties` file - I think AGP is automatically writing the path to that file.

.java-version
jrodbx: Cool, we don't use jenv, but it's a simple dot-file, so sgtm

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -36,21 +36,23 @@ open class PrepareResourcesTask : DefaultTask() {
   internal lateinit var mergeResourcesProvider: TaskProvider<MergeResources>
     @Internal get
 
-  internal var outputDir: Provider<Directory> = project.objects.directoryProperty()
+  internal var outputResourcesFile: Provider<RegularFile> = project.objects.fileProperty()
```
jrodbx, line 39: cool

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -29,27 +27,20 @@ data class Environment(
...
+  File(paparazziResourcesDetailsFile).readLines()
+          .run {
+            val packageName = this[0]
+            val resDir = this[1]
+            val compileSdkVersion = this[2].toInt()
+            val platformDir = this[3]
 
-  return Environment(platformDir, userDir, resDir, packageName, compileSdkVersion)
+            return Environment(platformDir, userDir, resDir, packageName, compileSdkVersion)
+          }
```
jrodbx, line 43: let's remove the style changes (run{}) and trim this down to the core changes?

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -29,27 +27,20 @@ data class Environment(
   val assetsDir = "$appTestDir/src/main/assets/"
 }
 
-fun detectEnvironment(): Environment {
+fun detectEnvironment(paparazziResourcesDetailsFile: String = System.getProperty("PAPARAZZI_RESOURCES_DETAILS_FILE_KEY")): Environment {
```
jrodbx, line 30: i like the system property.  does it need to be passed as an argument, vs looking it up similar to the properties below?  if not necessary, let's remove the argument.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -37,17 +39,16 @@ class PaparazziPlugin : Plugin<Project> {
...
     variants.all { variant ->
+      val paparazziResourcesDetailsFile = "intermediates/paparazzi/${variant.name}/resources.txt"
       val variantSlug = variant.name.capitalize(Locale.US)
-
       val writeResourcesTask = project.tasks.register(
           "preparePaparazzi${variantSlug}Resources", PrepareResourcesTask::class.java
       ) {
-        // TODO: variant-aware file path
-        it.outputs.file("${project.buildDir}/intermediates/paparazzi/resources.txt")
+        it.outputs.file("${project.buildDir}/${paparazziResourcesDetailsFile}")
```
jrodbx, line 47: is it possible to set this using the same layout.buildDirectory style below?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: it's a generated build/temp artifact.  moving it out would likely impose setting gitignores in consuming repos, which would be slightly annoying.

jrodbx (changes requested):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
menny: I didn't mean to generate it and store it under `src` folder (so git would see it). Basically, write that file into the `build` folder and add that to the resources. But, this is for future work, not in this PR.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
menny: I wonder if marking `outputResourcesFile` as `@OutputFile` is enough in this case. I'll do that.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
menny: I provided a default value to the argument, but I'm not sure it is possible to detect where it is located because of variant paths, build-system, and plugin version.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
menny: I don't mind reverting back, but don't you think this is a nice Kotlin readable code?

menny:
@jrodbx, this is ready for another review

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: re: array element access vs creating of another list + popping elements, the latter doesn't really buy much.

I like using extension functions, like apply, run, etc., but this[0] doesn't read as nice IMO.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -179,7 +179,7 @@ class PaparazziPluginTest {
 
     assertThat(result.task(":preparePaparazziDebugResources")).isNotNull()
 
-    val resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/resources.txt")
+    val resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
```
jrodbx, line 182: let's move variant-awareness to a separate PR

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: yea, using the system property is fine.  i'm suggesting to move it out of the argument list, and have it similar to the userDir system property below.

jrodbx (changes requested):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -70,6 +68,7 @@ class PaparazziPlugin : Plugin<Project> {
       val verifyTaskProvider = project.tasks.register("verifyPaparazzi${variantSlug}")
 
       val testTaskProvider = project.tasks.named("test${testVariantSlug}", Test::class.java) { test ->
+        test.systemProperties["PAPARAZZI_RESOURCES_DETAILS_FILE_KEY"] = "${project.buildDir}/${paparazziResourcesDetailsFile}"
```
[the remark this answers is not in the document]
jrodbx: keeps the system property naming consistent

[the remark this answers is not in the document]
jrodbx: also, can we keep this consistent with the call style on line 73?

i'm open to changing it to the operator [] access style in a separate PR.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
menny: In this case, I find it less error-prone to use `removeFirst()` rather than `list[0]`, `list[1]` etc: the chances you'll mess up the index are gone and the call order is identical to what is expected in the file.

menny: but, again, please let me know what you prefer here.

menny:
@jrodbx ready to go.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -66,6 +66,9 @@ class PaparazziPlugin : Plugin<Project> {
       val verifyTaskProvider = project.tasks.register("verifyPaparazzi${variantSlug}")
 
       val testTaskProvider = project.tasks.named("test${testVariantSlug}", Test::class.java) { test ->
+        test.systemProperty(
+                "paparazzi.file.resources-details",
+                "${project.buildDir}/${paparazziResourcesDetailsFile}")
```
jrodbx, line 71:
```suggestion
                writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -40,14 +40,14 @@ class PaparazziPlugin : Plugin<Project> {
     val variants = project.extensions.getByType(LibraryExtension::class.java)
         .libraryVariants
     variants.all { variant ->
+      val paparazziResourcesDetailsFile = "intermediates/paparazzi/resources.txt"
```
jrodbx, line 43: this will become the only usage, with the below suggestion, so this can be inlined.

paparazzi/build.gradle
jrodbx: makes sense, as is.

jrodbx, line 56: nit: `androidHome` for consistency.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: If i move the vals around in modifying code, now I've inadvertently broken behavior, which would not happen with array access.

I'd prefer to leave as is.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -29,27 +27,22 @@ data class Environment(
   val assetsDir = "$appTestDir/src/main/assets/"
 }
 
-fun androidHome() = System.getenv("ANDROID_SDK_ROOT")
-    ?: System.getenv("ANDROID_HOME")
-    ?: "${System.getProperty("user.home")}/Library/Android/sdk"
+fun paparazziResourcesDetailsFilePath(): String = System.getProperty("paparazzi.file.resources-details")
 
-fun detectEnvironment(): Environment {
+fun detectEnvironment(paparazziResourcesDetailsFile: String): Environment {
```
jrodbx, line 32: Per the earlier feedback, let's remove this argument and instead do something like:

```
  val resourcesFile = File(System.getProperty("paparazzi.test.resources"))
```

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -54,7 +54,7 @@ import java.util.Date
 import java.util.concurrent.TimeUnit
 
 class Paparazzi(
-  private val environment: Environment = detectEnvironment(),
+  private val environment: Environment = detectEnvironment(paparazziResourcesDetailsFilePath()),
```
jrodbx, line 57: revert, per above

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -29,7 +29,12 @@ import java.util.concurrent.TimeUnit
 
 class PaparazziTest {
   @get:Rule
-  val paparazzi = Paparazzi()
+  val paparazzi = Paparazzi(environment = testEnvironment())
+
+  private fun testEnvironment(): Environment {
+    //this file is generated by the task 'generateTestConfig'
+    return detectEnvironment(paparazziResourcesDetailsFile =  "build/intermediates/paparazzi/resources.txt")
+  }
```
jrodbx, line 37: revert, per above

paparazzi/build.gradle
```diff
@@ -53,6 +53,10 @@ def generateTestConfig = tasks.register("generateTestConfig") {
...
+              ?: System.getenv("ANDROID_HOME")
+              ?: "$userHome/Library/Android/sdk"
+      writer.writeLine("${androidSdkFolder}/platforms/android-28/")
     }
   }
 }
```
jrodbx, line 62: to setup, we can use a test setup similar to what the plugin would do:
```
tasks.withType(Test) { task ->
  task.dependsOn(generateTestConfig)
  task.systemProperty(
      "paparazzi.test.resources",
      generateTestConfig.map { it.outputs.files.singleFile }.get().path
  )
}
```

jrodbx (changes requested):

jrodbx:
Oh, and please sign our CLA: http://squ.re/sign-the-cla

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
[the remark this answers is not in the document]
menny: It's a different path, here I need to prefix with the build-folder.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
menny: I have a branch with this change. Once this PR is merged I'll open the other PR.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
menny: neat!

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
menny: having this as an argument will make testing easier: a method with clear inputs and outputs. I suggest to bring back the default value to the argument.

paparazzi/build.gradle
menny: I'm trying to figure a way to completely remove this from the gradle build file and have it contained in the test class itself. I'll do that in a separate PR, if that's fine.

menny:
ready for review, @jrodbx

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: I'd prefer `detectEnvironment` to remain parameter-less and instead users wishing to customize can instead create a custom Environment.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -66,6 +65,9 @@ class PaparazziPlugin : Plugin<Project> {
       val verifyTaskProvider = project.tasks.register("verifyPaparazzi${variantSlug}")
 
       val testTaskProvider = project.tasks.named("test${testVariantSlug}", Test::class.java) { test ->
+        test.systemProperty(
+                "paparazzi.test.resources",
+                writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path)
```
jrodbx, line 70: nit: formatting

jrodbx:
Addressed remaining feedback and merged in 04ead1549ff6fad322f3934589ec557dfcee791e.  Thanks!

PR #136, opened 2020-08-07
Binary resources

menny:


menny:
These are the generated reports for a sample app:
[paparazzi-app-reports.zip](https://github.com/cashapp/paparazzi/files/5075923/paparazzi-app-reports.zip)

PR #140, opened 2020-09-14
Introduce record and verify mode for image snapshots

jrodbx:
Wanted some quick feedback before I continue down this path.  Will add
tests before merging.

Unlike the first attempt (https://github.com/cashapp/paparazzi/pull/69/files),
toggling between test modes is done via JVM system properties vs updating sources.

I removed the default snapshot index count for 2 reasons:
1) Iterative updates to tests quickly cause indexing to lose meaning.
Better to force callers to provide a label when multiple snapshots exist
in a given test method.
2) Video snapshots will use a similar index naming convention.

Will follow up with video snapshots in a separate PR.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -21,10 +21,12 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.logging.LogLevel.LIFECYCLE
 import org.gradle.api.plugins.JavaBasePlugin
+import org.gradle.api.tasks.testing.Test
 import org.jetbrains.kotlin.gradle.plugin.KotlinBasePluginWrapper
 import java.io.File
 import java.util.Locale
 
+@Suppress("unused")
 class PaparazziPlugin : Plugin<Project> {
   @OptIn(ExperimentalStdlibApi::class)
   override fun apply(project: Project) {
@@ -65,13 +67,28 @@ class PaparazziPlugin : Plugin<Project> {
             .configure { it.dependsOn(writeResourcesTask) }
       }
 
-      project.tasks.named("test${testVariantSlug}")
-          .configure {
-            it.doLast {
-              val uri = File(project.buildDir, "reports/paparazzi/index.html").toURI()
-              project.logger.log(LIFECYCLE, "Click here for Paparazzi report: $uri")
-            }
-          }
+      val recordTaskProvider = project.tasks.register("recordPaparazzi${variantSlug}") {
+        System.setProperty("paparazzi.record", "true")
+      }
+
+      val verifyTaskProvider = project.tasks.register("verifyPaparazzi${variantSlug}") {
+        System.setProperty("paparazzi.verify", "true")
+      }
+
+      val testTaskProvider = project.tasks.named("test${testVariantSlug}", Test::class.java) {
+        it.systemProperty("paparazzi.test.record", System.getProperty("paparazzi.record") ?: "false")
+        it.systemProperty("paparazzi.test.verify", System.getProperty("paparazzi.verify") ?: "false")
+      }
+
+      recordTaskProvider.configure { it.dependsOn(testTaskProvider) }
+      verifyTaskProvider.configure { it.dependsOn(testTaskProvider) }
+
+      testTaskProvider.configure {
+        it.doLast {
+          val uri = File(project.buildDir, "reports/paparazzi/index.html").toURI()
+          project.logger.log(LIFECYCLE, "Click here for Paparazzi report: $uri")
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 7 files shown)


paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -88,20 +97,47 @@ internal class HtmlReportWriter(
       }
 
       override fun close() {
-        if (hashes.size == 1) {
-          shots += snapshot.copy(file = "images/${hashes[0]}.png")
+        val shot = if (hashes.size == 1) {
+          val original = File(imagesDirectory, "${hashes[0]}.png")
+          if (isRecording) {
```
jrodbx, line 102: i'm back and forth about whether to extract this out of HtmlReportWriter and into a, say, SnapshotRecorder, mainly because there'd be a bit of duplication.

See here for previous attempt: https://github.com/cashapp/paparazzi/pull/69/files#r326872374

paparazzi/src/main/java/app/cash/paparazzi/Snapshot.kt
```diff
@@ -17,13 +17,25 @@ package app.cash.paparazzi
 
 import com.squareup.moshi.JsonClass
 import java.util.Date
+import java.util.Locale
 
 @JsonClass(generateAdapter = true)
 data class Snapshot(
-  val name: String,
+  val name: String?,
```
jrodbx, line 24: currently, this means successive snapshot calls without a label will no-op on recording, due to the file existence check.  might be good to throw in this case.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -62,6 +64,22 @@ class PaparazziPlugin : Plugin<Project> {
...
+        System.setProperty("paparazzi.record", "true")
+      }
+
+      val verifyTaskProvider = project.tasks.register("verifyPaparazzi${variantSlug}") {
+        System.setProperty("paparazzi.verify", "true")
+      }
+
+      val testTaskProvider = project.tasks.named("test${testVariantSlug}", Test::class.java) {
+        it.systemProperty("paparazzi.test.record", System.getProperty("paparazzi.record") ?: "false")
+        it.systemProperty("paparazzi.test.verify", System.getProperty("paparazzi.verify") ?: "false")
```
jrodbx, line 78: System.setProperty lets the gradle task invoker set the property, but tests run in a separate forked JVM, so this pipes that setting through.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -399,5 +397,16 @@ class Paparazzi(
   companion object {
     /** The choreographer doesn't like 0 as a frame time, so start an hour later. */
     internal val TIME_OFFSET_NANOS = TimeUnit.HOURS.toNanos(1L)
+
+    private val isVerifying: Boolean by lazy {
```
JakeWharton, line 401: Probably don't need lazy

JakeWharton: Actually it's only invoked once below anyway. Could inline.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
JakeWharton: What about using `gradle.taskGraph.hasTask("record${slug}")` instead of using task configuration to side-effect setting a system property for this to be enabled.

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -59,13 +59,22 @@ import javax.imageio.ImageIO
...
   private val runsDirectory: File = File(rootDirectory, "runs")
   private val imagesDirectory: File = File(rootDirectory, "images")
   private val videosDirectory: File = File(rootDirectory, "videos")
+
+  private val goldenImagesDirectory = File(snapshotRootDirectory, "images")
+  private val goldenVideosDirectory = File(snapshotRootDirectory, "videos")
+
   private val shots = mutableListOf<Snapshot>()
 
+  private val isRecording: Boolean by lazy {
```
JakeWharton, line 74: Probably don't need lazy

JakeWharton (approved):

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: removed lazy

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
jrodbx: removed lazy

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Two thoughts on that:
1) I think peeking into the taskGraph is considered a smell by the Gradle team.
2) This would couple test run configuration to Gradle.  While it's our preferred build tool, it would be nice to enable extension/contributions from others for non-Gradle build systems.

PR #145, opened 2020-09-16
Add method interceptor for View.isInEditMode

jrodbx:
LayoutLib's View_Delegate currently intercepts [all calls to View.isInEditMode](https://cs.android.com/android/platform/superproject/+/master:frameworks/layoutlib/bridge/src/android/view/View_Delegate.java;l=37?q=View_Delegate) returning true.  While this default behavior makes sense for AndroidStudio's Layout Editor, which has historically not been able to render custom views very well, it gets in Paparazzi's way and requires authors to remove any isInEditMode checks that they would otherwise have wanted.

This provides a workaround, by intercepting the calls before they get intercepted by LayoutLib.

### DIFF
--- a/paparazzi-agent/src/main/java/app/cash/paparazzi/agent/AgentTestRule.kt
+++ b/paparazzi-agent/src/main/java/app/cash/paparazzi/agent/AgentTestRule.kt
@@ -13,8 +13,11 @@ class AgentTestRule : TestRule {
     override fun evaluate() {
       ByteBuddyAgent.install()
       InterceptorRegistrar.registerMethodInterceptors()
-
-      base.evaluate()
+      try {
+        base.evaluate()
+      } finally {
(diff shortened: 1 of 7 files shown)


paparazzi-agent/src/main/java/app/cash/paparazzi/agent/AgentTestRule.kt
```diff
@@ -15,6 +15,8 @@ class AgentTestRule : TestRule {
       InterceptorRegistrar.registerMethodInterceptors()
 
       base.evaluate()
```
JakeWharton, line 17: Need a try/finally around this so you always clear even if the test throws

JakeWharton (approved):

gamepro65 (approved):
LGTM

PR #148, opened 2020-09-17
Binary parsing module

menny:
Uses 'com.android.tools.apkparser:binary-resources'


menny:
@jrodbx  It will be better to review this after https://github.com/cashapp/paparazzi/pull/161 and https://github.com/cashapp/paparazzi/pull/160 are merged.

AlexBeggs:
if the conflicts are resolved is there a possibility to get this merged?

jrodbx:
> if the conflicts are resolved is there a possibility to get this merged?

My current thoughts are:
* we have to rewrite the resources loading logic soon, because more recent versions of layoutlib appear to have moved away from `com.android.ide.common.resources.deprecated.ResourceRepository` to `com.android.ide.common.resources.ResourceRepository`.  Last I looked into this, Android Studio looks at and parses the resources directly from the xml resource files, vs using the merged resource file Paparazzi currently depends on.  If I'm correct on this, the rewrite would then work for library AND application modules.  While I don't have a timeline for this, I _believe_ this also blocks us from upgrading to the version of layoutlib that's used in Android Studio 4.2 which would unblock Compose support, so there is some motivation to tackle this "soon".

* Depending on the APK packaging task to then decode the binary resources is slow, since you'd also have to wait for dexing when what you really want is resources.asrc.  Is there a way to obtain that prior to APK packaging?

* Up to now, the suggested workaround has been to move views under test to library modules.  Does that work for you?

jrodbx:
Obsoleted in favor of #524

PR #149, opened 2020-09-17
Fix race in record/verify modes

jrodbx:
Shame on me for not adding the tests I said I would in #140. This would have otherwise been caught before release.


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -66,17 +66,20 @@ class PaparazziPlugin : Plugin<Project> {
             .configure { it.dependsOn(writeResourcesTask) }
       }
 
-      val recordTaskProvider = project.tasks.register("recordPaparazzi${variantSlug}") {
-        System.setProperty("paparazzi.record", "true")
```
menny, line 70: I wonder if changing this to:
```
val recordTaskProvider = project.tasks.register("recordPaparazzi${variantSlug}") {
        doLast { System.setProperty("paparazzi.record", "true") }
}
```

menny: The difference is that in the original snippet was executed in the configuration phase (and both tasks had the potential be configured). With this suggestion, it will only execute if the task was executed.

jrodbx: The record task executes after the test task, so moving the closure to a task execution action would be deterministically missed by the test run which would depend on this system property.

JakeWharton (approved):

PR #152, opened 2020-09-17
Prepare version 0.5.2

jrodbx:


gradle.properties
```diff
@@ -1,5 +1,5 @@
 GROUP=app.cash.paparazzi
-VERSION_NAME=1.0.0-SNAPSHOT
```
jamesmosley55, line 2: Was this intentionally not set to 0.5.1?

jamesmosley55 (approved):

gradle.properties
jrodbx: Yea, it was set back to SNAPSHOT after deploying the release: https://github.com/cashapp/paparazzi/commit/93a089f1e1fb3ae55e4a6159f4595b744dd0d78a

PR #155, opened 2020-09-21
Add device configs for Pixel 4, 4XL, 4A

jrodbx:


JakeWharton (approved):

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -281,14 +279,12 @@ class PaparazziPluginTest {
         .runFixture(fixtureRoot) { build() }
 
     val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/images")
-    val snapshotFile = File(snapshotsDir, "be373e00b3ff3a319e40b9d798b1d2a1e40dec7d.png")
-    assertThat(snapshotFile.exists()).isTrue()
+    val snapshots = snapshotsDir.listFiles()
+    assertThat(snapshots!!).hasLength(1)
```
gamepro65, line 283: Should verifySnapshot_withoutFonts(), withoutAppCompat() and verifySnapshot() (even though its @Ignore) all be updated to this same logic for consistency even if they work as is?

jrodbx: I've been updating them only as CI failures occur due to the differences in rendering between the Travis/GitHub workers and my Mac OS laptop arises.

PR #170, opened 2020-10-08
Let RenderSessionImpl manage Choreographer_Delegate.doFrame

gamepro65:
We currently execute the doFrame before the view has been laid out. This means callbacks scheduled in the Choreographer expecting the view to be laid out won't have all the information available. By allowing RenderSessionImpl to call doFrame for us, the callbacks will execute _after_ the view is laid out.

For example, previously, an animation dependent on a view's bounds that is registered in onAttachedToWindow would not execute correctly.

### DIFF
--- a/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
@@ -19,10 +19,12 @@ import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.animation.ValueAnimator
 import android.graphics.Canvas
+import android.view.Choreographer
+import android.view.Choreographer.CALLBACK_ANIMATION
 import android.view.View
-import android.view.animation.AnimationUtils
 import android.view.animation.LinearInterpolator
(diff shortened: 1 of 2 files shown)


paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -19,10 +19,12 @@ import android.animation.Animator
...
 import android.animation.ValueAnimator
 import android.graphics.Canvas
+import android.view.Choreographer
+import android.view.Choreographer.CALLBACK_ANIMATION
 import android.view.View
-import android.view.animation.AnimationUtils
 import android.view.animation.LinearInterpolator
-import org.assertj.core.api.Assertions.assertThat
+import com.android.tools.layoutlib.java.System_Delegate
+import org.assertj.core.api.Java6Assertions.assertThat
```
jrodbx, line 27: wrong import?

jrodbx (approved):

jrodbx:
@swankjesse @JakeWharton thoughts?

JakeWharton (approved):
Not really. If the tests are passing then :shipit: !

PR #172, opened 2020-10-13
Enforce java 11 as min version

jrodbx:
Since our version of layoutlib now requires a min Java 11 JDK, let's enforce that early on.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
@@ -46,21 +46,18 @@ fun detectEnvironment(): Environment {
 }
 
 private fun checkInstalledJvm() {
-  val jvmVendor = System.getProperty("java.vendor")
-  val jvmVersion = System.getProperty("java.version")
-  if (jvmVendor == null || jvmVersion == null) return // we tried...
+  val feature = try {
+    // Runtime#version() only available as of Java 9.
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -46,21 +46,12 @@ fun detectEnvironment(): Environment {
...
-  if (jvmVendor.startsWith("Oracle") && major.toInt() == 1 && minor.toInt() <= 8) {
-    println(
-        """
-          |Unsupported JRE detected!!!
-          |
-          |Some custom fonts may not render correctly.  To avoid this, please install and run 
-          |Paparazzi test suites on OpenJDK version 8 or greater.
-          |See https://github.com/cashapp/paparazzi/issues/33 for additional context.
-          |""".trimMargin()
+  if (jvmVendor.startsWith("Oracle") || jvmVersion.feature() < 11) {
```
JakeWharton, line 52: Why does an Oracle JDK not work? It should be a superset of any OpenJDK build now.

jrodbx: Oracle JDK 8 had resulted in different image renderings due to underlying font metrics implementation; see [here](https://github.com/cashapp/paparazzi/pull/70/files).  But now that we're raising the minimum JDK level, that is likely fixed.

JakeWharton (approved):

PR #174, opened 2020-10-13
Allow selection of night mode in DeviceConfig.kt

dri94:
When testing different widgets we want to be able to test them in both NOT_NIGHT and NIGHT mode to ensure they appear correctly in both. We can simply add another parameter to enable this.


jrodbx (approved):
Thanks, please sign the CLA! http://squ.re/sign-the-cla

jrodbx:
Oh, just clicked on your avatar and saw that we're coworkers.  My bad!

PR #177, opened 2020-10-16
Add support for onGlobalLayout

gamepro65:
ViewRootImpl is responsible for scheduling / traversing view callbacks with the Choreographer. In order to support onGlobalLayout we must use ViewRootImpl#setView with RenderSession's rootView to allow the proper callbacks to be triggered.  The problem is that #setView interacts with DisplayManager and WindowSession which are not fully setup by layout lib. So we can use ByteBuddy to stub out the integration with those classes and provide sensible defaults to allow access to this method.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -17,12 +17,16 @@ package app.cash.paparazzi
 
 import android.content.Context
 import android.content.res.Resources
+import android.os.ServiceManager
 import android.util.AttributeSet
 import android.view.BridgeInflater
 import android.view.Choreographer_Delegate
+import android.view.IWindowManager
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.view.WindowManager.LayoutParams
+import android.view.animation.AnimationUtils
 import androidx.annotation.LayoutRes
 import app.cash.paparazzi.agent.AgentTestRule
 import app.cash.paparazzi.agent.InterceptorRegistrar
@@ -33,7 +37,9 @@ import app.cash.paparazzi.internal.PaparazziCallback
 import app.cash.paparazzi.internal.PaparazziLogger
 import app.cash.paparazzi.internal.Renderer
 import app.cash.paparazzi.internal.ResourcesInterceptor
+import app.cash.paparazzi.internal.ServiceManagerInterceptor
 import app.cash.paparazzi.internal.SessionParamsBuilder
+import app.cash.paparazzi.internal.WindowManagerInterceptor
 import com.android.ide.common.rendering.api.RenderSession
 import com.android.ide.common.rendering.api.Result
 import com.android.ide.common.rendering.api.SessionParams
@@ -101,6 +107,8 @@ class Paparazzi(
 
     registerFontLookupInterceptionIfResourceCompatDetected()
     registerViewEditModeInterception()
+    registerDisplayService()
+    registerWindowSession()
 
     val outerRule = AgentTestRule()
     return outerRule.apply(statement, description)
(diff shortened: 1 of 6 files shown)


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -272,12 +280,21 @@ class Paparazzi(
...
+        val viewRootImpl = viewGroup.viewRootImpl
+        val params = sessionParamsBuilder.build()
+        viewRootImpl.setView(viewGroup.findRoot(), LayoutParams(params.screenWidth, params.screenHeight, 0, 0, 0), null)
+      }
     } catch (e: Exception) {
       throw RuntimeException(e)
     }
   }
 
+  private fun View.findRoot(): View {
```
JakeWharton, line 294: super nit: +`tailrec`

You may have to pull the base case into its own conditional separate from the recursion case though

paparazzi/src/main/java/app/cash/paparazzi/internal/ServiceManagerInterceptor.kt
```diff
@@ -0,0 +1,32 @@
...
+package app.cash.paparazzi.internal
+
+import android.os.IBinder
+
+object ServiceManagerInterceptor {
+  @JvmStatic
+  fun intercept(
+    name: String
+  ): IBinder? {
+    if (name == "display") {
```
JakeWharton, line 26: could be a `when`, doesn't matter

JakeWharton (approved):
Wow this wiring is a pain in the ass. But if it works, 👍 .

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
gamepro65: 👍  Nice, ended up changing the compiled code into a while loop with no recursion

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziDisplayManager.kt
```diff
@@ -0,0 +1,184 @@
...
+import android.hardware.display.IDisplayManager
+import android.hardware.display.IDisplayManagerCallback
+import android.hardware.display.IVirtualDisplayCallback
+import android.hardware.display.WifiDisplayStatus
+import android.media.projection.IMediaProjection
+import android.view.DisplayInfo
+import android.view.Surface
+
+class PaparazziDisplayManager : IDisplayManager.Stub() {
+
```
jrodbx, line 32:
```suggestion
```

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziDisplayManager.kt
```diff
@@ -0,0 +1,184 @@
...
+  override fun getPreferredWideGamutColorSpaceId(): Int {
+    return 0
+  }
+
+  override fun registerCallback(p0: IDisplayManagerCallback?) {
+
```
jrodbx, line 38:
```suggestion
```

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziDisplayManager.kt
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  override fun setTemporaryAutoBrightnessAdjustment(p0: Float) {
+    throw UnsupportedOperationException("Minimum IDisplayManager methods are supported.")
+  }
+
+  override fun getMinimumBrightnessCurve(): Curve {
+    throw UnsupportedOperationException("Minimum IDisplayManager methods are supported.")
+  }
+
```
jrodbx, line 183:
```suggestion
```

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -124,6 +124,24 @@ class PaparazziTest {
...
+        viewTreeObserver.addOnGlobalLayoutListener {
+          onGlobalLayout = true
+        }
+      }
+    }
+
+    paparazzi.snapshot(view)
+
+    assertThat(onGlobalLayout).isTrue
+  }
```
jrodbx, line 143: nice

jrodbx (approved):

PR #184, opened 2020-12-22
Add render extension api to extend paparazzi

gpowell13:
The initial thought behind adding this api was to give devs a way to add metadata to the paparazzi snapshot.

This is the first pr to set up the plumbing for adding basic accessibility snapshots, similar to [Accessibility Snapshot](https://github.com/cashapp/AccessibilitySnapshot).

PR #185, opened 2020-12-22
Add accessibility render extension to render metadata for accessibility

geoff-powell:
Based on https://github.com/cashapp/paparazzi/pull/184

This pr builds on top of the render extension api to add simple rendering of accessibility information to view snapshots. This was initially inspired by [Accessibility Snapshot](https://github.com/cashapp/AccessibilitySnapshot) for ios.

## Example
- Sample Test - [link](https://github.com/cashapp/paparazzi/blob/380de34027f85f06146cbf92f3e74db2c5785532/paparazzi/extensions/accessibility/src/test/java/app/cash/paparazzi/extensions/accessibility/AccessibilityRenderExtensionTest.kt)
6ba3fb61f7b66371790b94604b4fe3ea70ab298e


paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
+apply plugin: 'java-library'
```
jrodbx, line 1: not needed?

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+apply plugin: 'kotlin'
```
jrodbx, line 2:
```suggestion
apply plugin: 'org.jetbrains.kotlin.jvm'
```

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+apply plugin: 'org.jetbrains.kotlin.kapt'
```
jrodbx, line 3: prob need to add `apply plugin: 'com.vanniktech.maven.publish'` now?

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+sourceCompatibility = '1.8'
+targetCompatibility = '1.8'
```
jrodbx, line 6: not needed?

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+      writer.writeLine(".")
+      writer.writeLine("29")
+      def androidHome = System.getenv("ANDROID_SDK_ROOT")
+          ?: System.getenv("ANDROID_HOME")
+          ?: "$userHome/Library/Android/sdk"
+      writer.writeLine("${androidHome}/platforms/android-29/")
+      writer.writeLine(".")
+    }
+  }
+}
```
jrodbx, line 33: not needed?

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+tasks.withType(Test).configureEach { task ->
+  task.dependsOn(generateTestConfig)
+  task.systemProperty(
+      "paparazzi.test.resources",
+      generateTestConfig.map { it.outputs.files.singleFile }.get().path
+  )
+}
```
jrodbx, line 41: not needed?

paparazzi/extensions/accessibility/src/main/java/app/cash/paparazzi/extensions/accessibility/RenderSettings.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
```
jrodbx, line 2: use current year in all new copyrights

paparazzi/extensions/accessibility/src/test/java/app/cash/paparazzi/R.kt
```diff
@@ -0,0 +1,4 @@
+package app.cash.paparazzi
+
+/** Simulate an empty R.java for this package. */
+class R
```
jrodbx, line 4: why is this needed?

paparazzi/extensions/accessibility/build.gradle
```diff
@@ -0,0 +1,41 @@
+apply plugin: 'java-library'
+apply plugin: 'kotlin'
+apply plugin: 'org.jetbrains.kotlin.kapt'
+
+sourceCompatibility = '1.8'
+targetCompatibility = '1.8'
+
+dependencies {
+  api project(':paparazzi')
```
jrodbx, line 9: what types does this need from paparazzi?  might be worth pulling into a commons?

settings.gradle
```diff
@@ -2,6 +2,7 @@ rootProject.name = 'paparazzi-parent'
 
 include ':sample'
 include ':paparazzi'
+include ':paparazzi:extensions:accessibility'
```
jrodbx, line 5: paparazzi-accessibility?

paparazzi/src/main/java/app/cash/paparazzi/RenderExtension.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+RenderExtension is an interface that can render additional information on top of the ui Paparazzi renders for each frame
+ */
+interface RenderExtension {
+  fun render(
+    snapshot: Snapshot,
+    view: View,
+    image: BufferedImage,
+  ): BufferedImage
+}
```
jrodbx, line 30: ah, is this the needed type?  maybe this is what goes into its own module then?

jrodbx (changes requested):

paparazzi/extensions/accessibility/src/test/java/app/cash/paparazzi/R.kt
geoff-powell: Its needed for the unit test

.github/workflows/build.yml
```diff
@@ -62,7 +62,9 @@ jobs:
           java-version: 11
 
       - name: Publish Artifacts
-        run: ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publish
+        run: |
+          ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publish
+          ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-accessibility:publish
```
geoff-powell, line 67: @jrodbx Not as familure with the maven publish plugin we are using but I added this to make sure the artifact is also published

paparazzi-accessibility/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+      writer.writeLine(".")
+      writer.writeLine("29")
+      writer.writeLine("platforms/android-29/")
+      writer.writeLine(".")
+      writer.writeLine(configurations.unzip.singleFile.path)
+    }
+  }
+}
+
+// TODO: Remove this duplicate of logic in paparazzi/build.gradle
```
jrodbx, line 55: Is depending on paparazzi module making this script boilerplate necessary?  If so, then we might have to go with the other modular structure after all?

geoff-powell: Nope this is due to the unit tests I have in the module. I could move them to one of the ones run in the `paparazzi-gradle-plugin` and then we don't need any of the logic since it runs against paparazzi directly via the plugin`

geoff-powell: So after testing this out it seems I'm going to need to invert the module dependency. I tried using the accessibility repo in the sample module and it was causing issues with the gradle plugin dependencies

jrodbx (approved):
nice!

geoff-powell:
The diff is pretty small. Upping the test to 0.2% max.
delta-accessibility

PR #186, opened 2021-01-04
Support opening files from assets

### LINKED ISSUE #179: Not possible to open assets
**Environment:**
- Paparazzi: 0.6.0
- OpenJDK: 11.0.9
- Mac OS: 10.15.6

**Steps:**
- `assertNotNull(paparazzi.context.assets.open("asset_name"))` fails

**Description:**

Right now I'm using `detectEnvironment().assetsDir` (because `Paparazzi.environment` is private) to get the assets directory.

omainegra:
Closes #179

A common use case is to load images from assets (using `file:///android_asset/...` URI) during tests while pointing to real images URL in prod


jrodbx:
According to https://stackoverflow.com/a/45908665, `file:///android_asset/...` is only accessible in WebViews, which isn't supported by Paparazzi.  However, there is a bug/crash when accessing asset files.

omainegra:
Most image loaders support loading images from assets using this URI (at least Glide, Picasso, Coil). My use case is to use a Coil's test `ImageLoader` during snapshots, being able to record images too.

omainegra:
Even tho, I saw you fixed the bug in another PR 👍

PR #189, opened 2021-01-20
Setup max percent difference as a parameter

### LINKED ISSUE #188: Max percent difference causes false positives
**Environment:**
- Paparazzi: 0.6.0
- OpenJDK: 11.0.9
- Mac OS: 10.15.6

**Description:**

The `MAX_PERCENT_DIFFERENCE` value may be too high (currently `0.1`) and small changes in big snapshots (like a lighter `textColor`) don't fail the test. 

Also, this constant could be a parameter of the Paparazzi configuration, falling back to the current value as default:

```
@get:Rule
internal var paparazzi = Paparazzi(maxPercentDifference = 0.01)
```

**Steps:**

1. **Given the golden image**
ca skipthedishes customer ui_MenuItemOptionRadioGroupComponentTests_screenshot_ThemeOrange_idle

2. **And the new change**

ca skipthedishes customer ui_MenuItemOptionRadioGroupComponentTests_screenshot_ThemeOrange_idle

3. With `MAX_PERCENT_DIFFERENCE = 0.1` the test still pass

4. With `MAX_PERCENT_DIFFERENCE = 0.01` the test fails with the expected delta

delta-ca skipthedishes customer ui_MenuItemOptionRadioGroupComponentTests_screenshot_ThemeOrange_idle

omainegra:
Replace `MAX_PERCENT_DIFFERENCE` constant with `maxPercentDifference` parameter.

Fixes #188

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
@@ -86,7 +86,7 @@ internal object ImageUtils {
             relativePath,
             goldenImage,
             thumbnail,
-            maxPercentDifferent
+            maxPercentDifference
         )
       } finally {
         `is`.close()
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
@@ -49,8 +49,6 @@ internal object ImageUtils {
 
   private val THUMBNAIL_SIZE = 1000
 
-  val MAX_PERCENT_DIFFERENCE = 0.1
-
   /** Directory where to write the thumbnails and deltas. */
   private val failureDir: File
     get() {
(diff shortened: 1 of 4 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -63,7 +61,8 @@ internal object ImageUtils {
   @Throws(IOException::class)
   fun requireSimilar(
     relativePath: String,
-    image: BufferedImage
+    image: BufferedImage,
+    maxPercentDifferent: Double
```
jrodbx, line 65:
```suggestion
    maxPercentDifference: Double
```

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -87,7 +86,7 @@ internal object ImageUtils {
             relativePath,
             goldenImage,
             thumbnail,
-            MAX_PERCENT_DIFFERENCE
+            maxPercentDifferent
```
jrodbx, line 89:
```suggestion
            maxPercentDifference
```

jrodbx (approved):

PR #190, opened 2021-01-29
Do not require valid test class to get test name

mattprecious:
This enables tests using Burst, which generates test class names like
TestClass[PARAMETER].

Screen Shot 2021-01-29 at 2 26 05 PM


mattprecious:
Uhhh, so CI is trying to run the burst test which crashes on the production version of paparazzi. Is this a useful CI job if it's not pointing towards the in-repo version?

mattprecious:
I guess if the sample doesn't compile or run tests then that's bad. I've ignored the test for now!

jrodbx (approved):

PR #206, opened 2021-02-26
Force test re-runs if generated report or snapshot dirs are deleted

### LINKED ISSUE #178: Gradle skips Paparazzi tasks with no actions
**Environment:**
- Paparazzi: 0.6.0
- OpenJDK: 11.0.8
- Mac OS: 10.15.4

**Steps:**
- Clone [paparazzi repository](https://github.com/cashapp/paparazzi)
- Collect snapshots: `./gradlew sample:recordPaparazziDebug --info`
- Verify snapshots: `./gradlew sample:verifyPaparazziDebug --info`
- Remove snapshot images
- Collect snapshots: `./gradlew sample:recordPaparazziDebug --info`

**Actual result:**
- Gradle skips paparazzi task:
```
Skipping task ':sample:recordPaparazziDebug' as it has no actions
```

**Additional info**:
It's useful to note that this is not a permanent bug for me and it can be tricky to reproduce. Execution of paparazzi gradle task with an `--rerun-tasks` argument is kind of a workaround: `./gradlew sample:recordPaparazziDebug --rerun-tasks --info`. But I would like to keep gradle cache and avoid using this argument if it's possible. Thanks!

jrodbx:
Similar to #204, but with task outputs.  Closes #178.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -48,6 +48,8 @@ class PaparazziPlugin : Plugin<Project> {
 
       val mergeResourcesOutputDir = variant.mergeResourcesProvider.flatMap { it.outputDir }
       val mergeAssetsOutputDir = variant.mergeAssetsProvider.flatMap { it.outputDir }
+      val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
+      val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
 
       val writeResourcesTask = project.tasks.register(
           "preparePaparazzi${variantSlug}Resources", PrepareResourcesTask::class.java
@@ -80,6 +82,8 @@ class PaparazziPlugin : Plugin<Project> {
 
         test.inputs.dir(mergeResourcesOutputDir)
         test.inputs.dir(mergeAssetsOutputDir)
+        test.outputs.dir(reportOutputDir)
+        test.outputs.dir(snapshotOutputDir)
 
         test.doFirst {
           test.systemProperties["paparazzi.test.record"] =
@@ -97,7 +101,7 @@ class PaparazziPlugin : Plugin<Project> {
 
       testTaskProvider.configure {
         it.doLast {
-          val uri = project.buildDir.toPath().resolve("reports/paparazzi/index.html").toUri()
+          val uri = reportOutputDir.get().asFile.toPath().resolve("index.html").toUri()
           project.logger.log(LIFECYCLE, "See the Paparazzi report at: $uri")
         }
       }
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -254,6 +254,83 @@ class PaparazziPluginTest {
   }
 
   @Test
+  fun rerunOnReportDeletion() {
+    val fixtureRoot = File("src/test/projects/rerun-report")
+
+    val reportDir = File(fixtureRoot, "build/reports/paparazzi")
+    val reportHtml = File(reportDir, "index.html")
+    assertThat(reportHtml.exists()).isFalse()
+
+    // Take 1
+    val firstRunResult = gradleRunner
+        .withArguments("recordPaparazziDebug", "--stacktrace")
+        .forwardOutput()
+        .runFixture(fixtureRoot) { build() }
+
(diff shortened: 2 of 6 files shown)


saket (approved):

alashow:
After running `./gradlew testDebugUnitTest` and then running `./gradlew verifyPaparazziDebug` doesn't do anything, by saying tasks all tasks are uptodate. This was happening every time on CI and only sometimes locally. After seeing this PR . linked issues, I added `sh 'shopt -s globstar; rm -rf **/build/reports/paparazzi'` step after running `testDebugUnitTest` and before `verifyPaparazziDebug`, and now it works properly.
It's working but I find having to remove previous reports a little hacky and cleaning the cache or using --rerun-tasks causes longer builds.

Is there anyway to improve this caching behavior?

TWiStErRob:
potential other workaround:
In recent Gradle releases there's [--rerun-task verifyPaparazziDebug`](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:builtin_task_options), similar to `-x`. This could be used instead of manually cleaning.

alashow:
@TWiStErRob You linked to [`--rerun`](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:builtin_task_options), which doesn't seem to work. But I had already tried [`--rerun-tasks`](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:rerun_tasks), it works same as using `clean verifyPaparazziDebug`, which I'm trying to avoid:
> This will force test and all task dependencies of test to execute. It’s a little like running gradle clean test, but without the build’s generated output being deleted.

TWiStErRob:
Oh, you are right about the rerun option, [it's not like `-x` 🤦‍♂️](https://docs.gradle.org/7.6/release-notes.html#individual-rerun). Note that there's also a clean* task for every task. So `gradlew cleanTestDebugUnitTest verifyPaparazziDebug` should also rerun.

Can you please report a new issue with a full repro? Including version numbers. According to the code there's a property added to the test task, that should trigger a rerun. Do you use Gradle predictive test selection or similar?

alashow:
>  According to the code there's a property added to the test task, that should trigger a rerun.

What property is that?

> Do you use Gradle predictive test selection or similar?

No

I wasn't able to reproduce this properly yet. Deleting reports trick doesn't even work with paparazzi/sample. It just re-creates the report & but tests are not run

For example:
1. Clone this repo
2. Change some test to fail when verifying. For ex, `val text = "Hello, Paparazzi"` -> `val text = "Hello, Paparazzi World"`
3. Run `./gradlew sample:testDebugUnitTest` - it will create the reports with "Hello, Paparazzi World"
4. Then running `./gradlew sample:verifyPaparazziDebug` doesn't do anything with all up-to-date tasks.

Deleting the report doesn't work in this case - `verifyPaparazziDebug` just "recreates" the report (doesn't actually print "See the Paparazzi report at:" but the report files are re-created or copied from somewhere maybe because the task runs very fast, 2 seconds )

TWiStErRob:
> It just re-creates the report & but tests are not run
> verifyPaparazziDebug just "recreates" the report

That's probably because caching is enabled:
https://github.com/cashapp/paparazzi/blob/17e35388c1b369c7b70294425c31801037c2ccb1/gradle.properties#L1

> According to the code there's a property added to the test task, that should trigger a rerun.

https://github.com/cashapp/paparazzi/blob/17e35388c1b369c7b70294425c31801037c2ccb1/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L173-L174

Which means `testDebugUnitTest` should have those values `false`, and `verifyPaparazziDebug` should have those values `true`, which would normally invalidate the cache and cause a re-run.

> Deleting the report doesn't work in this case - verifyPaparazziDebug just "recreates" the report

Also, here, you must be aware that `verifyPaparazziDebug` doesn't actually do anything other than configuring the test task. You can see this by adding `--console=verbose` to `gradlew`:
```
> Task :sample:testDebugUnitTest UP-TO-DATE
> Task :sample:verifyPaparazziDebug UP-TO-DATE
```
and just as I was writing this, I realized the problem...
the property setting above is added via `test.doFirst(object : Action<Task> {` which runs AFTER the caching key (including UP-TO-DATE check) is calculated, so it would affect the execution, if the incremental build calculations allowed it.

All the systemProperties manipulation is eager, and the `isRecordRun` value is only set after `afterEvaluate` completes, so there's no chance to do `afterEvaluate { test.systemProperties...`. This is why it was necessary to do it in `doFirst`, but this is incorrect and exhibits your problem. It needs to be made lazy in a different way.

I have a feeling as quick solution we might be able to add:
```
test.inputs.property("paparazzi.test.record", isRecordRun)
test.inputs.property("paparazzi.test.verify", isVerifyRun)
```
so that when it comes to UP-TO-DATE check, it has the providers as dependencies.

TWiStErRob:
Created unit test and fix based on above findings https://github.com/cashapp/paparazzi/pull/690

PR #229, opened 2021-04-29
Fix link for sample in README

tasomaniac:
Fix the link for sample in the documentation website by giving a full link. 
The relative link does not work on the website and gives 404.


jrodbx:
thanks!

PR #236, opened 2021-07-28
Add paparazzi tasks to Gradle task verification group

### LINKED ISSUE #227: Put Paparazzi tasks into Verification group
In Gradle there is a way to organize tasks in groups, and there are default groups such as `build`, `install` etc.
And there is a `verification` group for tasks designed to do some linting and tests. 
Paparazzi's tasks seem to be ideal candidates to be put into `verification` group. 
WDYT?

jrodbx:
Closes #227


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -119,4 +123,8 @@ class PaparazziPlugin : Plugin<Project> {
       return this
     }
   }
+
+  internal companion object {
+    const val VERIFICATION_GROUP = "verification"
```
JakeWharton, line 128: This is available at `org.gradle.language.base.plugins.LifecycleBasePlugin.VERIFICATION_GROUP`.

JakeWharton (approved):

PR #237, opened 2021-08-01
Deploy website docs using Github Action

jrodbx:


RELEASING.md
```diff
@@ -11,8 +11,7 @@ Releasing
  8. Update the `gradle.properties` to the next SNAPSHOT version.
  9. `git commit -am "Prepare next development version"`
  10. `git push && git push --tags`
- 11. `./deploy_website.sh`
```
JakeWharton, line 14: Delete the script?

.github/workflows/build.yml
```diff
@@ -66,3 +66,11 @@ jobs:
         env:
           ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
           ORG_GRADLE_PROJECT_SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+
+      - name: Deploy docs to website
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages
+          FOLDER: build/dokka/html
```
JakeWharton, line 75: I suppose `uploadArchives` creates this folder?

JakeWharton (approved):

.github/workflows/build.yml
jrodbx: hmm, doesn't look like it.  I just cribbed this from `https://github.com/cashapp/turbine/blob/trunk/.github/workflows/build.yaml#L50`

jrodbx: Ah, my guess is that, in build.yaml, either `build` or `publish` (prob the former) executes the dokka task which creates that folder?

jrodbx: And in release.yaml, `dokkaHtml` does it?  https://github.com/cashapp/turbine/blob/trunk/.github/workflows/release.yaml#L64-L65

jrodbx: hmm, i should also update the dokka plugin...

jrodbx: Ok, had to re-learn the dokka config 😅 

This repo doesn't use the html style that Turbine uses, but rather the gfm style that SqlDelight et al. use.

PR #246, opened 2021-08-20
Updating README.md with new Jetifier workaround property.

### LINKED ISSUE #245: android.jetifier.blacklist deprecated and leading to AmbiguousStringJetifierException
`blacklist` has been deprecated in favor of `ignorelist` and will be removed in v7.0 of the AGP. It also seems to be causing the following error in my build: 

 ```
Reason: AmbiguousStringJetifierException, message: 
The given artifact contains a string literal with a package reference 'android.support.design.widget' that
cannot be safely rewritten. Libraries using reflection such as annotation processors need to be updated 
manually to add support for androidx.. 
```

Build completes successfully when using
`android.jetifier.ignorelist=android-base-common,common`.

Versions: [JDK:11.0.9, AGP:4.2.2, Kotlin:1.5.21, Gradle:7.1.1]

luis-cortes:
Resolves #245

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ If using Jetifier to migrate off Support libraries, add the following to your `g
 exclude bundled Android dependencies.
 
 ```groovy
-android.jetifier.blacklist=android-base-common,common
+android.jetifier.ignorelist=android-base-common,common
 ```
 
 Releases


jrodbx (approved):

jrodbx:
Thanks!

PR #251, opened 2021-09-02
Migrate Paparazzi to use native layoutlib

jrodbx:
This migrates from the layoutlib bundled in Android Studio 4.0.x to the one bundled in 4.1.x.

Add new native runtime artifacts to support platform-specific runtime for layoutlib
As a result, Bridge.init now needs non-null icuPath and nativeLibPath.
Uses Gradle artifact transforms to download and explode runtimes, icu and font data.


libs/native-win/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=native-win
```
jrodbx, line 1: open to suggestion on artifact id @JakeWharton

geoff-powell (approved):

libs/native-win/gradle.properties
JakeWharton: I would probably do whatever aapt2 does for its native artifact

JakeWharton (approved):

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -24,15 +24,17 @@ data class Environment(
...
+)
 
 @Suppress("unused")
 fun androidHome() = System.getenv("ANDROID_SDK_ROOT")
     ?: System.getenv("ANDROID_HOME")
     ?: "${System.getProperty("user.home")}/Library/Android/sdk"
 
 fun detectEnvironment(): Environment {
-  checkInstalledJvm()
+//  checkInstalledJvm()
```
luis-cortes, line 37: Nit: delete?

libs/native-win/gradle.properties
jrodbx: I'm not sure what the suggestion based on this: 
`/Users/jrod/.gradle/caches//modules-2/files-2.1/com.android.tools.build/aapt2/4.1.2-6503028/9b3759a6b73cf339d81c99794fb99c402da9a00c/aapt2-4.1.2-6503028-osx.jar`

Can you clarify?

jrodbx: `layoutlib-native-win`?

geoff-powell: I like `layoutlib-native-win`. It's a bit more descriptive.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: oops, good call out.  reverted, since that check should still occur!

libs/native-win/gradle.properties
jrodbx: updated!

PR #256, opened 2021-09-07
Add support non transitive resources.

luis-cortes:
The [android.nonTransitiveRClass](https://developer.android.com/studio/releases/gradle-plugin#4.1-nontransitive-r-class) flag enables namespacing of each library's R class so that its R class includes only the resources declared in the library itself and none from the library's dependencies, thereby reducing the size of the R class for that library.

Paparazzi does not currently support loading resources outside of `packageName.R`, which results in `ResourceNotFoundException`s when client code references resources from dependencies by ID. 

This PR adds support for transitive resources by using AGP internals to build a list of all packages containing resources and loading each `package.R` class. 

If `android.nonTransitiveRClass=false`, the behavior falls back to the original implementation.


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -65,6 +65,8 @@ open class PrepareResourcesTask : DefaultTask() {
           it.newLine()
           it.write(platformDataRoot.get().asFile.path)
           it.newLine()
+          it.write("build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/debugUnitTest")
```
luis-cortes, line 68: I searched for a property that would give me this but didn't find anything. Hopefully there's a better way.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -727,6 +746,7 @@ class PaparazziPluginTest {
     if (!gradleProperties.exists()) {
       gradleProperties.createNewFile()
       gradleProperties.writeText("android.useAndroidX=true")
+      gradleProperties.writeText("android.nonTransitiveRClass=true")
```
luis-cortes, line 749: Initially I added a `gradle.properties` to the test project, but figured there wasn't any harm in just adding it here. Let me know which you prefer.

jrodbx: This will add that property to every test, vs having one test that highlights the specific change.

luis-cortes: Alrighty. I'll add the `gradle.properties` to the individual project. 👍

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
autonomousapps: I'm not entirely sure how this task works, but hard-coding a path like this is, generally, not safe. For one thing, note the terminal element, `debugUnitTest`. Does that mean this task only works on debug build? What about flavors? Other build types?

jrodbx: I have some Gradle snippet code that i typically use to analyze/print task inputs and outputs over which i would then grep, to find the appropriate task output dir to which I'd then wire the plugin task.

Afk at the moment, but can paste what I mean later!

luis-cortes: @autonomousapps Agreed. Don't want to merge with this just trying to get the conversation started because I wasn't able to find an API that gave me what we want here. Will keep digging while I wait on @jrodbx's snippet.

luis-cortes: I managed to piece together the path using an object that I found looking through AGP's source code and the unit test variant name. The only thing that I'm not excited about is how the object that I'm using is in an internal package. 
Probably not a great idea to depend on it. 🙁

autonomousapps: I think the new snippet is closer, but what you really want is to depend on that task and feed its outputs into this task's inputs. In fact, you kind of have to, because I think this would fail in Gradle 8 -- they're getting more strict. On the plus side, if I'm right, if you run this code with Gradle 7, then it will tell you which task you're stepping on and that would be a big hint what to use.

jrodbx: Ok, in a past talk, I offered some hacks/debugging tips to introspect what tasks get added by AGP (or any plugin, really):
https://speakerdeck.com/jrodbx/deep-er-dive-into-the-android-gradle-plugin-droidcon-sf-2018?slide=74
https://speakerdeck.com/jrodbx/deep-er-dive-into-the-android-gradle-plugin-droidcon-sf-2018?slide=76

By adding this to `paparazzi/build.gradle`:
```
gradle.taskGraph.whenReady {
  gradle.taskGraph.allTasks.each { task ->
    println "task: " + task.name + ", class: " + task.class.name
  }
}
gradle.taskGraph.afterTask { task ->
  println "  inputs: "
  task.inputs.files.each { println "    $it" }
  println "  outputs: "
  task.outputs.files.each { println "    $it" }
}
```
and running `./gradlew sample:test --console=plain`, 
then searching for `compile_and_runtime_not_namespaced_r_class_jar` (given how unique that segment is) and finding the task that outputs something at that path:

```
> Task :sample:generateDebugUnitTestStubRFile
  ...
  outputs:
    /Users/jrod/Development/paparazzi/sample/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/debugUnitTest/R.jar
```

then searching for `generateDebugUnitTestStubRFile`, you'll find the task class itself (ignoring the _Decorated part, that's a Gradle thing)!:
```
task: generateDebugUnitTestStubRFile, class: com.android.build.gradle.internal.res.GenerateLibraryRFileTask_Decorated
```

Alternatively, if you're familiar with AGP internals, the variant is "DebugUnitTest" so the task name is usually a concatenation of the variant with a prefix and/or suffix, i.e., "generate${VARIANT}StubRFile".

So either: 
https://github.com/jrodbx/agp-sources/search?q=StubRFile
or
https://github.com/jrodbx/agp-sources/search?q=GenerateLibraryRFileTask

will get you to some version of:
https://github.com/jrodbx/agp-sources/blob/a9517caa493219cfea2360f6ebb84daf5a5505f3/7.0.0-alpha07/com.android.tools.build/gradle/com/android/build/gradle/internal/res/GenerateLibraryRFileTask.kt

Looking at that file, you'll see things like:
```
    @get:OutputFile
    abstract val rClassOutputJar: RegularFileProperty
```
and
```
          creationConfig.artifacts.setInitialProvider(
                taskProvider,
                GenerateLibraryRFileTask::rClassOutputJar
            ).withName("R.jar").on(InternalArtifactType.COMPILE_AND_RUNTIME_NOT_NAMESPACED_R_CLASS_JAR)
```

Now, in the sample on master, there's only 1 R.jar.  But in the change proposed by this PR, i'm guessing there will be multiple GenerateLibraryRFileTask instances, each of which will produce an R.jar.

Looking at the current plugin code, there are some exposed TaskProviders on the variant type (i.e., BaseVariant); however, none seem to match the type we're looking for.  However, BaseVariant does expose `DomainObjectCollection<BaseVariantOutput> getOutputs()` and `BaseVariantOutput` exposes:

```
    /**
     * Returns the {@link TaskProvider} for the Android Resources processing task.
     *
     * <p>Prefer this to {@link #getProcessResources()} as it triggers eager configuration of the
     * task.
     */
    @NonNull
    TaskProvider<ProcessAndroidResources> getProcessResourcesProvider();
```

so maybe in the variant block, something like:
```
      val libraryRFileTaskProviders = variant.outputs.map { it.processResourcesProvider as TaskProvider<GenerateLibraryRFileTask> }
```
will finallly give you access to:

```
    @get:OutputFile
    abstract val rClassOutputJar: RegularFileProperty
```
or maybe?
```
    @get:Input
    abstract val packageForR: Property<String>
```

autonomousapps: Honestly I love that. Very cool process. Your point about artifacts is a good reminder that recent AGPs are starting to expose their artifacts directly, via either `android` or `androidComponents`. Might be worth checking what's available there.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -56,7 +58,12 @@ class PaparazziPlugin : Plugin<Project> {
         .libraryVariants
     variants.all { variant ->
       val variantSlug = variant.name.capitalize(Locale.US)
+      val testVariantSlug = variant.unitTestVariant.name.capitalize(Locale.US)
+
+      val nonTransitiveRClassEnabled = (project.findProperty("android.nonTransitiveRClass") as String?)?.toBoolean() ?: false
 
+      val generateStubRFileProvider = project.tasks.named("generate${testVariantSlug}StubRFile") as Provider<GenerateLibraryRFileTask>
```
jrodbx, line 65: temporary, until we can use artifact view apis to ask for artifacts of type `SYMBOL_LIST_WITH_PACKAGE_NAME`

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -75,10 +95,12 @@ open class PrepareResourcesTask : DefaultTask() {
         it.newLine()
         it.write(platformDataRoot.get().asFile.path)
         it.newLine()
+        it.write(resourcePackageNames)
+        it.newLine()
       }
   }
 
-  private fun Directory.relativize(child: Directory): String {
+  private fun Directory.relativize(child: FileSystemLocation): String {
```
jrodbx, line 103: do we still need this?

jrodbx (approved):

jrodbx:
Can the description be updated to reflect the slightly different approach, i.e., instead of scanning the R.jar for R.classes, we use AGP internals to determine the package names to class load?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
luis-cortes: Don't think so. Updating.

PR #257, opened 2021-09-08
Add support for fontScale in DeviceConfig

mattprecious:
This enables testing accessibility font sizes without having to
manually mutate the context with a test helper function.

---

Note: This was unblocked by the layoutlib bump, as the version that was being used before ignored the `fontScale` property on `SessionParams`.


jrodbx (approved):
nice!

Issue #243, opened 2021-08-16
Support Gradle configuration cache


Egorand:
https://docs.gradle.org/current/userguide/configuration_cache.html

Currently, running `./gradlew sample:verifyPaparazziDebug` produces the following errors when configuration cache is enabled:

```
* What went wrong:
Configuration cache problems found in this build.

10 problems were found storing the configuration cache, 5 of which seem unique.
- Task `:sample:preparePaparazziDebugResources` of type `app.cash.paparazzi.gradle.PrepareResourcesTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/6.9/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
- Task `:sample:testDebugUnitTest` of type `com.android.build.gradle.tasks.factory.AndroidUnitTest`: cannot serialize object of type 'app.cash.paparazzi.gradle.PaparazziPlugin$PaparazziTask', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.9/userguide/configuration_cache.html#config_cache:requirements:task_access
- Task `:sample:testDebugUnitTest` of type `com.android.build.gradle.tasks.factory.AndroidUnitTest`: cannot serialize object of type 'app.cash.paparazzi.gradle.PrepareResourcesTask', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.9/userguide/configuration_cache.html#config_cache:requirements:task_access
- Task `:sample:testDebugUnitTest` of type `com.android.build.gradle.tasks.factory.AndroidUnitTest`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.9/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:sample:testDebugUnitTest` of type `com.android.build.gradle.tasks.factory.AndroidUnitTest`: cannot serialize object of type 'org.gradle.api.DefaultTask', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.
  See https://docs.gradle.org/6.9/userguide/configuration_cache.html#config_cache:requirements:task_access
```

Here are the fixes that need to be made:

[...]

PR #261, opened 2021-09-10
Fix remaining configuration cache issues

jrodbx:
Thanks to @gavra0 and @bamboo for the pointers!

Closes https://github.com/cashapp/paparazzi/issues/243


geoff-powell (approved):
Nice! 🎉

PR #278, opened 2021-10-20
Add windows default android sdk path to environment class

rCollet:
I encountered an issue on Windows when running the sample tests:
`java.io.FileNotFoundException: C:\Users\rc\Library\Android\sdk\platforms\android-29\build.prop (The system cannot find the path specified)`

The android SDK path set in the Environment class only works for mac os, so I added a logic to use the default android SDK path on windows to make it work, in case no env. variables are present.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
@@ -17,6 +17,7 @@ package app.cash.paparazzi
 
 import java.io.File
 import java.nio.file.Paths
+import java.util.Locale
 
 data class Environment(
   val platformDir: String,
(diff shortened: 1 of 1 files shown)


jrodbx (approved):
thanks!

jrodbx:
I guess I should run this build on a window env before merging...

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -31,7 +32,15 @@ data class Environment(
 @Suppress("unused")
 fun androidHome() = System.getenv("ANDROID_SDK_ROOT")
     ?: System.getenv("ANDROID_HOME")
-    ?: "${System.getProperty("user.home")}/Library/Android/sdk"
+    ?: androidSDKPath()
+
+private fun androidSDKPath(): String {
```
jrodbx, line 37:
```suggestion
private fun androidSdkPath(): String {
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -31,7 +32,15 @@ data class Environment(
...
+    val homeDir = System.getProperty("user.home")
+    val osName = System.getProperty("os.name").lowercase(Locale.US)
+    val sdkPathDir = if (osName.startsWith("windows")) "\\AppData\\Local\\Android\\Sdk"
+        else "/Library/Android/sdk"
```
jrodbx, line 41: nit: braces

PR #283, opened 2021-11-01
Add internal repository configuration

jrodbx:
Makes deploying builds to internal repositories a bit easier

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,19 @@ subprojects {
           name = "projectLocalMaven"
           url = "${rootProject.buildDir}/localMaven"
         }
+        /**
+         * Want to push to an internal repository for testing?
+         * Set the following properties in ~/.gradle/gradle.properties.
+         *
+         * internalUrl=YOUR_INTERNAL_URL
+         * internalUsername=YOUR_USERNAME
+         * internalPassword=YOUR_PASSWORD
+         */
+        maven {
+          name = "internal"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -122,6 +122,18 @@ subprojects {
...
+        /**
+         * Are you a Square employee?  If so, then set the following properties in
+         * ~/.gradle/gradle.properties to enable pushing to our internal repository
+         *
+         * artifactoryUsername=YOUR_LDAP
+         * artifactoryPassword=YOUR_API_KEY
+         */
+        maven {
+          name = "artifactory"
+          url = "https://maven.global.square/artifactory/releases"
```
JakeWharton, line 134: We're not supposed to leak internal URLs. Can you use an environment variable? For Treehouse we use `-PRELEASE_REPOSITORY_URL="the url"` on the CLI to release internally.

jrodbx: ok, addressed by generalizing to any internal repository.  PTAL!

gradle.properties
```diff
@@ -1,5 +1,5 @@
 GROUP=app.cash.paparazzi
-VERSION_NAME=0.9.0-SNAPSHOT
+VERSION_NAME=0.9.0-alpha09
```
JakeWharton, line 2: Revert?

JakeWharton (approved):

gradle.properties
jrodbx: oops lol

PR #290, opened 2021-11-02
Use list instead of set for render extensions

geoff-powell:


jrodbx (commented):
How come?  For example, would it ever make sense to have multiple AccessibilityRenderExtensions?

geoff-powell:
I guess it didn't make sense to limit render extensions by a single instance. Since the use case is only one example (accessibility) I can close this for now

Issue #282, opened 2021-10-29
Decouple Paparazzi from JUnit 4 (TestRule)


TWiStErRob:
I'll add an example what this means, how JUnit 5 extensions work.

TWiStErRob:
See https://github.com/cashapp/paparazzi/pull/293

kirillzh:
+1 for this request, would love to see paparazzi get decoupled from Junit 4 test rule. In our use case, we are using Kotest framework, and from what I have found so far, there's no integration for Junit 4 test rules. The solution that I came up with is to implement my own extension that mimics what `Paparazzi` rule logic does, looks like this:

```kotlin
class PaparazziExtension(
 val paparazzi: Paparazzi,
) : BeforeTestListener, AfterTestListener {

  override suspend fun beforeTest(testCase: TestCase) {
    val junit4Description = testCase.junit4Description
    paparazzi.prepare(junit4Description)
    paparazzi.apply(
      base = object : Statement() {
        override fun evaluate() = Unit
      },
      description = junit4Description,
    )
  }

  override suspend fun afterTest(testCase: TestCase, result: TestResult) {
    paparazzi.close()
  }
}

private val TestCase.junit4Description: Description
  get() {
    return Description.createTestDescription(descriptor.parent.id.value, name.testName)
  }
```

Usage looks like this:

```kotlin
class ButtonSnapshots : FunSpec({
  val paparazzi = Paparazzi(deviceConfig = PIXEL_6)
  extension(PaparazziExtension(paparazzi))

  test("button - primary treatment, compact size, enabled, no icon") {
    paparazzi.snapshot {
      Button(
        treatment = ButtonTreatment.Primary,
        size = ButtonSize.Compact
      )
    }
  }
})
```

anhanh11001:
+1 for this request. 

Can I also ask if there are estimated timeline for this + Milestone 1.2 (seems like end of 2022, but not sure if this is correct)?

nuhkoca:
Hello,

Is it possible to see that is happening in [2.0.0-alpha03](https://github.com/cashapp/paparazzi/milestone/19)?

PR #293, opened 2021-11-03
JUnit 4 and Jupiter integration

TWiStErRob:
As promised in #282. See the individual commits.

Let me know what you think of it so far, and happy to help with other things, if necessary, but also feel free to commit and finish this branch/PR.

Few things of note left to do (anyone can chip in!):

- [ ] Introduce a Paparazzi Interface which can be implemented by the JUnit rule and delegates to instance  
      (unless you're ok with `paparazzi.instance.snapshot`)
- [ ] Add the logic of AgentTestRule to PaparazziExtension (not sure if it can be composed the same way, might need to inline it there as one extension.
- [ ] Add `registerInterceptors` to PaparazziExtension (I didn't do this because it wasn't clear what was your intended lifecycle of this. I'm not sure if it's defined when the body of `TestRule.apply` is executed. I don't think that can be relied on as a "lifecycle" event. Anyway, I'm sure there's a fitting lifecycle interface in `org.junit.jupiter.api.extension` package.
- [ ] Consider renaming junit5 to jupiter, 5 is not the most fitting, see https://junit.org/junit5/docs/current/user-guide/#overview-what-is-junit-5.
- [ ] Mention in Paparazzi docs and on extension class that any kind of [parallel execution](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution) of jupiter is not supported by Paparazzi because &lt;reason>.
- [ ] Fix all the compilation errors and tests, and provide migration path / docs for users.
- [ ] Review TestName implementation, I'm not sure that's the right thing to do in Paparazzi, it looks like doesn't take into account JUnit `@Parameterized` tests, JUnitParams `@Parameters(...)` or JUnit Jupiter `@ParameterizedTest`.


TWiStErRob:
I'm closing this POC as I'm not planning to complete it. If anyone picks it up in the future, can use it as a reference for what to do, probably in separate PRs.

nishatoma:
@TWiStErRob  Thank you for detailed work! I started work on this before stumbling on this PR.

My solution is a bit different (a bit more on the consumer side). I tested locally and made it work with couple different libraries (TestNg, Kotest, Jupiter, etc).

I can demo this with a PR.

Happy new year!!

PR #307, opened 2021-11-07
Add limitation on JDK version

TWiStErRob:


jrodbx:
This is imposed as of a recent version of Android Studio (Arctic Fox) so all Android engineers must upgrade to build their apps on the new stack.

TWiStErRob:
I was using AGP 4.2.2 with IntelliJ IDEA 2021.2.3 with Gradle 6.9.1 on Java 8. No issues.
If someone cannot upgrade AGP just yet, this is an important consideration as they will have to up their Java to integrate paparazzi. Would love to live on the edge, but many enterprises are different.

jrodbx:
Closing as JDK 11 is now the default expected minimum for Android apps.

PR #310, opened 2021-11-08
Add warning about limits of LFS

TWiStErRob:
It wasn't apparent from https://git-lfs.github.com/, only realized this after I got mail.
image

Essentially if one has 100 MBs of screenshots, one can do only 10 GitHub action CI builds with verifyPaparazzi.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ android.jetifier.ignorelist=android-base-common,common
 
 Git LFS
 --------
-It is recommended you use [Git LFS][lfs] to store your snapshots.  Here's a template to get started.
+It is recommended you use [Git LFS][lfs] to store your snapshots. Here's a template to get started.
+_Beware that on GitHub there are [limitations on size and bandwidth][lfs-limit] and using LFS may incur [additional cost][lfs-cost]._
 
 #### **`.gitattributes`**
 ```bash
@@ -212,3 +213,5 @@ limitations under the License.
  [sample]: https://github.com/cashapp/paparazzi/tree/master/sample
  [snap]: https://oss.sonatype.org/content/repositories/snapshots/app/cash/paparazzi/
  [lfs]: https://git-lfs.github.com/
+ [lfs-limit]: https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
+ [lfs-cost]: https://docs.github.com/en/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage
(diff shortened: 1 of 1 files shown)


jrodbx:
This is a GitHub constraint, not a Paparazzi one.  For example, we use GitHub Enterprise on CashApp.  Other projects may use other repos, so this advisory doesn't feel right here since it'll depend on the project's setup.

TWiStErRob:
Ah, I thought LFS was a GitHub cloud service, [turns out I was wrong](https://github.com/git-lfs/git-lfs/wiki/Implementations). I agree it's not strictly related to paparazzi, but I think it's an important point when recommending LFS in the readme. I found out the hard way, If there's space for the hooks (which are generated similarly via `git lfs install --local`, there should be space for a little sentence. Maybe a better phrased one?

jrodbx:
The billing policies of the repo that provides LFS support is out of scope for this project.  It's only a recommendation to use LFS given the way binary content is tracked.

TWiStErRob:
How about this?

jrodbx:
The thing is, this still makes recommendations on implementations, which really isn't the project's concern.  

Wanna use an implementation that charges for storage?  Cool.  Wanna use a free version?  Cool.

How the project's users want to implement Git LFS is their call.  Similar to branching strategy or other processes, Paparazzi will work with all of them!

TWiStErRob:
Is spreading the word that bad? I have never heard about LFS before, and even after spending a weekend trying to integrate with it I found out the hard way that GitHub has limitations on it.

I'm just trying to help fellows here who might go through the same cycle: "What LFS? Ah cool. GitHub is providing an awesome service yet again! Oh 1GB only, paid?! Ah, there are other choices.".

jrodbx:
> Is spreading the word that bad? I have never heard about LFS before, and even after spending a weekend trying to integrate with it I found out the hard way that GitHub has limitations on it.
> 
> I'm just trying to help fellows here who might go through the same cycle: "What LFS? Ah cool. GitHub is providing an awesome service yet again! Oh 1GB only, paid?! Ah, there are other choices.". You might be looking at it from the perspective of someone who already knows, but most devs will need to learn about this.

Spreading the word about your experience learning the difference between Git LFS and Github's paid LFS service is not bad at all.  However, my stance is that that learned lesson doesn't belong in this repo.

Instead, how about tweeting/blogging about your experience setting up Paparazzi in our project once we land some of the other issues you highlighted?

TWiStErRob:
Ok, got it. Thoughts on other improvements to README > LFS?

1. reorder to keep "getting started" on top?
2. remove redundant hooks as `git lfs install` does some automatically?

PR #319, opened 2021-11-16
Reset AnimationHandler singleton after snapshot

jrodbx:
When testing Windows CI builds, this bug mysteriously arose due to the natural platform-invocation ordering for AccessibilityRenderExtensionTest.test and PaparazziTest.animationEvents.  Both tests require and create a AnimationHandler singleton. Whichever test creates that singleton first, leaks it, and prevents animation update callbacks from being enqueued in subsequent tests.

As a result, it turns out that on Windows, AccessibilityRenderExtensionTest.test ran first, created the AnimationHandler singleton due to the inclusion of a Button in the snapshotted view hierarchy, then prevented PaparazziTest.animationEvents from receiving any animation updates after the first frame.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -15,6 +15,7 @@
  */
 package app.cash.paparazzi
 
+import android.animation.AnimationHandler
 import android.content.Context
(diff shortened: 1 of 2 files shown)


geoff-powell (approved):
Hmm, well I'm glad the AccessibilityRenderExtension was helpful in discovering this 😄

PR #325, opened 2021-11-18
Improve usage by Java consumers

### LINKED ISSUE #312: Usage from Java requires red code
The equivalent code for
```kotlin
Paparazzi(
    theme = "AppTheme.ScreenshotTest",
    maxPercentDifference = 0.0,
    renderExtensions = listOf(CropViewRenderExtension()),
)
```
used in a Java project:
```java
new Paparazzi(
		EnvironmentKt.detectEnvironment(),
		DeviceConfig.Companion.getNEXUS_5(),
		"AppTheme.ScreenshotTest",
		true,
		0.0,
		Boolean.parseBoolean(System.getProperty("paparazzi.test.verify"))
				? new SnapshotVerifier(0.0, new File("src/test/snapshots"))
				: new HtmlReportWriter(),
		new HashSet<>(Collections.singletonList(new CropViewRenderExtension()))
);
```
gives an error in the IDE (fine during compilation).
image

jrodbx:
Closes #312

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
@@ -57,7 +57,7 @@ import javax.imageio.ImageIO
  * paparazzi.js
  * ```
  */
-internal class HtmlReportWriter(
+class HtmlReportWriter @JvmOverloads constructor(
   private val runName: String = defaultRunName(),
   private val rootDirectory: File = File("build/reports/paparazzi"),
   snapshotRootDirectory: File = File("src/test/snapshots")
--- a/paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
@@ -118,8 +118,10 @@ data class DeviceConfig(
    */
 
   companion object {
+    @JvmField
     val NEXUS_4 = DeviceConfig()
 
+    @JvmField
     val NEXUS_5 = DeviceConfig(
         screenHeight = 1920,
         screenWidth = 1080,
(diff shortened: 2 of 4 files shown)


JakeWharton (approved):

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -57,7 +57,7 @@ import javax.imageio.ImageIO
  * paparazzi.js
  * ```
  */
-internal class HtmlReportWriter(
+class HtmlReportWriter @JvmOverloads constructor(
```
Egorand, line 60: Does this class need to be in the public API?

Egorand (approved):

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
jrodbx: Good question.  In the referenced issue, the snapshot handler is the second-to-last arg, so the equivalent java overload makes it required to set this when setting RenderExtensions.

I _could_ swap the parameter ordering, but wasn't sure if that was good form, i.e., does that mean all future parameter additions need to not be last?

Egorand: Interesting. For Java we'd probably just want a separate `Builder`..

jrodbx: Ah true, didn't even think of providing _that_ level of support.  Isn't everyone on Kotlin by now?  😛 

I'll merge this for now and follow-up with Builder.

Egorand: We can maybe just say "sorry no explicit Java support"? IDK

PR #326, opened 2021-11-19
Use managed properties to clean up PrepareResourcesTask

jrodbx:
https://docs.gradle.org/current/userguide/custom_gradle_types.html#managed_properties


luis-cortes (approved):

Egorand (commented):
Awesome, looks like this should help with configuration caching support 👍

jrodbx:
> Awesome, looks like this should help with configuration caching support 👍 

Managed properties have no impact on configuration caching support.  Besides, the plugin has full support as of https://github.com/cashapp/paparazzi/pull/261.

Egorand:
Ah, okay, just noticed that this change removes references to `project` which config caching prohibits.

jrodbx:
Ah, I see.  Actually, configuration caching prohibits references to project in task actions (annotated by @TaskAction) but not all references to project.  In fact, it would be difficult to remove _all_ references since Project to Gradle is like Context to Android 😂

Egorand:
Haha, fair enough. But glad to hear the project supports config caching!

PR #338, opened 2021-11-29
Add NPE check for AccessibilityRenderExtension when layout params are null

geoff-powell:
Small fix to handle views under test that do not have layout params set.

Tested by running AccessibilityRenderExtensionTest.kt with a view without params set.

```
java.lang.NullPointerException
	at app.cash.paparazzi.accessibility.AccessibilityRenderExtension.renderView(AccessibilityRenderExtension.kt:71)
	at app.cash.paparazzi.Paparazzi.takeSnapshots(Paparazzi.kt:246)
	at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:176)
	at app.cash.paparazzi.Paparazzi.snapshot$default(Paparazzi.kt:169)
	at app.cash.paparazzi.accessibility.AccessibilityRenderExtensionTest.test(AccessibilityRenderExtensionTest.kt:40)
	...
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
```


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -66,8 +66,8 @@ class AccessibilityRenderExtension : RenderExtension {
       addView(
         contentView,
         LinearLayout.LayoutParams(
-          contentView.layoutParams.width,
-          contentView.layoutParams.height,
+          contentView.layoutParams?.width ?: ViewGroup.LayoutParams.WRAP_CONTENT,
+          contentView.layoutParams?.height ?: ViewGroup.LayoutParams.WRAP_CONTENT,
```
jrodbx, line 70: wouldn't it be better to fail fast?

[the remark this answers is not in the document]
geoff-powell: I guess my thought here is that if the provided view under test doesn't have layout params, should the test fail?

My thought is that it shouldn't fail but open to hear any alternatives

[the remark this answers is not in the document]
saket: I'm thinking that it is acceptable for a View to not have any layout params. Android will assign it a default one by asking its parent (using `ViewGroup#generateLayoutParams()`).

Should paparazzi also accept this scenario?

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -23,23 +23,34 @@ import java.io.File
...
   fun test() {
     val view = buildView(paparazzi.context)
+    snapshotHandler.shouldVerifySnapshot = true
+
     paparazzi.snapshot(view)
   }
 
+  @Test
+  fun `test without layout params set`() {
+    snapshotHandler.shouldVerifySnapshot = false
```
jrodbx, line 49: why wouldn't the snapshot be verified here?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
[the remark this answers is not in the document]
jrodbx: fair point, @saket !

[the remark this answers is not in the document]
geoff-powell: fixed this to use `ViewGroup#generateLayoutParams()` as default.

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -85,14 +94,18 @@ class AccessibilityRenderExtensionTest {
...
+
     override fun newFrameHandler(
       snapshot: Snapshot,
       frameCount: Int,
       fps: Int
     ): SnapshotHandler.FrameHandler {
       return object : SnapshotHandler.FrameHandler {
         override fun handle(image: BufferedImage) {
-          val expected = File("src/test/resources/accessibility.png")
+          snapshotPath ?: return
```
jrodbx, line 106: instead of early return, how about providing a golden value for the added test?

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -105,6 +118,8 @@ class AccessibilityRenderExtensionTest {
       }
     }
 
-    override fun close() = Unit
+    override fun close() {
+      snapshotPath = null
```
jrodbx, line 122: revert?  string path isn't a Closeable/resource.

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -16,37 +16,49 @@ import app.cash.paparazzi.Paparazzi
 import app.cash.paparazzi.Snapshot
 import app.cash.paparazzi.SnapshotHandler
 import app.cash.paparazzi.internal.ImageUtils
+import com.google.android.apps.common.testing.accessibility.framework.replacements.LayoutParams
```
jrodbx, line 19: wrong import?

jrodbx (approved):

PR #347, opened 2022-01-05
Use BuildList for resourcePackageNames to remove hard coded separator

### LINKED ISSUE #346: Using non transitive resources without additional dependencies leads to  ClassNotFoundException
When the module using Paparazzi has no additional dependencies, setting `android.nonTransitiveRClass=true` in `gradle.properties` can lead to this error: `java.lang.ClassNotFoundException: /R`

Turns out that this code 

https://github.com/cashapp/paparazzi/blob/4890fdc31d266f68e8e79e21c6a879064d590a99/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt#L74-L80

adds a "," after the `mainPackage` even if `artifactFiles` is empty, causing an empty string to be returned when splitting the string here

https://github.com/cashapp/paparazzi/blob/4890fdc31d266f68e8e79e21c6a879064d590a99/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt#L54

The returned list looks like this `["$mainPackage", ""]`

I'll add a potential fix and test cases later.

LuK1709:
Resolves #346

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -66,15 +66,19 @@ abstract class PrepareResourcesTask : DefaultTask() {
   private val projectDirectory = project.layout.projectDirectory
 
   @TaskAction
+  @OptIn(ExperimentalStdlibApi::class)
   fun writeResourcesFile() {
     val out = paparazziResources.get().asFile
     out.delete()
 
     val mainPackage = packageName.get()
     val resourcePackageNames = if (nonTransitiveRClassEnabled.get()) {
-      "$mainPackage," + artifactFiles.files.joinToString(",") { file ->
-        file.useLines { lines -> lines.first() }
-      }
+      buildList {
+        add(mainPackage)
+        addAll(artifactFiles.files.map { file ->
+          file.useLines { lines -> lines.first() }
+        })
+      }.joinToString(",")
     } else {
       mainPackage
     }
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -75,9 +75,9 @@ abstract class PrepareResourcesTask : DefaultTask() {
     val resourcePackageNames = if (nonTransitiveRClassEnabled.get()) {
       buildList {
         add(mainPackage)
-        addAll(artifactFiles.files.map { file ->
-          file.useLines { lines -> lines.first() }
-        })
+        artifactFiles.files.map { file ->
+          add(file.useLines { lines -> lines.first() })
+        }
       }.joinToString(",")
     } else {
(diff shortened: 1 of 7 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -66,15 +66,19 @@ abstract class PrepareResourcesTask : DefaultTask() {
   private val projectDirectory = project.layout.projectDirectory
 
   @TaskAction
+  @OptIn(ExperimentalStdlibApi::class)
```
jrodbx, line 69: @JakeWharton, correct if I'm wrong, but I think you had said in the past (maybe in a Timber PR?) that using experimental apis, like buildList etc. in libraries, isn't ideal.  Is that guidance still accurate?

LuK1709: There where already some instances of `ExperimentalStdlibApi`, one example being https://github.com/cashapp/paparazzi/blob/9fcffd16467697d85cb4518a188cc7fbfab93092/paparazzi/src/main/java/app/cash/paparazzi/internal/parsers/ResourceParser.kt#L124-L126
That's why I included it here. 

Apart from that, are there any changes required?

jrodbx: fair enough.  i forgot we used that elsewhere.  we can migrate later then, if an issue!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -66,15 +66,19 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
     out.delete()
 
     val mainPackage = packageName.get()
     val resourcePackageNames = if (nonTransitiveRClassEnabled.get()) {
-      "$mainPackage," + artifactFiles.files.joinToString(",") { file ->
-        file.useLines { lines -> lines.first() }
-      }
+      buildList {
+        add(mainPackage)
+        addAll(artifactFiles.files.map { file ->
```
jrodbx, line 78: nit: if we move `add` inside the map, we avoid creation of a temp list?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -66,15 +66,19 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+          file.useLines { lines -> lines.first() }
+        })
```
jrodbx, line 80:
```suggestion
        artifactFiles.files.map { file ->
          add(file.useLines { lines -> lines.first() })
        }
```

ChristianBecker: Shouldn't the `map` become a `forEach` to get rid of the temp list? I just created #352 for it.

PR #350, opened 2022-01-07
Use View.id to generate colors for accessibility entries

### LINKED ISSUE #348: Accessibility snapshots highlight colours change when the view is edited
After updating a `View` class that has an accessibility snapshot created for it, the accessibility Paparazzi tests for it fail as the colours used to highlight the children in the image change.

Ideally, the accessibility test would still generate a snapshot with the same colours used previously to avoid needing to replace the existing images.

geoff-powell:
Fixes #348


jrodbx (approved):
nice

PR #355, opened 2022-01-18
Support M1 in renderer

### LINKED ISSUE #354: Renderer still looks for incorrect layoutlib on Mac M1
It looks like this logic does not account for M1.

```kotlin
    val osLabel = when {
      osName.startsWith("windows") -> "win"
      osName.startsWith("mac") -> "mac"
      else -> "linux"
    }
    return "$osLabel/lib64"
 ```

The library will be found in `mac-arm/lib64`

https://github.com/cashapp/paparazzi/blob/f315460e40ed3deca3d954a9d515a1c3466423e3/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt#L110

nicbell:
This should fix #354


hshandilya-swiggy (approved):
I've tested this on an M1 MBP and can confirm this fixes bridge initialization.

jrodbx (approved):
oops, thanks!

PR #357, opened 2022-01-20
Require view id to generate color palette for accessibility render extension

geoff-powell:
Makes tests more deterministic by generating view ids for views to generate colors for accessibility.

Using `System.identityHashCode(view)` ended up using memory register location which is not deterministic.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
@@ -31,9 +31,9 @@ import app.cash.paparazzi.accessibility.RenderSettings.DEFAULT_TEXT_SIZE
 import app.cash.paparazzi.accessibility.RenderSettings.getColor
 import app.cash.paparazzi.accessibility.RenderSettings.toColorInt
 import app.cash.paparazzi.accessibility.RenderSettings.withAlpha
+import java.util.concurrent.atomic.AtomicInteger
 
 class AccessibilityRenderExtension : RenderExtension {
-
   override fun renderView(
     contentView: View
   ): View {
@@ -80,6 +80,8 @@ class AccessibilityRenderExtension : RenderExtension {
           1f
         )
       )
+    }.also {
+      RenderSettings.resetGeneratedId()
     }
   }
 
--- a/paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
@@ -17,8 +17,11 @@ package app.cash.paparazzi.accessibility
 
 import android.view.View
 import java.awt.Color
+import java.util.concurrent.atomic.AtomicInteger
 
 internal object RenderSettings {
+  private val nextGeneratedId = AtomicInteger(1)
+
   const val DEFAULT_RENDER_ALPHA = 40
   val DEFAULT_RENDER_COLORS = listOf(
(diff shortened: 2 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -44,8 +44,8 @@ internal object RenderSettings {
   }
 
   private fun nextColor(view: View): Color {
-    val viewId = if (view.id != View.NO_ID) view.id else System.identityHashCode(view)
-    return DEFAULT_RENDER_COLORS[viewId % DEFAULT_RENDER_COLORS.size]
+    require(view.id != View.NO_ID) { "Views must have ids to generate deterministic colors" }
```
jrodbx, line 47: no need for require, when it's a private method.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -44,8 +44,8 @@ internal object RenderSettings {
   }
```
jrodbx, line 44: how about:
```
  fun getColor(view: View): Color {
    if (view.id == View.NO_ID) {
      view.id = generateViewId()
    }
    return getColor(view.id)
  }

  private fun getColor(viewId: Int): Color {
    return colorMap.getOrPut(viewId) {
      nextColor(viewId).withAlpha(DEFAULT_RENDER_ALPHA)
    }
  }
```

jrodbx: also, colorMap can become mutableMapOf<Int, Color>() as a result

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -31,9 +31,12 @@ import app.cash.paparazzi.accessibility.RenderSettings.DEFAULT_TEXT_SIZE
 import app.cash.paparazzi.accessibility.RenderSettings.getColor
 import app.cash.paparazzi.accessibility.RenderSettings.toColorInt
 import app.cash.paparazzi.accessibility.RenderSettings.withAlpha
+import java.util.concurrent.atomic.AtomicInteger
 
 class AccessibilityRenderExtension : RenderExtension {
 
+  private val nextGeneratedId = AtomicInteger(1)
```
jrodbx, line 38: another option, instead of having the renderExtension manage the state, we can move this to RenderSettings and update in getColor, as suggested below.

If we do that though, then either:
1) RenderSettings should no be longer be an object, since it's stateful
2) or workaround this, by adding 

  fun resetGeneratedId() = nextGeneratedId.set(1)

and calling resetGeneratedId() towards the end of AccessibilityRenderExtension.renderView

jrodbx (approved):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -54,4 +63,17 @@ internal object RenderSettings {
   internal fun Color.withAlpha(alpha: Int): Color {
     return Color(red, green, blue, alpha)
   }
+
+  internal fun resetGeneratedId() = nextGeneratedId.set(1)
+
+  // Taken from View.generateViewId()
+  private fun generateViewId(): Int {
```
swankjesse, line 70: You can avoid this loop with `nextGeneratedId.getAndIncrement()`

PR #361, opened 2022-01-22
Fix deprecation warnings in sample

TWiStErRob:
```
w: P:\projects\contrib\github-paparazzi\sample\src\test\java\app\cash\paparazzi\sample\KeypadViewTest.kt: (46, 40): 'getColor(Int): Int' is deprecated. Deprecated in Java
w: P:\projects\contrib\github-paparazzi\sample\src\test\java\app\cash\paparazzi\sample\KeypadViewTest.kt: (52, 51): 'getColor(Int): Int' is deprecated. Deprecated in Java
```

Had the option of `paparazzi.resources.getColor(..., null)` or `paparazzi.context.getColor(...)`, latter looks simpler and more likely used (indirectly via `ContextCompat.getColor(context, ...)`).


jrodbx (approved):

jrodbx:
thanks!

Issue #351, opened 2022-01-07
NoSuchFieldError


sdoward:
When running `recordPaparazzi` I see the following error.

I can only find a reference to IStart in  `CamColor.java` in androidx.core. 

```
java.lang.NoSuchFieldError: lStar
	at org.sharethemeal.android.view.R$styleable.<clinit>(Unknown Source)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1042)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorFactory.newFieldAccessor(UnsafeFieldAccessorFactory.java:43)
	at java.base/jdk.internal.reflect.ReflectionFactory.newFieldAccessor(ReflectionFactory.java:186)
	at java.base/java.lang.reflect.Field.acquireFieldAccessor(Field.java:1105)
	at java.base/java.lang.reflect.Field.getFieldAccessor(Field.java:1086)
	at java.base/java.lang.reflect.Field.get(Field.java:418)
	at app.cash.paparazzi.internal.PaparazziCallback.initResources(PaparazziCallback.kt:73)
	at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:128)
	at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:106)
	at app.cash.paparazzi.agent.AgentTestRule$apply$1.evaluate(AgentTestRule.kt:17)
```

sdoward:
This issue only seems to be present on version 1.7.0 of the androidx lib 1.6.0 works correctly

mochadwi:
> This issue only seems to be present on version 1.7.0 of the androidx lib 1.6.0 works correctly

how to use androidx lib 1.6.0?

sdoward:
@mochadwi I might have been wrong with that. I am currently using with 1.5.0

mochadwi:
well noted, which dependency should we point to 1.5.0? does it only `androidx.core`? 

<img width="370" alt="Screen Shot 2022-01-30 at 5 49 52 PM" src="https://user-images.githubusercontent.com/7057959/151696578-b3a7c91e-1b17-48f5-8562-7766fbb8c013.png">

sdoward:
`androidx.core` is the one that cause the initial problems for me. But be aware you might have transitive dependencies using higher version

[...]

Issue #264, opened 2021-09-11
Test Paparazzi with Android S


jrodbx:
Tests broke when I ran locally, setting the testConfig in `paparazzi/build.gradle` to point to API 31.

```
$ ./gradlew paparazzi:test --tests=app.cash.paparazzi.PaparazziTest

> Task :paparazzi:test

app.cash.paparazzi.PaparazziTest > throwsRenderingExceptions FAILED
    java.lang.ExceptionInInitializerError
        at android.view.autofill.AutofillManager.<clinit>(AutofillManager.java:239)
        at com.android.layoutlib.bridge.android.BridgeContext.<init>(BridgeContext.java:256)
        at com.android.layoutlib.bridge.impl.RenderAction.init(RenderAction.java:133)
        at com.android.layoutlib.bridge.impl.RenderSessionImpl.init(RenderSessionImpl.java:180)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:139)
        ...

        Caused by:
        java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
            at android.os._Original_Build.<clinit>(Build.java:214)
            ... 42 more

app.cash.paparazzi.PaparazziTest > drawCalls FAILED
    java.lang.NullPointerException
        at com.android.layoutlib.bridge.impl.Layout$Builder.findAppCompatActionBar(Layout.java:455)
        at com.android.layoutlib.bridge.impl.Layout$Builder.<init>(Layout.java:422)
        at com.android.layoutlib.bridge.impl.RenderSessionImpl.init(RenderSessionImpl.java:193)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:139)
        ...

app.cash.paparazzi.PaparazziTest > animationEvents FAILED
    java.lang.NullPointerException
        at com.android.layoutlib.bridge.impl.Layout$Builder.findAppCompatActionBar(Layout.java:455)
        at com.android.layoutlib.bridge.impl.Layout$Builder.<init>(Layout.java:422)
        at com.android.layoutlib.bridge.impl.RenderSessionImpl.init(RenderSessionImpl.java:193)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:139)
        ...
```

TWiStErRob:
Same here

TWiStErRob:
Note this also prevents using latest appcompat:
```
* What went wrong:
[...]

PR #368, opened 2022-02-02
Reproduce androidx.core bug

sdoward:
I have been struggling with a bug concerning `androidx.core:core.1.7.0`. See https://github.com/cashapp/paparazzi/issues/351

I have been able to reproduce in the sample project by bumping the Android gradle plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
 
   ext.versions = [
     kotlin: '1.5.31',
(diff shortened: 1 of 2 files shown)


jrodbx:
Related to https://github.com/cashapp/paparazzi/issues/264 as mentioned in https://github.com/cashapp/paparazzi/issues/351

Issue #225, opened 2021-03-30
Compose support


benkay:
I took a stab at getting a ComposeView working with Paparazzi. It doesn't work with 0.7.0, but I managed to get it working on the native layoutlib branch with a bit of hackery. I think most of this is solved by using native/newer layoutlib, but It'd be great if there was a built-in `paparazzi.snapshot(content: @Composable () -> Unit)` at some point

- Compose calls ViewTreeLifecycleOwner/ViewTreeSavedStateRegistryOwner.get(rootView). A workaround for now is to wrap the ComposeView in a FrameLayout which sets these on the root view in onAttachedToWindow
- I had to provide mocked versions of AutofillManager and ClipboardManager as the 4.1.0 layoutlib doesn't provide them from BridgeContext, and Compose looks them both up when initialising. It looks like the latest prebuilt from mirror-goog-studio-master-dev does, but I can't get past an unsatisfiedlinkerror when trying to use it. Pre-releases of studio aren't tagged, so I have no idea whether the 4.2.0 version has support for these or not..

timrijckaert:
Hello @jrodbx I saw you committed https://github.com/cashapp/paparazzi/commit/e71a024d823cafa0cd2dc91765e8840d6ab3af67

Does that work for you?
I'm getting 

```
java.lang.ExceptionInInitializerError
	at android.view.autofill.AutofillManager.<clinit>(AutofillManager.java:252)
	at com.android.layoutlib.bridge.android.BridgeContext.<init>(BridgeContext.java:265)
	at com.android.layoutlib.bridge.impl.RenderAction.init(RenderAction.java:143)
	at com.android.layoutlib.bridge.impl.RenderSessionImpl.init(RenderSessionImpl.java:186)
	at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:146)
	at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:106)
	at app.cash.paparazzi.agent.AgentTestRule$apply$1.evaluate(AgentTestRule.kt:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
[...]

PR #371, opened 2022-02-15
Initial support for rendering Compose snapshots

jrodbx:
🙈 

Closes https://github.com/cashapp/paparazzi/issues/225


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -165,6 +177,22 @@ class Paparazzi @JvmOverloads constructor(
 
   fun <V : View> inflate(@LayoutRes layoutId: Int): V = layoutInflater.inflate(layoutId, null) as V
 
+  @JvmOverloads
+  fun snapshot(
+    composable: @Composable () -> Unit,
+    name: String? = null,
+    deviceConfig: DeviceConfig? = null,
+    theme: String? = null,
+    renderingMode: RenderingMode? = null
```
saket, line 186: Can I suggest moving `composable` to the last param? This would let us write tests in this manner:

```kotlin
paparazzi.snapshot(name = "foo") {
  Text("whoa")
}
```

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -492,6 +520,24 @@ class Paparazzi @JvmOverloads constructor(
     )
   }
 
+  class PaparazziComposeOwner private constructor() : LifecycleOwner, SavedStateRegistryOwner {
```
saket, line 523: does this need to be public?

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -165,6 +177,22 @@ class Paparazzi @JvmOverloads constructor(
...
+  @JvmOverloads
+  fun snapshot(
+    composable: @Composable () -> Unit,
+    name: String? = null,
+    deviceConfig: DeviceConfig? = null,
+    theme: String? = null,
+    renderingMode: RenderingMode? = null
+  ) {
+    val hostView = ComposeView(context)
+    val parent = FrameLayout(context).apply { id = android.R.id.content }
```
saket, line 189: Thoughts on adding a comment on why this ID is needed for our future selves? This is such a hidden implementation detail.

saket (approved):

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: I was opting for consistency, but you raise a fair point.  Is it a weird api to have View be the first arg in legacy Android snapshots, while a Composable be the last arg?

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: nope, will fix!

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
saket: I actually think they are both consistent because I can call `paparazzi.snapshot` without any extra configuration:

```kotlin
paparazzi.snapshot(ImageView())

paparazzi.snapshot {
  Image()
}
```

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
saket: hidden and impressive*

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: What about the cases where extra configuration is preferred?

As is, without configuration:
```
paparazzi.snapshot(ImageView())
paparazzi.snapshot({Image()})
```

As is, with configuration:
```
paparazzi.snapshot(ImageView(), name = "foo")
paparazzi.snapshot({Image()}, name = "foo")
```

Proposed, without configuration:
```
paparazzi.snapshot(ImageView())
paparazzi.snapshot { Image() }
```

Proposed, with configuration:
```
paparazzi.snapshot(ImageView(), name = "foo")
paparazzi.snapshot(name = "foo") { Image() }
```

Of course, we can move View to the last param, but that also will break everyone's current tests...

saket: image

I'd say this is good enough! Both these APIs are in line with their respective worlds.

swankjesse (approved):

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jingibus: Or it could be fluent:

```
interface PaparazziConfiguration {
  fun deviceConfig(deviceConfig: DeviceConfig): PaparazziConfiguration
  fun theme(theme: String): PaparazziConfiguration
  fun renderingMode(renderingMode: RenderingMode): PaparazziConfiguration

  fun snapshot(view: View)
  fun snapshot(composable: @Composable ()->Unit)
}
```

Has the advantage of providing a smaller ABI surface area on each method call!

paparazzi-gradle-plugin/src/test/projects/compose/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+android {
+  compileSdkVersion 30
+  defaultConfig {
+    minSdkVersion 25
+  }
+  buildFeatures {
+    compose true
+  }
+  composeOptions {
+    kotlinCompilerExtensionVersion '1.0.5'
```
saket, line 25: Can apps use other versions of Compose?

jrodbx: i don't know, @JakeWharton any ideas?

JakeWharton: Yes. All newer compiler+runtime pairs are required to handle class files compiled by older, stable compiler+runtime pairs. Otherwise each new version of Compose would require _every_ library to recompile and ship new versions.

paparazzi/build.gradle
```diff
@@ -22,6 +24,12 @@ dependencies {
 }
 
 dependencies {
+  api files('libs/compose-runtime-1.0.5.jar')
+  compileOnly files('libs/compose-ui-1.0.5.jar')
+  compileOnly deps.androidx.lifecycleCommon
+  compileOnly files('libs/lifecycle-runtime-2.4.0.jar')
+  compileOnly files('libs/savedstate-1.1.0.jar')
```
JakeWharton, line 31: May be worth leaving a comment as to why we had to do this / chose to do this.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -165,6 +177,25 @@ class Paparazzi @JvmOverloads constructor(
 
   fun <V : View> inflate(@LayoutRes layoutId: Int): V = layoutInflater.inflate(layoutId, null) as V
 
+  @JvmOverloads
```
JakeWharton, line 180:
```suggestion
```
Java users can't use Compose

JakeWharton (approved):

PR #376, opened 2022-02-19
Introduce externally-managed RendererScope to share Renderer across test suite

### LINKED ISSUE #369: FrameworkResources  take a significant amount of time to load
I've been noticing that my paparazzi tests tend to take between 10-15 seconds each. After digging in I've found that loading framework resources seems to be taking the lion's share of test time for each test. This seems to take around 5-8 seconds per test. 

This is happening here:
https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt#L50

It looks like that's going through all of the android framework resources and loading some (or maybe all?) of them from disk (this amounts to around 2500 files for `android-30`).

I'm guessing this needs to happen so we have system resources available when we take the snapshots, but I'm wondering if maybe this could be done once statically and shared between tests? Having this happen per test adds a fair amount of overhead which can really accumulate when you have a bunch of snapshot tests.

jrodbx:
Closes #369, #247 

An alternative to #247 which allows for a user-provided scope to manage the Renderer lifecycle across test methods.  Creating and preparing the Renderer (which loads Android framework resources) is expensive and seems to be constant anyway.

While I'm a bit wary of making Paparazzi into a @ClassRule, I'm also not 100% sold on this implementation.  Feedback requested.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -82,7 +82,8 @@ class Paparazzi @JvmOverloads constructor(
   private val appCompatEnabled: Boolean = true,
   private val maxPercentDifference: Double = 0.1,
   private val snapshotHandler: SnapshotHandler = determineHandler(maxPercentDifference),
-  private val renderExtensions: Set<RenderExtension> = setOf()
+  private val renderExtensions: Set<RenderExtension> = setOf(),
+  private val managedRendererScope: RendererScope? = null,
 ) : TestRule {
   private val THUMBNAIL_SIZE = 1000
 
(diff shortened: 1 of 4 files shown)


mattprecious (approved):
I don't have the context to comment on this approach vs others specifically, but +1 to caching things in general. Is the renderer the bulk of initializing a Paparazzi instance? Would we want to do more in the future and be stuck due to the API of this first pass at caching?

jrodbx:
> Is the renderer the bulk of initializing a Paparazzi instance?

Yea, it loads all framework resources then initializes the Bridge instance, which loads the native layout lib.

> Would we want to do more in the future and be stuck due to the API of this first pass at caching?

Hmm, not sure, probably not?  Could you elaborate on how we might be stuck?

mattprecious:
> > Would we want to do more in the future and be stuck due to the API of this first pass at caching?
> 
> Hmm, not sure, probably not? Could you elaborate on how we might be stuck?

Not really stuck, just having to make breaking API changes if we want to go beyond a managed renderer.

swankjesse (approved):
This is awesome. It gets close to something even more powerful, which is a global RendererScope

chrisbanes:
Just as a data point, using this fixed my memory and performance issues with paramaterized tests 🥳 . A simple static: `val GlobalRendererScope by lazy { RendererScope() }` was all I needed.

benkay:
Just gave this a spin - definitely looks like a nicer approach than the ClassRule one!
It's a little faster, but I was still finding it significantly slower than the old approach. Running ~300 tests on this branch still takes about 5 minutes, whereas the ClassRule branch takes 20 seconds (10 seconds of which are setup and before the initial test runs).
Most of this seems to be setting up the ByteBuddy interceptors, which is taking anywhere between 300ms to over a second per test.

I have an experimental branch which only does this once per RendererScope. It also caches RenderSessions, which further takes each test from ~80-100ms to ~30 - 50ms on my machine https://github.com/citymapper/paparazzi/tree/benkay/renderer-perf

jrodbx:
Superseded by #390

PR #378, opened 2022-03-03
Prepend paths with file:// in error output for ergonomics

Egorand:
This allows clicking on the path URLs in Android Studio:

Before

image

After:

Screen Shot 2022-03-03 at 10 51 36 AM

Unfortunately, as you can see, AS trips up on URLs that contain "[]", but that feels like an AS bug that I can report. This should work fine for simple paths.


jrodbx:
Thanks!

PR #379, opened 2022-03-04
Use layoutlib SNAPSHOTs part 2

chrisbanes:
Follow on from #377

I just realised that I didn't update the plugin or library to actually use the SNAPSHOT versions. This PR does that my tweaking the layoutlib version which is dynamically added as a dependency.

I also removed hardcoded FQN references to the layoutlib in :paparazzi. Using `project` dependencies means that it automatically uses the correct layoutlib version, and doesn't require the SNAPSHOTs to already be pre-pushed to Sonatype.


chrisbanes:
Ah, the `unzip` configuration can't work as a project dependency right now, as it doesn't output the jar. Closing this for now.

PR #390, opened 2022-03-18
Make Renderer static to share across test suite

jrodbx:
Closes https://github.com/cashapp/paparazzi/issues/369, https://github.com/cashapp/paparazzi/pull/247, https://github.com/cashapp/paparazzi/pull/376

In chatting with @JakeWharton, he convinced me that since the test process should die after a run, we mind as well just make the Renderer static and simplify things.

In the process of debugging this, I noticed that DelegateManager does hold onto delegates and creates seemingly additional delegates between snapshot calls, but that's a different problem.  TODO: profile and show this change doesn't exacerbate that issue.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -88,8 +88,6 @@ class Paparazzi @JvmOverloads constructor(
   private val THUMBNAIL_SIZE = 1000
 
   private val logger = PaparazziLogger()
-  private lateinit var sessionParamsBuilder: SessionParamsBuilder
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

chrisbanes:
Any ETA to when this will land? 🥹

prfarlow1:
>After that, I want to cherry pick this and check to see how much improvement we get before merging.

Does it seem like this PR improves loading time?

jrodbx:
> > After that, I want to cherry pick this and check to see how much improvement we get before merging.
> 
> Does it seem like this PR improves loading time?

Indeed!

JakeWharton (approved):

PR #395, opened 2022-03-20
Update default SDK path on Linux

### LINKED ISSUE #394: macOS Android SDK path used by default on Linux
If neither`ANDROID_SDK_ROOT` nor `ANDROID_HOME` are defined as environment variables, then the fallback is the default SDK location based on the current OS. Right now, this fallback only accounts for Windows or macOS, which results in Linux usually failing when trying to use the macOS location.

To improve default behavior on Linux, it should be pretty straightforward to default to `{home}/Android/Sdk` instead (I'll open a PR shortly)

https://github.com/cashapp/paparazzi/blob/2543296badfd9297ccf805ea072e5c97d7751651/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt#L60-L64

alexvanyo:
Updates the default path logic to use `/Android/Sdk` on Linux, mirroring the logic here: https://github.com/cashapp/paparazzi/blob/2543296badfd9297ccf805ea072e5c97d7751651/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt#L109-L114

This fixes #394

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
@@ -59,8 +59,10 @@ private fun androidSdkPath(): String {
   val osName = System.getProperty("os.name").lowercase(Locale.US)
   val sdkPathDir = if (osName.startsWith("windows")) {
     "\\AppData\\Local\\Android\\Sdk"
-  } else {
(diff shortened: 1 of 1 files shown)


jrodbx (approved):
Thanks!

PR #399, opened 2022-03-28
Mapping overlay colors by hashcode

nak5ive:
View IDs can change with outside refactoring.
Concatenating simple view name and the accessible text could stabilize the overlay color selection.

Open to suggestions!


paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -41,20 +38,24 @@ internal object RenderSettings {
   private val colorMap = mutableMapOf<Int, Color>()
 
   fun getColor(view: View): Color {
-    if (view.id == View.NO_ID) {
-      view.id = generateViewId()
-    }
-    return getColor(view.id)
+    val hashCode = "${view::class.simpleName}(${view.iterableTextForAccessibility})".hashCode()
```
jrodbx, line 41: what does `iterableTextForAccessibility` do here?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -41,20 +38,24 @@ internal object RenderSettings {
...
+    return getColor(hashCode)
```
jrodbx, line 42: nit: push the string key calculated on line 41 here...

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -41,20 +38,24 @@ internal object RenderSettings {
...
   }
 
-  private fun getColor(viewId: Int): Color {
-    return colorMap.getOrPut(viewId) {
-      nextColor(viewId).withAlpha(DEFAULT_RENDER_ALPHA)
+  private fun getColor(hashCode: Int): Color {
```
jrodbx, line 45:
```suggestion
  private fun getColor(key: String): Color {
```

jrodbx, line 45: ...and calculate hashCode in here

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
```diff
@@ -41,20 +38,24 @@ internal object RenderSettings {
...
-  private fun nextColor(viewId: Int): Color {
-    return DEFAULT_RENDER_COLORS[viewId  % DEFAULT_RENDER_COLORS.size]
+  private fun nextColor(hashCode: Int): Color {
+    return DEFAULT_RENDER_COLORS[colorIndex(hashCode)]
+  }
+
+  private fun colorIndex(hashCode: Int): Int {
+    val size = DEFAULT_RENDER_COLORS.size
+    val i = hashCode % size
+    return if (i < 0) i + size else i
```
jrodbx, line 58: i know this is general hashcode practice, but can the hash code of a String ever be negative in practice?

if not, let's inline back to how it was?

jrodbx (changes requested):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
jrodbx: ah, reread your description.  so two different TextViews would be colored by different accessible text?  But if they're blank or similar, the tiebreaker is moot.  Is there a better tiebreaker?  Or maybe punt for a future pr?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
JakeWharton:
```
jshell> "jrodbx".hashCode()
$1 ==> -1151589837
```

paparazzi/src/main/java/app/cash/paparazzi/accessibility/RenderSettings.kt
nak5ive: You are correct though, if more than one of the same views in the test has the same content, then they will be rendered with the same overlay color. But, i'm not sure it's worth fighting duplicates there since we're already cutting the max int-range down to 8 possible colors.

I guess the idea is to derive a generally-unique color that won't change if it's hierarchy changes within the layout.

jrodbx (approved):

PR #402, opened 2022-03-31
Updates CI to publish all publications to Maven Central.

rharter:
Fixes #401

PR #403, opened 2022-03-31
Publish plugin marker on snapshot builds

### LINKED ISSUE #401: Missing Gradle plugin marker publication
It looks like since November, the Gradle plugin marker is missing from [the snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/app/cash/paparazzi/app.cash.paparazzi.gradle.plugin/).  This means that the plugin is no longer available using the new `plugins` block syntax.

It looks like this was caused by [this](https://github.com/cashapp/paparazzi/commit/8438bf63ed823c93ff306e6eb516804afaa55c61) change.  

Instead of only publishing specific publications, however, I think this could be fixed by changing that update to use the `publishAllPublicationsToMavenCentralRepository` task, which will publish all publications, but only to Maven Central.

jrodbx:
Closes #401

cc: @rharter

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
           java-version: 11
 
       - name: Publish Artifacts
-        run: ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publishPluginMavenPublicationToMavenCentralRepository
+        run: ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publishPluginMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publishPaparazziPluginMarkerMavenPublicationToMavenCentralRepository
         env:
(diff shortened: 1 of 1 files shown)


oldergod (approved):

jrodbx:
> Might be a little more succinct to make the Gradle plugin publish all, IMHO, but otherwise this looks good to me.

Unfortunately, layoutlib binaries need to be deployed *before* paparazzi binaries, per: https://github.com/cashapp/paparazzi/pull/402#discussion_r839877434

I wonder if it would make sense for a feature request to be made to the vanniktech plugin such that. `publishMavenPublicationToInternalRepository` could automatically handle plugin marker deploys if detected that the module being deployed is a plugin module?

jrodbx:
https://oss.sonatype.org/content/repositories/snapshots/app/cash/paparazzi/app.cash.paparazzi.gradle.plugin/1.0.0-SNAPSHOT/

🎉

PR #408, opened 2022-04-05
Fixes issue causing `HtmlReportWriter` to ignore configured project.buildDir.

rharter:
We use a customized `project.buildDir` in our project, and while the plugin reads it appropriately to be able to print a string to the console, it didn't pass it to the HtmlReportWriter.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -145,6 +145,7 @@ class PaparazziPlugin : Plugin<Project> {
           override fun execute(t: Task) {
             test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
             test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+            test.systemProperties["paparazzi.test.reportDir"] = reportOutputDir.get()
             test.systemProperties.putAll(paparazziProperties)
           }
         })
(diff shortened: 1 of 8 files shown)


rharter:
Error appears to be a transient network error. Mind poking it again?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -145,6 +145,7 @@ class PaparazziPlugin : Plugin<Project> {
           override fun execute(t: Task) {
             test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
             test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+            test.systemProperties["paparazzi.test.reportDir"] = reportOutputDir.get()
```
jrodbx, line 148: Mind writing a plugin test to a custom build directory setup?

jrodbx (approved):
Ideally, this would be plumbed via the Environment class and not use test task system properties (that's meant for test/record/verify) switching only, but this can be cleaned up in a follow up PR.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
rharter: Fixed in 8c92dbd

chrisbanes:
Any movement on this? Looks like CI needs to be kicked off

rharter:
Just wanted to bump this.  I see there are conflicts due to the project restructure, but I've also got other open PRs that haven't moved and would rather not spend my time updating this one if there isn't a chance of action on it.

PR #421, opened 2022-05-03
Invert logic for selecting SnapshotHandler

chrisbanes:
Currently the private `determineHandler()` selects the `SnapshotHandler` based on the `paparazzi.test.verify` system property, which the Gradle tasks set. Unfortunately this forces users to use the `verifyPaparazzi` task to perform snapshot verification. Using `testDebug` (or `check`, etc) will actually perform a recording instead which feels counter intuitive.

This PR flips that logic around so that it checks for the `paparazzi.test.record` property instead. This results in `test` tasks performing a verification, and users needing to use the `recordPaparazzi` task to perform the record. The `verifyPaparazzi` task could probably be removed after this PR but I'll leave that for now.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -571,14 +571,14 @@ class Paparazzi @JvmOverloads constructor(
     /** The choreographer doesn't like 0 as a frame time, so start an hour later. */
     internal val TIME_OFFSET_NANOS = TimeUnit.HOURS.toNanos(1L)
 
-    private val isVerifying: Boolean =
-      System.getProperty("paparazzi.test.verify")?.toBoolean() == true
+    private val isRecording: Boolean =
+      System.getProperty("paparazzi.test.record")?.toBoolean() == true
 
     private fun determineHandler(maxPercentDifference: Double): SnapshotHandler =
-      if (isVerifying) {
-        SnapshotVerifier(maxPercentDifference)
-      } else {
+      if (isRecording) {
         HtmlReportWriter()
+      } else {
+        SnapshotVerifier(maxPercentDifference)
       }
   }
 }


jrodbx:
> Unfortunately this forces users to use the verifyPaparazzi task to perform snapshot verification. Using testDebug (or check, etc) will actually perform a recording instead which feels counter intuitive.

Not unfortunate; this is by design!  => https://github.com/cashapp/paparazzi#tasks

test => updates a web page (index.html) showing development progress of a view (think of it as a potential alternative to LayoutInspector, although atm still very MVP)
verify => verify stuff
record => record stuff

I think the underlying problem is HtmlReportWriter currently has a bit of extra side-effect logic, which is confusing.  That needs some cleanup/refactoring.

chrisbanes:
It might be by design but I still think it's counter intuitive.

Can you point to any other library or tool which uses the test task for development logging? I can't think of one.

jrodbx:
> Can you point to any other library or tool which uses the test task for development logging? I can't think of one.

Not per se, but sharing more about the design, the experience was modeled after the facebook screenshot plugin which has explicit record/verify tasks.

```
record<App Variant>ScreenshotTest - Installs and runs screenshot tests, then records their output for later verification
run<App Variant>ScreenshotTest - Installs and runs screenshot tests, then generates a report
verify<App Variant>ScreenshotTest - Installs and runs screenshot tests, then verifies their output against previously recorded screenshots
```

The issue I have with having verification be implicit in "test", is that since a record would need to happen before one could verify, an initial run of "test" also seems non-intuitive.

I had considered a more explicit task name (e.g., devPaparazzi${VARIANT}, or maybe just "run" as in Facebook's example), but had concluded at the time that "record" and "verify" were enough and the dev experience could simply be achieved by "test".

I'm open to replacing "test" with something else, if it's really unintuitive (I know Ryan Harter shared a similar viewpoint), but I'm not inclined to have "test" act as "verify", given the above reasons.

alexvanyo:
I agree with Chris that the current behavior here is unintuitive due to the strong expectations that `test` is performing the validation, and that's built into tools surrounding running tests as well.

Running an individual test using the UI in Android Studio will run the `test` task, and with the current behavior, those tests will always pass.

> The issue I have with having verification be implicit in "test", is that, since a record would need to happen before one could verify, an initial run of "test" would then become non-intuitive.

Could that be solved with a more helpful error message pointing to potential causes of why the goldens are missing? Perhaps something along the lines of "No golden found to validate against! Have you ran `recordPaparazzi` to generate the expected snapshots? If this is running in CI, are the snapshots checked in to version control?"

rharter:
I just want to add my voice here and say that I agree with @chrisbanes on this one.  I would argue that the Facebook screenshot testing library does things poorly, due to other constraints, since they do diffing of screenshots on the host machine, not within the test on the emulator.

The `test` task is used to validate all of the assertions in your test code, and it doesn't make sense for screenshot tests to be any different.

It makes sense to me that a separate `record*` task might exist to update the reference images during your workflow, and you could use that or a `dev*` task while writing your screenshot tests to manually verify them.

In my opinion this behavior changes the meaning and guarantees of the `test` task, since now running `test` is no longer sufficient to validate that your tests pass.

jrodbx:
Closing per the discussion.  Can always revisit in the future.

liutikas:
As a note, in AndroidX, we are writing our own plugin and avoiding paparazzi gradle plugin in big part because of this default of not doing verification by default.

liutikas:
Our implementation will be verify by default, test tasks will also output actual images to an output directory, and we'll add a separate task of `updateScreenshotGoldens` that will copy from that output directory and put them into the golden images directories.

jrodbx:
> As a note, in AndroidX, we are writing our own plugin and avoiding paparazzi gradle plugin in big part because of this default of not doing verification by default.

In big part?  Instead of invoking the verify task, the preference is to rewrite the entire plugin.  Am i missing something more here? Feels like an interesting hill to justify such level of effort and maintenance.

FWIW, since this default task behavior appears desirable to some, I think adding a Gradle property which defaults to current behavior but otherwise can be set to verify by default, would be a good compromise.  Thoughts?

chrisbanes:
Yes, the current task naming makes little sense to me (but this is my PR so maybe I'm biased).

We've written logic to avoid this issue, so it's less of an issue for us right now.

rharter:
FWIW I have a custom SnapshotHandler in our wrapper rule that verifies by default, and records if there is an env variable set from a Gradle property. We also use our own differ, so a custom snapshot handler is already there regardless.

rharter:
As for thoughts on your proposal, I think that would be reasonable. That's already basically what the tasks do, by setting the env variables to identify verify or record.

I would, however, still consider inverting the selection: verify by default and record based on the property. The fact that this is billed as a testing library **but doesn't test by default** seems dangerous and confusing.

My understanding of the reasoning for the current default behavior is that you use it as an iterative development tool, but that's not how it's advertised or proposed. Therefore, IMHO, it should act like a test tool out of the box, with the option to change it's behavior into an iterative development tool if desired.

isaac-udy-anz:
I'm new to Paparazzi, and I am also confused by this logic. I also do not think that the documentation is clear. 

> Runs tests and generates an HTML report at sample/build/reports/paparazzi/ showing all test runs and snapshots.

To me "runs tests" means "run and verify tests". I can't think of another library where "running" a failing test would cause the test to pass. 

I believe that inverting the selection as described in this PR would be of benefit.

PR #422, opened 2022-05-04
Test MarginLayoutParams vs LinearLayout.LayoutParams

swankjesse:


swankjesse:
This test shows that `LinearLayout.LayoutParams` has different behavior from `ViewGroup.MarginLayoutParams` when both are used in a `LinearLayout` with Paparazzi.

I believe both types have the same behavior on real devices.

It’s unlikely that the root cause of this is Paparazzi; it’s much more likely that layoutlib has the original bug.

I’m unsure of next steps here. We could report it upstream to layoutlib? And we should advise developers to use `LinearLayout.LayoutParams` when configuring a `LinearLayout` as that works consistently across platforms.

swankjesse:
No action to take in Paparazzi on this.

PR #427, opened 2022-05-10
Test showing resolveAttribute() doesn't resolve dimensions

swankjesse:
BridgeContext.resolveThemeAttribute() doesn't handle dimension
attributes in the same way as real layouts.

The expected outputs should show three toolbars with different
colors and the same heights. Instead we see the one sized by
calling resolveAttribute() has height 0.


swankjesse:
This is a layoutlib bug. I don’t think we can/should fix directly in Paparazzi.

swankjesse:
Reported to AOSP
https://issuetracker.google.com/issues/232007236

PR #431, opened 2022-05-26
Introduce LeakWatcherRule to monitor Paparazzi leaks

jrodbx:
POC to help figure out situations like Compose Runtime leaks


pyricau:
Love it

PR #438, opened 2022-05-31
Detect if Android platform is missing and throw a more helpful exception

### LINKED ISSUE #419: Add helpful error message when missing platform
I see that you already don't swallow FileNotFound exception when platform version is missing #224. It might be overkill, but maybe helpful to some to catch and rethrow with an additional error message to install requested version:

`Missing platform version 30. Install with $ sdkmanager --install "platforms;android-30"`

luis-cortes:
Resolves #419


jrodbx (approved):

tir38:
❤️

Issue #105, opened 2019-10-31
AGP library plugin requirement


ming13:
The current version of the `PaparazziPlugin` requires to have `com.android.library` declared. Is it intentional? I’m asking because I’ve tried to use it with a usual application module and saw the plugin complaint.

```
The Android Gradle library plugin must be applied before the Paparazzi plugin.
```

jrodbx:
That is correct.  This is because the merged resources for library modules are in xml, while for application modules, they are transformed to a binary flat format, which Paparazzi is currently unable to parse.

In a future release, we hope to be able to support app modules.  Workaround for now would be to move the layout resources or custom views into a separate library module and apply the plugin in that module.

PR #301, opened 2021-11-04
Don't require explicit Gradle plugin ordering

TWiStErRob:
Fixes #297.

PR #465, opened 2022-06-18
Update README.md

derandecker:
Based on my own experience, being able to use this library with the main app module would be expected without much fuss. I tried and finally realized what the error message was telling me, and along with this issue: [https://github.com/cashapp/paparazzi/issues/105](https://github.com/cashapp/paparazzi/issues/105) I was able to determine that Paparazzi does not support app modules. It should be made clear right up front so that no one else wastes their time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Paparazzi
 ========
 
-An Android library to render your application screens without a physical device or emulator.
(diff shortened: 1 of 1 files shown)


jrodbx:
> It should be made clear right up front so that no one else wastes their time.

💯  And it looks like I forgot to follow up on [my comment to this earlier PR](https://github.com/cashapp/paparazzi/pull/301/files#r752877840).  Sorry for the wasted time!

PR #480, opened 2022-06-30
Allow for custom appCompatViewInflater #274

sebastienrouif:
and thus support for MaterialComponents or other custom UI library

It would be great to resolve the viewInfalter from the theme but we don't have a context before constructing Paparazzi. We could resolve it in the test function but it would add a lot of boilerplate. And we can't resolve it from within Parazzi as we need to obtain theme attribute as done in [AppCompatDelegateImpl](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:appcompat/appcompat/src/main/java/androidx/appcompat/app/AppCompatDelegateImpl.java;l=1526?q=AppCompatDelegateImpl) 

we could maybe remove the `appCompatEnabled` and have a sealed class for the `appCompatViewInflaterClassName` but I didn't want to change the Paparazzi API

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -696,7 +696,6 @@ class PaparazziPluginTest {
     assertThat(snapshotImage).isSimilarTo(goldenImage).withDefaultThreshold()
   }
 
-  @Test
   @Ignore
   fun withMaterialComponents() {
     val fixtureRoot = File("src/test/projects/material-components-present")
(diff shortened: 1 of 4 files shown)


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -696,7 +696,6 @@ class PaparazziPluginTest {
     assertThat(snapshotImage).isSimilarTo(goldenImage).withDefaultThreshold()
   }
 
-  @Test
   @Ignore
```
msfjarvis, line 699: I think you wanna remove the `@Ignore` here rather than the `@Test`

sebastienrouif:
@jrodbx are you ok with the approach? should I put documentation somewhere?

PR #470, opened 2022-06-24
Quote the recommended git lfs track action

PaulWoitaschek:
Else it will glob and just track the individual files

PR #492, opened 2022-07-08
Add quotation marks to the git lfs track command in ReadMe

fredliao123:
If you run the current command `git lfs track **/snapshots/**/*.png` in the Git LFS section, it returns this error:
```
zsh: no matches found: **/snapshots/**/*.png
``` 
Added quotation marks to the `**/snapshots/**/*.png` part so it actually adds this into the `.gitattributes`


jrodbx:
Closed by https://github.com/cashapp/paparazzi/pull/470

PR #499, opened 2022-07-14
PrepareResourcesTask - Use PathSensitivity.ABSOLUTE for platformDataRoot 

### LINKED ISSUE #495: IllegalStateException: Failed to init Bridge caused by UnsatisfiedLinkError: Can't load library
**Description**
On our CI we have "intermittent" failures on Paparazzi, the error is : 
```
java.lang.IllegalStateException: Failed to init Bridge.
	at app.cash.paparazzi.internal.Renderer.prepare(Renderer.kt:83)
	at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:158)
	at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:122)
	at app.cash.paparazzi.agent.AgentTestRule$apply$1.evaluate(AgentTestRule.kt:17)
	at com.google.testing.junit.testparameterinjector.PluggableTestRunner$ContextMethodRule$1.evaluate(PluggableTestRunner.java:420)
	at ...
```
Which is caused by
```
Jul 12, 2022 5:16:01 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: broken: Native layoutlib failed to load
java.lang.UnsatisfiedLinkError: Can't load library: /agent/workspace/Project-ABC/label/awstest_android/my_variants/unitTestDebug/.gradle/caches/transforms-3/57eded7ec1aadfb67857557a846e1fd3/transformed/layoutlib-native-linux-2021.1.1-573f070/data/linux/lib64/libandroid_runtime.so
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2630)
	at java.base/java.lang.Runtime.load0(Runtime.java:768)
	at java.base/java.lang.System.load(System.java:1837)
	at com.android.layoutlib.bridge.Bridge.loadNativeLibraries(Bridge.java:720)
	at com.android.layoutlib.bridge.Bridge.loadNativeLibrariesIfNeeded(Bridge.java:695)
	at com.android.layoutlib.bridge.Bridge.init(Bridge.java:178)
	at app.cash.paparazzi.internal.Renderer.prepare(Renderer.kt:84)
```

The `libandroid_runtime.so` is definitely not present in the `/agent/workspace/Project-ABC/.../lib64/` folder. 

What's also interesting, the `/agent/workspace/Project-ABC/` folder is not where the build is supposed, this is another build project (running on another CI agent). That's probably why the `.so` file does not exists.

**Additional information:**
- Paparazzi Version: 1.0.0
- OS: Linux 5.4.0-1080-aws
- Compile SDK: 31
- Gradle Version: 7.4.2
- Android Gradle Plugin Version: 7.2.1

eboudrant:
Fix #495, since `platformDataRoot` is a path contained in gradle home it's safer to use absolute path for cache input.

Here the build input comparison after fix when using steps to repro on #495: 

image

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -50,7 +50,7 @@ abstract class PrepareResourcesTask : DefaultTask() {
   abstract val mergeAssetsOutput: DirectoryProperty
 
   @get:InputDirectory
-  @get:PathSensitive(PathSensitivity.RELATIVE)
+  @get:PathSensitive(PathSensitivity.ABSOLUTE)
   abstract val platformDataRoot: DirectoryProperty
 
   @get:Input


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -50,7 +50,7 @@ abstract class PrepareResourcesTask : DefaultTask() {
   abstract val mergeAssetsOutput: DirectoryProperty
 
   @get:InputDirectory
-  @get:PathSensitive(PathSensitivity.RELATIVE)
+  @get:PathSensitive(PathSensitivity.ABSOLUTE)
```
liutikas, line 53: That is a bad change, this would make all remote build cache hits have a miss. If anything, we should be using `PathSensitivity.NONE` since we do not really care where the directory is, other than the contents of it.

eboudrant: @liutikas sorry for delay, I had a doubt and I tried again: using `PathSensitivity.NONE` would fails on the same error. The task `preparePaparazziReleaseResources` is still resolved from the cache from the steps described on #495: 

image

eboudrant: I think the reason of this task not working well with Gradle cache is because the actual jar file is downloaded and stored via the Gradle dependency manager, it's not an output of the task. The values stored into the Gradle cache for this task is the path to the jar file. Using `ABSOLUTE` will cause some cache-miss but will fix this issue.

liutikas: @jrodbx and I discussed how to handle `platformDataRoot` properly. I'll create a PR on Monday with a fix that should address this issue without use of ABSOLUTE normalizer.

jrodbx:
Obsoleted by https://github.com/cashapp/paparazzi/pull/536

PR #501, opened 2022-07-19
Paparazzi.js fixes

ehoogend:
While using Paparazzi I ran into a few issues regarding paparazzi.js when running . These are fixed in this MR:
- Syntax error on line 16
- `Shot.TestMethodRegex` was failing to parse `this.package`, `this.clazz`, `this.method` from `test` properly.
- The path to the scripts was incorrect, pointing to `${runId}/run.js` instead of `runs/${runId}.js`
image

This fixes paparazzi.js and ensures the HTML report can be opened without errors.

### DIFF
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -12,7 +12,7 @@ class Shot {
   constructor(name, test) {
     this.name = name;
     this.test = test;
-    
+
     [, this.package, this.clazz, this.method] = Shot.TestMethodRegex.exec(test);
 
     this.runs = [];
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -13,7 +13,7 @@ class Shot {
     this.name = name;
     this.test = test;
     
-    [this.package, this.clazz, this.method] = Shot.TestMethodRegex.exec(test);
+    [, this.package, this.clazz, this.method] = Shot.TestMethodRegex.exec(test);
 
     this.runs = [];
   }
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -181,7 +181,7 @@ class PaparazziRenderer {
         continue;
       }
       // The js loading is async so the rendering can happen in the next refresh
-      this.loadRunScript(`${runId}/run.js`);
+      this.loadRunScript(`runs/${runId}.js`);
 
       this.render(new Run(runId, window.runs[runId]));
 
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -19,7 +19,7 @@ class Shot {
   }
 
   static get TestMethodRegex() {
-    return /^(.*)\.(.*)#(.*)\[/;
+    return /^(.*)\.(.*)#([^\[]*)\[?.*$/;
   }
 
   addRun(runId, file, timestamp) {
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -19,7 +19,7 @@ class Shot {
   }
 
   static get TestMethodRegex() {
-    return /^(.*)\.(.*)#([^\[]*)\[?.*$/;
+    return /^(.*)\.(.*)#(.*)$/;
   }
 
   addRun(runId, file, timestamp) {
--- a/paparazzi/src/main/resources/paparazzi.js
+++ b/paparazzi/src/main/resources/paparazzi.js
@@ -12,14 +12,14 @@ class Shot {
   constructor(name, test) {
     this.name = name;
     this.test = test;
-
-    [, this.package, this.clazz, this.method] = Shot.TestMethodRegex.exec(test);
+    
+    [this.package, this.clazz, this.method] = Shot.TestMethodRegex.exec(test);
 
     this.runs = [];
   }
 
   static get TestMethodRegex() {
-    return /^(.*)\.([^.]*)#([^.]*)$/;
+    return /^(.*)\.(.*)#(.*)\[/;
   }
 
   addRun(runId, file, timestamp) {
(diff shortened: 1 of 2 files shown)


fcduarte:
@ehoogend thanks for contributing! I see on Chrome's console an infinite loop when trying to find the js files.

Did you run `./gradlew clean :paparazzi:check` to check your changes? I'm seeing an error on Chrome's console when I apply it:

````
paparazzi.js:16 Uncaught TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
    at new Shot (paparazzi.js:16:47)
    at PaparazziRenderer.render (paparazzi.js:163:16)
    at PaparazziRenderer.renderAll (paparazzi.js:186:12)
    at PaparazziRenderer.refresh (paparazzi.js:200:10)
````

I think it's due to the regex change, could you explain it? Thanks!

**Comment by ehoogend** [comment]
@fcduarte Actually that's the same error I get _without_ the fixes in this MR. Are you sure you applied the fix?

I ran  `./gradlew clean :paparazzi:check` successfully.

Regarding the regex changes: when running paparazzi 1.0 I get a `Shot.test` argument as follows:
`com.sampleApp.elements.snapshot.tests.Pixel5SnapshotTests#screenShotTests[com.sampleApp.elements_null_Buttons_Primarybutton_Enabled]`
This is unpacked by `TestMethodRegex` into [this.package, this.clazz, this.method] the following way:
image

The first group captured is the package, second group the Class, third group the method. Everything after the first `[` is ignored.

FYI: the existing regex didn't match for me at all:
image

**Comment by fcduarte** [comment]
Oh now I see the issue. The current implementation doesn't parse your test value -- `com.sampleApp.elements.snapshot.tests.Pixel5SnapshotTests#screenShotTests[com.sampleApp.elements_null_Buttons_Primarybutton_Enabled]` -- but running with your changes it does. On the flip side, then your changes breaks the regular values like the one from main Paparazzi module like `app.cash.paparazzi.PaparazziTest#drawCalls`. So we need a regex that would work on all cases. Also I'm trying to understand which type of test would add that string inside the brackets, could you provide more info? Thanks!

**Comment by ehoogend** [comment]
Ah the difference is probably because I'm running parameterized tests with `com.google.testing.junit.testparameterinjector.TestParameterInjector` ! I'm running paparazzi screenshot tests for all Showkase components in the following way:
image

I'll push a change that handles both cases (with and without parameters)

**Comment by ehoogend** [comment]
@fcduarte The updated regex now handles both cases without errors:

image

image

**Comment by fcduarte** [comment]
@ehoogend thanks for checking again! I wish there was an easy way to unit test this class. The regex looks good but when running the JS code it's still not creating the `Shot` class correctly. If you open the web page, go to console and try:

````
var shot = new Shot(undefined, "com.sampleApp.elements.snapshot.tests.Pixel5SnapshotTests#screenShotTests[com.test.test_a]")
console.log(shot)
````

You'll see that `clazz`, `method` and `package` are not parsed correctly. I don't know much about JS but seems like keeping the code on line 16 solves this. Could you give another look? Thanks!

**Comment by ehoogend** [comment]
Ah you're right! I'm also no JS expert and indeed it turns out the issue in this case was there and not in the regex :) 

Retested it and it should be fixed now:
With parameterized test:
image

Without parameterized test:
image

File: paparazzi/src/main/resources/paparazzi.js [hunk 1]
```diff
@@ -12,14 +12,14 @@ class Shot {
   constructor(name, test) {
     this.name = name;
     this.test = test;
-
+
```
fcduarte, line 15: Nit: revert those lines.

ehoogend: Done!

fcduarte:
I tested locally here and looks good! 🎉

fcduarte:
Hey @ehoogend, we're almost there :) We discussed the PR internally today and there's a couple extra things we wanted to figure out:

1. First, it was my bad on `method` parsing. We want to display what's inside the brackets as well so if `test` is `com.sampleApp.elements.snapshot.tests.Pixel5SnapshotTests#screenShotTests[com.sampleApp.elements_null_Buttons_Primarybutton_Enabled]` then `method` should be `screenShotTests[com.sampleApp.elements_null_Buttons_Primarybutton_Enabled]`
2. We're struggling to reproduce your use case internally. If you take a look on `sample/src/test/java/app/cash/paparazzi/sample/TestParameterInjectorTest.kt` it's already working with `TestParameterInjector` and it produces what we want, for instance, `simpleWithTheme[NEXUS_5,LIGHT]`. But we couldn't reproduce one like yours -- `screenShotTests[com.sampleApp.elements_null_Buttons_Primarybutton_Enabled]` -- so request is: could you create a new test at `TestParameterInjectorTest.kt` that would demonstrate the need for a regex fix?

The other JS change looks good though, up to you if you want to fix both on this PR or to separate it.

Thanks!

ehoogend:
@fcduarte 

Thanks for the response!

1. That's actually a lot simpler! Changed this already.
2. Is slightly trickier. It will require quite some changes to the sample including adding Jetpack Compose to demonstrate the use case fully , which I can do. In short, I used [this gist](https://gist.github.com/alexvanyo/c4d2b6530f860f3c89a33324d4152ae8) as an example to generate parameterized tests for all my `@ShowkaseComponent`-annotated previews of composables.

I have this `ComponentPreview` class:

`class ComponentPreview(val showkaseBrowserComponent: ShowkaseBrowserComponent) {
    val content: @Composable () -> Unit = showkaseBrowserComponent.component
    override fun toString(): String = showkaseBrowserComponent.componentKey
}`

Which overrides the `toString` function to provide `ShowkaseBrowserComponent.componentKey` as part of the test name. This includes the package and function name of the composable, as well as the Showkase `styleName`.

Of course I could (and am now realizing probably should) change this to be shorter, for example function name and `styleName` only, but ideally the html report would not break because of dots in the parameters. so I think it's still good to fix the regex.

ehoogend:
@fcduarte Actually it turned out to be quite easy to add a test that illustrates the use case for the regex change :) . I added `amountProviderTest` to the `TestParameterInjectorTest` class to illustrate and enable reproduction of this issue.

jrodbx (approved):
awesome, nice test!

PR #517, opened 2022-07-28
Optional snapshot root directory gradle property

brenpearson:
Allows the consumer of the Paparazzi library to specify where they want the snapshots to be saved. This includes creating a gradle plugin extension for Paparazzi with this property in it, we can add other configurable properties to this extension in future.

This can be set as shown in the consumer's gradle file
```
paparazzi {
    snapshotRootDirectory = file("where/I/want/the/snapshots")
}
```

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -21,7 +21,6 @@ import com.android.build.gradle.BaseExtension
 import com.android.build.gradle.LibraryExtension
 import com.android.build.gradle.tasks.MergeSourceSetFolders
 import com.android.ide.common.symbols.getPackageNameFromManifest
-import java.io.File
 import org.gradle.api.Action
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
@@ -60,10 +59,13 @@ class PaparazziPlugin : Plugin<Project> {
       }
     }
 
-    project.plugins.withId("com.android.library") { setupPaparazzi(project) }
+    project.plugins.withId("com.android.library") {
+      val extension = project.extensions.create("paparazzi", PaparazziPluginExtension::class.java)
+      setupPaparazzi(project, extension)
+    }
   }
 
-  private fun setupPaparazzi(project: Project) {
+  private fun setupPaparazzi(project: Project, extension: PaparazziPluginExtension) {
     val unzipConfiguration = project.setupPlatformDataTransform()
 
     // Create anchor tasks for all variants.
@@ -82,7 +84,7 @@ class PaparazziPlugin : Plugin<Project> {
       val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
 
       val snapshotOutputDir =
-          project.extensions.extraProperties.properties["paparazzi_snapshot_dir"] as? File
+          extension.snapshotRootDirectory.orNull
           ?: project.layout.projectDirectory.dir("src/test/snapshots").asFile
 
       val packageAwareArtifacts = project.configurations.getByName("${variant.name}RuntimeClasspath")
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -84,8 +84,8 @@ class PaparazziPlugin : Plugin<Project> {
       val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
 
       val snapshotOutputDir =
-          extension.snapshotRootDirectory.orNull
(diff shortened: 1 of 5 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -76,7 +77,10 @@ class PaparazziPlugin : Plugin<Project> {
         project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
       val mergeAssetsOutputDir = mergeAssetsProvider.flatMap { it.outputDir }
       val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
-      val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
+
+      val snapshotOutputDir =
+          project.extensions.extraProperties.properties["paparazzi_snapshot_dir"] as? File
```
liutikas, line 82: It should not be an arbitrary value in extraProperties. We should create a paparazzi plugin extension where this value is explicitly specified.

brenpearson: That's a better idea - done now 💯 just need to fix up the unit tests now I think

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -76,7 +79,11 @@ class PaparazziPlugin : Plugin<Project> {
         project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
       val mergeAssetsOutputDir = mergeAssetsProvider.flatMap { it.outputDir }
       val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
-      val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
+
+      val snapshotOutputDir =
+          extension.snapshotRootDirectory.orNull
+          ?: project.layout.projectDirectory.dir("src/test/snapshots").asFile
+      snapshotOutputDir.mkdirs()
```
brenpearson, line 86: Seems like in PrepareResourcesTask to add this as an `InputDirectory` the directory has to exist, so in that case I have to add this to create the snapshot directory here instead of waiting until the task to take/verify the snapshots. Thoughts on if this is fine?

JavierSegoviaCordoba:
Is there any blocker for this PR?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: you went from a Provider to a Provider that's resolved in configuration phase. This is happening too early here I think. I think this would be something like:
```
extension.snapshotRootDirectory.orElse(project.layout.projectDirectory.dir("src/test/snapshots"))
```
and maybe that solves your need to create the dir too.

TWiStErRob: but it might be even better if the default was set up with `snapshotRootDirectory.convention` when the extension is registered:
```
val extension = project.extensions.create("paparazzi", PaparazziPluginExtension::class.java)
extension.snapshotRootDirectory.convention(project.layout.projectDirectory.dir("src/test/snapshots"))
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPluginExtension.kt
```diff
@@ -0,0 +1,8 @@
+package app.cash.paparazzi.gradle
+
+import org.gradle.api.provider.Property
+import java.io.File
+
+interface PaparazziPluginExtension {
+  val snapshotRootDirectory: Property<File?>
```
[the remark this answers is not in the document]
JavierSegoviaCordoba: In Sqldelight I moved to `ConfigurableFileCollection` as a recommendation from @hfhbd to emulate how `srcDirs` works on Java and Kotlin Gradle plugins.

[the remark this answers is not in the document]
TWiStErRob: But this is a single output directory, `srcDirs` are multiple inputs, feels very different.

brenpearson:
@JavierSegoviaCordoba I'm not going to be able to update this within the next couple of weeks. If you have time then by all means, otherwise I'll come back and take a look in probably 2 weeks time 🙏

JavierSegoviaCordoba:
Can I push on this branch or should I create a new PR? @brenpearson commits are preserved in both cases.

JavierSegoviaCordoba:
Updated PR with @TWiStErRob suggestion: https://github.com/cashapp/paparazzi/pull/736

PR #520, opened 2022-07-29
Wire sample project to use realtime plugin sources

jrodbx:
Using Molecule and Wire projects as inspiration, this moves all modules under a common subfolder as a separate project, leaving the sample at the top level.  This allows both the usage of includeBuilds and for all projects to be referenced in the IDE.  It seems that the latter is what makes an includeBuild('..') a non-workable option.

Verified the following as working:
* running sample and paparazzi module tests
* cloning layoutlib repo
* publishing artifacts

Also, updated READMEs


settings.gradle
```diff
@@ -1,13 +1,13 @@
...
-include ':libs:native-macarm'
-include ':libs:native-macosx'
-include ':libs:native-win'
-include ':libs:native-linux'
 
-enableFeaturePreview('TYPESAFE_PROJECT_ACCESSORS')
\ No newline at end of file
+includeBuild('paparazzi') {
+  dependencySubstitution {
+    substitute module('app.cash.paparazzi:paparazzi') with project(':paparazzi')
```
luis-cortes, line 7: `with` is deprecated now. Any reason we can't use `using` instead?

jrodbx: yep, you're right.  will update.

luis-cortes (approved):

PR #521, opened 2022-07-29
Move to settings.gradle for dependency repository configuration

liutikas:
This is the modern way to do it in Gradle as is project isolation compatible.


jrodbx:
The alternative is to use convention plugins to declare them in each project, which I tend to favor since it doesn't clutter the settings.gradle file.  Are there any pros/cons of the two approaches I might be missing?

PR #522, opened 2022-07-30
Move to using Gradle version catalog

liutikas:
Migrating the project to use Gradle version catalog instead of gradle/dependencies.gradle script.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,10 +11,10 @@ buildscript {
   }
 
   dependencies {
-    classpath deps.plugins.kotlin
-    classpath deps.plugins.android
-    classpath deps.plugins.versions
-    classpath deps.plugins.spotless
+    classpath(libs.kotlinGradlePlugin)
+    classpath(libs.androidGradlePlugin)
+    classpath(libs.versionsGradlePlugin)
+    classpath(libs.spotlessGradlePlugin)
 
     // Normally you would declare a version here, but we use dependency substitution in
     // settings.gradle to use the version built from inside the repo.
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,0 +1,41 @@
+[versions]
+agp = "7.1.2"
+bytebuddy = "1.12.10"
+jcodec = "0.2.5"
+kotlin = "1.6.10"
+moshi = "1.13.0"
+
+[libraries]
+# Gradle plugins
+androidGradlePlugin = { module = "com.android.tools.build:gradle", version.ref = "agp" }
+dokkaGradlePlugin = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version = "1.7.10" }
+grgitGradlePlugin = { module = "org.ajoberstar.grgit:grgit-gradle", version= "5.0.0" }
+kotlinGradlePlugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
+mavenPublishGradlePlugin = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.21.0" }
+spotlessGradlePlugin = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.8.0" }
+versionsGradlePlugin = { module = "com.github.ben-manes:gradle-versions-plugin", version = "0.42.0" }
+
+# Regular dependencies
+androidxAnnotations = { module = "androidx.annotation:annotation", version = "1.3.0" }
+androidxComposeCompiler = { module = "androidx.compose.compiler:compiler", version = "1.1.1" }
+androidxLifecycleCommon = { module = "androidx.lifecycle:lifecycle-common", version = "2.4.0" }
+assertj = { module = "org.assertj:assertj-core", version = "3.23.1" }
+bytebuddyAgent = { module = "net.bytebuddy:byte-buddy-agent", version.ref = "bytebuddy" }
+bytebuddyCore = { module = "net.bytebuddy:byte-buddy", version.ref = "bytebuddy" }
+guava =  { module = "com.google.guava:guava", version = "31.0.1-jre" }
+jcodecCore = { module = "org.jcodec:jcodec", version.ref = "jcodec" }
+jcodecJavase = { module = "org.jcodec:jcodec-javase", version.ref = "jcodec" }
+junit = { module = "junit:junit", version = "4.13.2" }
+kotlinBom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
+kotlinxCoroutinesCore = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.6.2" }
+kxml2 = { module = "kxml2:kxml2", version = "2.3.0" }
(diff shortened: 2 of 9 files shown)


settings.gradle
```diff
@@ -1,3 +1,12 @@
+pluginManagement {
+  repositories {
+    mavenCentral()
+    google()
+    gradlePluginPortal()
+    //mavenLocal()
+  }
+}
```
jrodbx, line 8: Revert

build.gradle
```diff
@@ -1,27 +1,16 @@
...
-    classpath deps.plugins.kotlin
-    classpath deps.plugins.android
-    classpath 'com.vanniktech:gradle-maven-publish-plugin:0.20.0'
-    classpath 'org.ajoberstar.grgit:grgit-gradle:5.0.0'
-    classpath 'org.jetbrains.dokka:dokka-gradle-plugin:1.6.21'
-    classpath deps.plugins.versions
-    classpath deps.plugins.spotless
-    classpath "app.cash.paparazzi:paparazzi-gradle-plugin:${versions.releasedPaparazzi}"
-  }
+plugins {
```
jrodbx, line 1: Thanks for contributing!

I'd prefer to keep using the `buildscript {}` here.

Also, can we removing the `[plugins]` section in the toml, and add a `plugins-` suffix to keep the `plugins.kotlin` notation as above.

jrodbx (changes requested):

settings.gradle
liutikas: Reverted back to buildscript.

build.gradle
liutikas: Reverted back to buildscript.

liutikas:
Any thoughts on this?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,41 @@
...
+kxml2 = { module = "kxml2:kxml2", version = "2.3.0" }
+moshiAdapters = { module = "com.squareup.moshi:moshi-adapters", version.ref = "moshi" }
+moshiCore = { module = "com.squareup.moshi:moshi", version.ref = "moshi" }
+moshiKotlinCodegen = { module = "com.squareup.moshi:moshi-kotlin-codegen", version.ref = "moshi" }
+testParameterInjector = { module = "com.google.testparameterinjector:test-parameter-injector", version = "1.8" }
+truth = { module = "com.google.truth:truth", version = "1.1.3" }
+toolsCommon = { module = "com.android.tools:common", version = "27.1.2" }
+toolsLayoutlib = { module = "com.android.tools.layoutlib:layoutlib-api", version = "27.2.2" }
+toolsNinepatch = { module = "com.android.tools:ninepatch", version = "30.1.2" }
+toolsSdkCommon = { module = "com.android.tools:sdk-common", version = "26.6.4" }
```
TWiStErRob, line 40: tip: if you use dashes, you can keep the usages nicely nested

`tools-common = {` will be `api(tools.common)`

I think this is what John meant when he asked for `plugins` prefix.

`androidGradlePlugin` -> `plugins-android=` -> `api(plugins.android)`

liutikas: > tip: if you use dashes, you can keep the usages nicely nested and grouped

Thanks for the suggestion, but I don't see it as a beneficial thing.

> I think this is what John meant when https://github.com/cashapp/paparazzi/pull/522#discussion_r933698236 for plugins prefix.

I chatted offline, and what John wanted was not to use `plugins {}` API and instead use `buildscript {}` to load plugins.

build.gradle
```diff
@@ -11,10 +11,10 @@ buildscript {
...
 
   dependencies {
-    classpath deps.plugins.kotlin
-    classpath deps.plugins.android
-    classpath deps.plugins.versions
-    classpath deps.plugins.spotless
+    classpath(libs.kotlinGradlePlugin)
+    classpath(libs.androidGradlePlugin)
+    classpath(libs.versionsGradlePlugin)
+    classpath(libs.spotlessGradlePlugin)
```
jrodbx, line 17:
```suggestion
    classpath libs.kotlinGradlePlugin
    classpath libs.androidGradlePlugin
    classpath libs.versionsGradlePlugin
    classpath libs.spotlessGradlePlugin
```

same with all the other configurations in the other build.gradles

gradle/libs.versions.toml
jrodbx: > I think this is what John meant when https://github.com/cashapp/paparazzi/pull/522#discussion_r933698236 for plugins prefix.

> Thanks for the suggestion, but I don't see it as a beneficial thing.

Yep, that's correct.  We'd prefer the dashes.

build.gradle
jrodbx: for plugins, let's go with `classpath libs.plugins.spotless`

gradle/libs.versions.toml
```diff
@@ -0,0 +1,41 @@
...
+moshi = "1.13.0"
+
+[libraries]
+# Gradle plugins
+androidGradlePlugin = { module = "com.android.tools.build:gradle", version.ref = "agp" }
+dokkaGradlePlugin = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version = "1.7.10" }
+grgitGradlePlugin = { module = "org.ajoberstar.grgit:grgit-gradle", version= "5.0.0" }
+kotlinGradlePlugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
+mavenPublishGradlePlugin = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.21.0" }
+spotlessGradlePlugin = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.8.0" }
```
jrodbx, line 15:
```suggestion
plugins-spotless = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.8.0" }
```

and similar, i.e., no need to reference Gradle, since that's the build tool that consumes this toml and move `Plugins` suffix to `plugins-` prefix.

jrodbx (changes requested):
another round of feedback, but getting closer

build.gradle
TWiStErRob: I can come up with why for this, but I'm curious what's your answer, why remove the parentheses?

gradle/libs.versions.toml
TWiStErRob: `plugins` is a reserved prefix because of `[plugins]` 😅, but `plugin` works, and makes even more sense.
```
  - Problem: In version catalog libs, alias 'plugins-grgit' is not a valid alias.

    Reason: Prefix for dependency shouldn't be equal to 'plugins'.

    Possible solution: Use a different alias which prefix is not equal to 'bundles', 'plugins' or 'versions'.

    Please refer to https://docs.gradle.org/7.5.1/userguide/version_catalog_problems.html#reserved_alias_name for more details about this problem.
```

fixed in https://github.com/cashapp/paparazzi/pull/571/commits/80bca8d9068dca5ae7c7ce7323431a01a467eace on #571

build.gradle
TWiStErRob: fixed in https://github.com/cashapp/paparazzi/pull/571/commits/d84ebe01d676b24721fb7898d913bcb3644b6964 on #571, but still curious why

Issue #468, opened 2022-06-23
Out of memory exception with a lot of screenshots (compose)


badoualy:
**Description**
I have a test that generates a lot of screenshots (over a 1 000). (One component with a lot of display state possible, tested on 2 devices config, + light/dark mode, the total count is quickly reaching 1 000).

The first ~ 950 or so screenshots are going well, then I get the OOM exception on each one of them.

I'm using the following compose integration with showkase:
https://gist.github.com/alexvanyo/c4d2b6530f860f3c89a33324d4152ae8

In the mean time my solution was to create a shell script to run the test by batch of ~600 tests each, killing the jvm each time between runs.

**Additional information:**
- Paparazzi Version: 1.0.0
- OS: Windows

fcduarte:
@badoualy that's an interesting case .. Do you happen to have a CI machine that runs in a different environment (Linux or MacOS)? This would help to pinpoint if this is an issue contained on Windows. Thanks!

fcduarte:
@badoualy also it'd be great if you could post the stack trace so we can try to figure out where the OOM is coming from, thanks!

jrodbx:
OOMs are extremely hard to repro or triage without some repro, however, if I had to guess, there are two likely causes for this:

1) Compose runtime has some known memory leaks.  This is out of our hands, although we have tried to remedy them somewhat here: https://github.com/cashapp/paparazzi/pull/430.
2) Your test jvm process needs more heap memory.  If you're using Gradle, the canonical way to bump that is in your gradle.properties, e.g.,: `org.gradle.jvmargs=-Xmx8192M`

Jan-Ove:
We also had issues with OutOfMemoryErrors, and increasing the memory via the gradle.properties didn't solve it. I've discovered that it is possible to add jvmArgs via the testOptions block in the Gradle file and that solved our problem.
```kotlin
android {
    testOptions {
        unitTests {
            all {
                it.allJvmArgs = it.allJvmArgs + listOf("-Xmx4192M")
            }
        }
    }
}
```

castorflex:
Note for future people: `jvmArgs = listOf("-Xmx4192M")` should be used instead of `allJvmArgs`, which can cause caching issues. See https://github.com/gradle/gradle/issues/25874

PR #533, opened 2022-08-04
Fix OOM error when a large number of compose snapshots are failing when verifying

rojanthomas:
I've noticed that if you have a large number of compose snapshot tests failing when running the `verifyPaparazziDebug` task, it will eventually lead to an `OutOfMemoryError: Java heap space`

[The following gist contains a test class that will simulate 2000 failing tests being run and leading to OOM errors (depending on heap size)](https://gist.github.com/rojanthomas/a200a1375b699525378d7d0410706575). (I didn't record a baseline so they would all fail).

This fix will ensure that the correct error is presented in the JUnit report & any failure diffs continue to be generated.

This may be related to #468

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -189,17 +189,19 @@ class Paparazzi @JvmOverloads constructor(
   fun <V : View> inflate(@LayoutRes layoutId: Int): V = layoutInflater.inflate(layoutId, null) as V
 
   fun snapshot(name: String? = null, composable: @Composable () -> Unit) {
-    val hostView = ComposeView(context)
-    // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's
-    // CompositionContext, which requires first finding the "content view", then using that to
-    // find a root view with a ViewTreeLifecycleOwner
-    val parent = FrameLayout(context).apply { id = android.R.id.content }
-    parent.addView(hostView, LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT)
-    PaparazziComposeOwner.register(parent)
-    hostView.setContent(composable)
-    snapshot(parent, name)
-
-    forceReleaseComposeReferenceLeaks()
+    try {
+      val hostView = ComposeView(context)
+      // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's
+      // CompositionContext, which requires first finding the "content view", then using that to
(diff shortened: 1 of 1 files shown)


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -189,17 +189,20 @@ class Paparazzi @JvmOverloads constructor(
...
+    try {
+      val hostView = ComposeView(context)
+      // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's
+      // CompositionContext, which requires first finding the "content view", then using that to
+      // find a root view with a ViewTreeLifecycleOwner
+      val parent = FrameLayout(context).apply { id = android.R.id.content }
+      parent.addView(hostView, LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT)
+      PaparazziComposeOwner.register(parent)
+      hostView.setContent(composable)
+      snapshot(parent, name)
```
rojanthomas, line 201: Compose references were not being released if `snapshot(parent, name)` throws an exception i.e. an assertion.

fcduarte:
@rojanthomas I ported your gist to `sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt` and ran `./gradlew :sample:test` and didn't get any OOM with the current Paparazzi's code. Am I missing something?

rojanthomas:
> @rojanthomas I ported your gist to `sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt` and ran `./gradlew :sample:test` and didn't get any OOM with the current Paparazzi's code. Am I missing something?

@fcduarte Apologies, I should be been clearer in the steps to reproduce the OOM error.

1. Make sure you don't have any baseline/golden images generated for the tests in the gist I provided. (From what I've seen, this OOM error only happens if the verification of the snapshot fails many times sequentially)
2. Run `./gradlew :sample:verifyPaparazziDebug`
3. Around the 1400th test failure, you should start to see OOM errors instead of an `AssertionError`.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: let's scope the try block to surround only the snapshot call then.

jrodbx (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
rojanthomas: Makes sense 👍🏽 

Done

PR #535, opened 2022-08-05
Bump publish plugin

jrodbx:


paparazzi/build.gradle
```diff
@@ -72,8 +72,12 @@ subprojects {
       }
     }
```
gabrielittner, line 74: You probably want something like this here to still sign artifacts and add sonatype/central as a publishing repo
```suggestion
    mavenPublishing {
        publishToMavenCentral()
        signAllPublications()
    }
```
Alternatively you can also set to Gradle properties like described here https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md#version-0210-2022-07-11

paparazzi/build.gradle
```diff
@@ -72,8 +72,12 @@ subprojects {
...
 
-    mavenPublish {
-      releaseSigningEnabled = hasProperty('signingKey')
+    plugins.withId('signing') {
+      signing {
+        def signingKey = findProperty('signingKey')
+        def signingPassword = ''
+        useInMemoryPgpKeys(signingKey, signingPassword)
```
gabrielittner, line 79: The plugin has something built in for picking up in memory signing keys you just need to rename your `signingKey` property to `signingInMemoryKey` https://github.com/vanniktech/gradle-maven-publish-plugin#signing

paparazzi/build.gradle
jrodbx: oh, i had done that (the 2nd approach), but looks like i forgot to add it.  that's what I get for late night PRs.  thanks for the sanity check!

oldergod (commented):
lgtm

gamepro65 (approved):

luis-cortes (approved):

PR #536, opened 2022-08-07
Configure native platform transformed path directly in test task

jrodbx:
Based on offline discussion with and suggestion by @liutikas.

Adding the native platform transformed unzipped directory to the `resources.txt` generated by PrepareResourcesTask was a design mistake and prevents that task from being truly cacheable, since it has an absolute file path to the user home directory and is also platform OS dependent.  Instead, we wire that directly as an input to the Test task.

According to @liutikas, this also brings some performance improvement as the transform/unzip process was being performed during configuration time.

Closes https://github.com/cashapp/paparazzi/issues/495 and https://github.com/cashapp/paparazzi/pull/499.

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -26,13 +26,13 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
-import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition
 import org.gradle.api.attributes.Attribute
-import org.gradle.api.internal.artifacts.ArtifactAttributes
+import org.gradle.api.file.FileCollection
 import org.gradle.api.internal.artifacts.transform.UnzipTransform
 import org.gradle.api.logging.LogLevel.LIFECYCLE
 import org.gradle.api.plugins.JavaBasePlugin
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.options.Option
 import org.gradle.api.tasks.testing.Test
@@ -63,7 +63,8 @@ class PaparazziPlugin : Plugin<Project> {
   }
 
   private fun setupPaparazzi(project: Project) {
-    val unzipConfiguration = project.setupPlatformDataTransform()
+    project.addTestDependency()
+    val nativePlatformFileCollection = project.setupNativePlatformDependency()
 
     // Create anchor tasks for all variants.
     val verifyVariants = project.tasks.register("verifyPaparazzi")
@@ -81,13 +82,11 @@ class PaparazziPlugin : Plugin<Project> {
       val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
       val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
 
-      val packageAwareArtifacts = project.configurations.getByName("${variant.name}RuntimeClasspath")
+      val packageAwareArtifacts = project.configurations
+        .getByName("${variant.name}RuntimeClasspath")
         .incoming
         .artifactView {
-          it.attributes.attribute(
-            Attribute.of("artifactType", String::class.java),
-            "android-symbol-with-package-name"
-          )
+          it.attributes.attribute(ARTIFACT_TYPE_ATTRIBUTE, "android-symbol-with-package-name")
         }
         .artifacts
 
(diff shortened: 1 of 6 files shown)


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -63,7 +62,7 @@ class PaparazziPlugin : Plugin<Project> {
   }
 
   private fun setupPaparazzi(project: Project) {
-    val unzipConfiguration = project.setupPlatformDataTransform()
+    val nativePlatformConfiguration = project.setupPlatformDataTransform()
```
liutikas, line 65: I'd rename setupPlatformDataTransform to getLayoutlibNative and split out the testImplementation dependency addition into a separate method since they serve very different purposes.

Return FileCollection instead of configuration and put the `.incoming.artifactView {}.files` inside of this method to hide away the implementation details.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -155,6 +164,9 @@ class PaparazziPlugin : Plugin<Project> {
 
         test.inputs.dir(mergeResourcesOutputDir)
         test.inputs.dir(mergeAssetsOutputDir)
+        test.inputs.files(nativePlatformFiles)
```
liutikas, line 167: I'd also add `.withPropertyName("paparazzi.layoutlibnative")` to name this input, so it shows up nicely in Gradle enterprise in case that is the cause for task being not up to date.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -218,18 +228,21 @@ class PaparazziPlugin : Plugin<Project> {
...
     )
     dependencies.registerTransform(UnzipTransform::class.java) { transform ->
-      transform.from.attribute(ArtifactAttributes.ARTIFACT_FORMAT, ArtifactTypeDefinition.JAR_TYPE)
+      transform.from.attribute(
+        ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE,
+        ArtifactTypeDefinition.JAR_TYPE
+      )
       transform.to.attribute(
-        ArtifactAttributes.ARTIFACT_FORMAT,
+        ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE,
```
liutikas, line 240: Note, this bumps your plugins minimum version of Gradle to 7.3, just so you are aware.

https://cs.android.com/android-studio/gradle/+/master:subprojects/core-api/src/main/java/org/gradle/api/artifacts/type/ArtifactTypeDefinition.java;l=39

In fact, I'd suggest setting up a test to use different versions of Gradle you wish to support.

paparazzi/paparazzi/build.gradle
```diff
@@ -107,8 +106,12 @@ def generateTestConfig = tasks.register("generateTestConfig") {
...
   dependsOn(generateTestConfig)
   systemProperty(
-      "paparazzi.test.resources",
-      generateTestConfig.map { it.outputs.files.singleFile }.get().path
+    "paparazzi.test.resources",
+    generateTestConfig.map { it.outputs.files.singleFile }.get().path
+  )
+  systemProperty(
+    "paparazzi.platform.data.root",
+    configurations.unzip.singleFile.path
```
liutikas, line 114: You probably want .absolutePath here

liutikas:
I was about to PR this (https://github.com/liutikas/paparazzi/tree/platform-input) :)

luis-cortes (approved):

liutikas:
Anything else missing for this to land?

jrodbx:
Was busy prepping then attending Droidcon NYC!

Starting to catch up on the repo.  Will aim to land this soon.

jrodbx:
Gonna try to land this today

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: great suggestion.  done (with slightly different naming).

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: done

paparazzi/paparazzi/build.gradle
jrodbx: good call, done

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: reverted back to original with comment to avoid burdening consumers who might be able to upgrade yet

Issue #482, opened 2022-07-02
Don’t resize snapshots to 1000px


sdoward:
**Description**
When recording content that is takes up a lot of height. The images that are produce are pixelated. 

**Steps to Reproduce**
- Create a scroll view or recycler view with content that is 3 or 4 times the length of a screen
- Configure paparazzi to use `renderingMode = SessionParams.RenderingMode.V_SCROLL`

**Expected behavior**
A clear images that are not pixelated

**Additional information:**
- Paparazzi Version: 1.0.0
- OS: mac OS Monterey 12.1
- Compile SDK: 21
- Gradle Version: 7.4.2
- Android Gradle Plugin Version: 7.1.3

**Screenshots**

![org sharethemeal android view dashboard_DashboardScreenShotTest_dashboardScreen SAMSUNG_GALAXY_S10,en_GB](https://user-images.githubusercontent.com/898078/177008959-c4550e36-353c-4847-a767-049d165b05ca.png)

fcduarte:
Interesting! This is reproducible by changing line 24 of `paparazzi-gradle-plugin/src/test/projects/verify-rendering-modes/src/test/java/app/cash/paparazzi/plugin/test/RenderingModesTest.kt` to `(0..25)` from `(0..2)`.

fcduarte:
Found it, this is due to every screenshot to be limited to a 1000px, either width or height: https://github.com/cashapp/paparazzi/blob/eb60debfd74975c8dd6b4a240f1c81c09dfd689e/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt#L93

Current implementation won't give you pixel perfect snapshots, I don't have context if this is intended. @jrodbx @luis-cortes any extra thoughts here? Does it make sense to implement a parameter like `createPixelPerfectSnapshot: Boolean` and then let library consumers pass it to their tests so they have the actual full not scaled image?

fcduarte:
Hi @sdoward, we discussed internally. Screenshots are limited to avoid the repo size to grow exponentially (if you use git LFS) and for making the HTML report easier to visualize. Could you share what's your use case for having an screenshot size closer to original? Thanks!

brudaswen:
Hacky workaround that might help you while the maxSie is not configurable:
```kotlin
    @Before
    fun initMaxImageSize() {
        maxImageSize?.let {
            Paparazzi::class.java.declaredFields.firstOrNull {
                it.name == "THUMBNAIL_SIZE"
[...]

PR #543, opened 2022-08-13
Add configuration for image size

sdoward:
As per conversation in https://github.com/cashapp/paparazzi/issues/482 this PR adds configuration so consumers can control the size of the images.

Api Considerations
-----

The changes introduces an addition to the api surface. In order to maintain backward compatibility I default to the previous behaviour (limit to 1000px). 

However the changes introduce 2 new options...

- Full bleed images (as per the original discussion)
- Image size limitation: it seems reasonable to allow consumers to set this limit if we also allow them to choose full bleed images

Testing
------

I wasn't sure how to write tests for this feature so I added to the sample instead. 

**Side note:** would it make sense to commit images produced in the sample project as some sort of end to end test?

Future Work
-------

The original issue mentioned that we might see issues with report rendering. I took the easier approach here of modifying css. A more solid solution could be to scale the image down when copying during html report generation. I was a little unsure how to do that.


fcduarte:
@sdoward thanks! For me looks good, will wait on @jrodbx's feedback ;)

micHar:
Hey, guys, any reason why this hasnt been merged? I'm pretty interested in this feature.

p-olszewski:
@fcduarte still waiting for the merge. Any updates?

npace:
Is there a specific person that can review these changes? Maybe @jrodbx or @TWiStErRob as they've reviewed recent PRs.
I mean it looks good to me, but what do I know as a mere user of the lib 😅

sdoward:
I updated the branch and resolved the conflicts

paparazzi/src/main/resources/index.html
```diff
@@ -16,6 +16,7 @@
 
     .screen img, .screen video {
       width: 300px;
+      max-height:1000px;
```
TWiStErRob, line 19:
```suggestion
      max-height: 1000px;
```

paparazzi/src/main/java/app/cash/paparazzi/ImageSize.kt
```diff
@@ -0,0 +1,8 @@
+package app.cash.paparazzi
+
+import app.cash.paparazzi.internal.ImageUtils
+
+sealed class ImageSize {
+  object FullBleed : ImageSize()
+  data class Limit(val height: Int = ImageUtils.THUMBNAIL_SIZE) : ImageSize()
+}
```
TWiStErRob, line 8: I think this might be unnecessary a simple `Int` with a `Int.MAX_VALUE` meaning full might be enough? (It's a nice abstraction, but doesn't fit into how the lib is abstracted right now. If it stays...
```suggestion
sealed class ImageSize {
  object Original : ImageSize()
  data class Limit(val height: Int = ImageUtils.THUMBNAIL_SIZE) : ImageSize()
}
```

paparazzi/src/main/java/app/cash/paparazzi/ImageSize.kt
```diff
@@ -0,0 +1,8 @@
+package app.cash.paparazzi
+
+import app.cash.paparazzi.internal.ImageUtils
+
+sealed class ImageSize {
+  object FullBleed : ImageSize()
+  data class Limit(val height: Int = ImageUtils.THUMBNAIL_SIZE) : ImageSize()
```
TWiStErRob, line 7: `ImageUtils` is internal, as a public user it's not possible to reproduce default behavior without using stable APIs. Maybe simply inline the default height into the constructor usage?

i.e. `private val imageSize: ImageSize = ImageSize.Limit(height = 1000)`

TWiStErRob (commented):
@npace I'm also just a user (contributor != collabolator), trying to be helpful with reviews.

<details><summary>I don't have rights to approve...</summary>

Review comments are my own opinions, while trying to take into account my knowledge of what Cash Inc. devs prefer based on my and other contributors' previous interactions with them. (I'm full on watching this repo, and can't get myself to unsubscribe, it's just too interesting.)

image

</details>

paparazzi/src/main/java/app/cash/paparazzi/ImageSize.kt
sdoward: I think using `Int.MAX_VALUE` in that way might be a little confusing. I think consumers would expect the image to scale to the given height. As it would scale down if you pass in a height that is smaller than the original height.

Perhaps a nullable `Int?` and we could name the parameter `maxImageHeight` passing null would mean there is no limit.

What do you think?

https://github.com/cashapp/paparazzi/pull/543/files#r1423206676

TWiStErRob: I don't like `Int?` in general, as no there's a practical limit to max bitmap size. So `maxHeight = 500000` (just a random large number) is the (practically) same as `maxHeight = Int.MAX`.

TWiStErRob: > I think consumers would expect the image to scale to the given height.

That's why I suggested to rename to `maxHeight` because "max" means if it's bigger, resize, otherwise, leave it alone.

Btw, don't let me sway you too much, wait for a Collaborator/Member to make a call here. They might actually prefer the sealed version.

sdoward: Thanks for that. I will hang on making changes for now

fcduarte:
For context, this is now fixed by https://github.com/cashapp/paparazzi/pull/1382

If you want to keep the original size of your snapshot image you can now do:

```kotlin
@get:Rule
val paparazzi = Paparazzi(
  deviceConfig = DeviceConfig.NEXUS_5,
  useDeviceResolution = true
)
```

Issue #544, opened 2022-08-15
ViewBinding Support


eschoenawa:
**Description**
Paparazzi uses `findViewById` to get references to the Views contained within the inflated layout. However, many modern apps rely on viewBinding to obtain these references to Views. So going back to `findViewById` for the screenshot tests not only feels like using out-of-date architecture, but also introduces inconsistency and is also more error-prone. I'd prefer to access the views in both tests and production code the same way.

**Expected behavior**
Paparazzi should support the use of ViewBinding to obtain references to the views.

**Additional information:**
- Paparazzi Version: 1.0.0

JakeWharton:
The library does not use or require the use of `findViewById`. You are free to obtains view instances in any way you see fit. 

Are you referring to the samples which use `findViewById`?

eschoenawa:
But don't I need to inflate the view using the library? How can I obtain a binding instance when I can't use `Binding.inflate`?

JakeWharton:
The `Paparazzi` instance exposes a `LayoutInflater` through the `layoutInflater` property that you can use to call `Binding.inflate`.

eschoenawa:
Ah, lovely! Thanks for the pointer, that solves it.

Maybe a view binding sample would give this more visibility. I somehow missed the `layoutInflater` property.

JakeWharton:
Yes, indeed. The samples using `findViewById` should be migrated.

PR #546, opened 2022-08-16
Update sample tests to use View Binding

eschoenawa:
Sample tests using `findViewById` are migrated to View Binding for easier access to the views and less boilerplate code.

I've done this migration because it wasn't obvious to me that Paparazzi worked with View Binding (see [my issue](https://github.com/cashapp/paparazzi/issues/544)). With this there's more visibility for this support.

I've decided not to migrate tests that don't access views besides the root one since those can be created quicker without View Binding. That way there are samples for inflating with the Paparazzi object _and_ inflating with Bindings.

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -9,6 +9,7 @@ android {
   }
   buildFeatures {
     compose true
+    viewBinding true
   }
   composeOptions {
     kotlinCompilerExtensionVersion '1.1.1'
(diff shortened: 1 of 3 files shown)


eschoenawa:
CI detected formatting issues, force-pushing another version with those addressed.

JakeWharton (approved):
Thanks!

Issue #424, opened 2022-05-05
Problem with compose and weight / verticalScroll modifiers


hardysim:
I'm using the current snapshot version via `classpath 'app.cash.paparazzi:paparazzi-gradle-plugin:1.0.0-SNAPSHOT'` to get compose support wich works already great but when I snapshot a composable which have special modifiers in it, I run into problems.

### `Modifier.weight()`
When `weight` is used on a composable, this single one is ignored and does not show up on the snapshot. Others around it are still fine.

I'm using this modifier for the texts on a dialog so that the message gets more space than the title. When taking a paparazzi snapshot, the contents whole text is gone because it's inside a Column using the weight-modifier:

![image](https://user-images.githubusercontent.com/3737512/166904370-f7f41709-aad0-4782-87f9-376f892ff6fd.png)

### `Modifier.verticalScroll()`
When using `verticalScroll` the whole snapshotis black / empty.

I'm using this modifier for the message on a dialog it can be scrolled when too long. When taking a paparazzi snapshot, the whole snapshot is broken:

![image](https://user-images.githubusercontent.com/3737512/166904908-9e161309-46a4-4f4c-b7b4-f74c7b68f77d.png)

jrodbx:
Can you provide a sample project/repro?

hardysim:
Sure. And as always, during creation of the example, I was able to narrow down the source of my problem 😊. Please have a look at the example at https://github.com/hardysim/paparazzi-weight-example:

I'm trying to use a `DeviceConfig` where `screenHeight` and `screenWidth` is set to `1`. The idea was that the snapshot-image is only as big as it needs to be (just like the preview in Android Studio). The `renderingMode` is set to `FULL_EXPAND`, so the image will expand as needed.

https://github.com/hardysim/paparazzi-weight-example/blob/master/mylibrary/src/test/java/com/example/mylibrary/ComposeTest.kt#L17

This works fine until I test a composable which uses a `weight()`- or `verticalScroll()`-modifier. When leaving the `DeviceConfig` untouched, the snapshot-images are rendered correctly (but bigger than they need to be).

So it seems like this is no bug but I want to ask: How can I get an "only as big as necessary" snapshot-image instead of one as big as the entire screen. Is this possible?

jrodbx:
> How can I get an "only as big as necessary" snapshot-image instead of one as big as the entire screen. Is this possible?

It is now as of #497 and #655.

Try:
```
    @get:Rule
[...]

Issue #367, opened 2022-02-01
ConstraintLayout in snapshot image using `match_parent` instead of `wrap_content`


mochadwi:
Actual:

![This is an image](https://gitlab.com/sampingan-mobile/mobile-android-native/uploads/3af40e43c30c80919045ee358b75d740/2022-02-01_11.48.35.gif)

Expected:

<img width="496" alt="Screen Shot 2022-02-01 at 12 03 16 PM" src="https://user-images.githubusercontent.com/7057959/151915280-df24313d-3594-47e7-95b4-fd6714713836.png">

<details><summary>XML (click to expand)</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:id="@+id/banner_root"
    android:layout_width="match_parent"
    android:visibility="gone"
    tools:visibility="visible"
    android:layout_height="wrap_content"
    android:paddingLeft="18dp"
    android:paddingStart="18dp"
    android:paddingTop="16dp"
    android:paddingRight="16dp"
    android:paddingEnd="16dp"
    android:paddingBottom="20dp">

    <ImageView
        android:id="@+id/imageview_empty_logo"
        android:layout_width="24dp"
        android:layout_height="24dp"
        android:adjustViewBounds="true"
        android:scaleType="fitCenter"
        android:layout_marginEnd="16dp"
        android:layout_marginRight="16dp"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:layout_constraintEnd_toStartOf="@id/textview_empty_description"
        tools:srcCompat="@drawable/ic_indicator_warning_yellow"/>

    <TextView
        android:id="@+id/textview_empty_description"
        style="@style/TextNormalBlack"
        android:layout_width="0dp"
        android:layout_height="wrap_content"
        app:layout_constraintBottom_toTopOf="@id/textview_banner_action"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toEndOf="@id/imageview_empty_logo"
        app:layout_constraintTop_toTopOf="parent"
        app:layout_constraintVertical_chainStyle="spread"
        tools:text="@tools:sample/lorem"
        tools:textColor="@color/text_normal"/>

[...]

Issue #37, opened 2019-04-05
Snapshot without a device frame


swankjesse:
Perhaps something like this:

```
    paparazzi.snapshot(avatar, "avatar", deviceFrame = DeviceFrame.NONE)
```

nikwotton:
+1

TWiStErRob:
Is this a duplicate of https://github.com/cashapp/paparazzi/issues/313?

TWiStErRob:
Isn't testing a widget (`View` with specific w/h) the same as testing a composable `Button`?

PR #497, opened 2022-07-13
Removes requirement to render entire device Screen 

rharter:
This adds support for `RenderMode.SHRINK`, allowing you to make view-only screenshots without saving an entire device frame.

PR #550, opened 2022-08-18
Add shrink mode to shrink views (XML and composables) to the smallest size required

BenedictP:
Added a shrink mode to shrink views to the direction the user wants. This should normally be part of the `RenderingMode` but we don't control this class so I needed to add an extra class to control the shrink mode.

When the `RenderingMode` is set to `SHRINK` and the `shrinkDirection` is `null` then XML layouts and composables are rendered with `wrap_content` and without any device frame (also respecting margins and padding):
|XML|Composable|
|----------|-------|
|image | image|

When `shrinkDirection` is set to `VERTICAL` the width is `match_parent` and the height is `wrap_content`:
|XML|Composable|
|----------|-------|
|image | image |

When `shrinkDirection` is set to `HORIZONTAL` the width is `wrap_content` but the height is `match_parent` (check the screenshot for this PR because the screenshots are very high).

Will fix:
- https://github.com/cashapp/paparazzi/issues/367
- https://github.com/cashapp/paparazzi/issues/37
- https://github.com/cashapp/paparazzi/issues/424


rharter:
This appears to duplicate the proposed change in #497, which I think is able to accomplish the same thing using `RenderingMode`'s `horizAction` and `vertAction` properties to avoid expanding the API surface.

BenedictP:
I think https://github.com/cashapp/paparazzi/pull/497 is only fixing composables and not XML layouts and when we are using `android:layout_width="match_parent"` for the root node inside a layout we mean the device width. Just using `SHRINK` would even shrink these views in their width and the device width will be ignored. That's the reason I needed to expand the API surface.

rharter:
I don't see anything in #497 that limits it to compose, and the tests even test legacy views. 

RenderMode has horizontal and vertical options, just like ShrinkDirection, so if you wanted a device width you'd choose a RenderMode that didn't shrink along the horizontal axis. 

It seems like ShrinkDirection duplicates the RenderMode settings, defining whether the view should shrink or not along either axis.

BenedictP:
I checked out https://github.com/cashapp/paparazzi/pull/497/files and created this XML layout (display.xml):
```
<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
  android:layout_height="wrap_content"
  android:layout_margin="16dp"
  android:background="@color/keypadGreen">

  <TextView
    android:id="@+id/number"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:gravity="center"
    android:text="5"
    android:textColor="@android:color/white"
    android:textFontWeight="100"
    android:textSize="70dp"
    app:layout_constraintBottom_toBottomOf="parent"
    app:layout_constraintStart_toStartOf="parent"
    app:layout_constraintTop_toTopOf="parent" />

  <ImageView
    android:layout_width="40dp"
    android:layout_height="40dp"
    android:layout_marginStart="16dp"
    android:src="@drawable/camera"
    app:layout_constraintBottom_toBottomOf="@id/number"
    app:layout_constraintStart_toEndOf="@id/number"
    app:layout_constraintTop_toTopOf="@id/number" />

</androidx.constraintlayout.widget.ConstraintLayout>
```

Then I created this test class:
```
class ShrinkDirectionTest {

  @get:Rule
  val paparazzi = Paparazzi(
    deviceConfig = DeviceConfig.NEXUS_5.copy(softButtons = false)
  )

  @Test
  fun `rendering mode shrink`() {
    paparazzi.unsafeUpdateConfig(renderingMode = SessionParams.RenderingMode.SHRINK)
    val launch = paparazzi.inflate<ConstraintLayout>(R.layout.display)
    paparazzi.snapshot(launch)
  }

  @Test
  fun `rendering mode normal`() {
    paparazzi.unsafeUpdateConfig(renderingMode = SessionParams.RenderingMode.NORMAL)
    val launch = paparazzi.inflate<ConstraintLayout>(R.layout.display)
    paparazzi.snapshot(launch)
  }

  @Test
  fun `rendering mode v scroll`() {
    paparazzi.unsafeUpdateConfig(renderingMode = SessionParams.RenderingMode.V_SCROLL)
    val launch = paparazzi.inflate<ConstraintLayout>(R.layout.display)
    paparazzi.snapshot(launch)
  }

  @Test
  fun `rendering mode h scroll`() {
    paparazzi.unsafeUpdateConfig(renderingMode = SessionParams.RenderingMode.H_SCROLL)
    val launch = paparazzi.inflate<ConstraintLayout>(R.layout.display)
    paparazzi.snapshot(launch)
  }

  @Test
  fun `rendering mode full expand`() {
    paparazzi.unsafeUpdateConfig(renderingMode = SessionParams.RenderingMode.FULL_EXPAND)
    val launch = paparazzi.inflate<ConstraintLayout>(R.layout.display)
    paparazzi.snapshot(launch)
  }
}
```

The result is always the same:
image

Would love to know what I'm doing wrong or which parameter I need to change/set because that was the reason to create this PR. Nothing I tried was working with the above example 😞

jrodbx:
Closed by #497

PR #555, opened 2022-08-27
Add compose accessibility support

yschimke:
Modelled off the AccessibilityRenderExtension. But instead using the `ViewRootForTest.onViewCreatedCallback` hook using in composeTestRule. And a separate SnapshotHandler to render instead of using views.

Derived from https://github.com/google/horologist/pull/524

image


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/A11ySnapshotHandler.kt
```diff
@@ -0,0 +1,217 @@
...
+          if (it.role != null || it.disabled) {
+            val role = if (it.role != null) "Role " + it.role + " " else ""
+            val heading = if (it.heading) "Heading " else ""
+            val disabled = if (it.disabled) "Disabled" else ""
+            drawString(role + heading + disabled, 50f, 28f * index++)
+          }
+          if (it.contentDescription != null) {
+            drawString(
+              "Content Description " + it.contentDescription.joinToString(", "),
+              1050f,
```
gmarques33, line 188: This should be 50f.

yschimke: Why? It will draw over the image in that case.

gmarques33: The image is created with width = 600f: https://github.com/cashapp/paparazzi/pull/555/files/85eaa50bce25715fa7d4c41eac6c870ba74fc067#diff-d564e3b448efd8866f9bbbf1ea2d0a900d0892e6a38c5d73785218a3906ddd2eR165 and all other calls use 50f for the x axis, for example: https://github.com/cashapp/paparazzi/pull/555/files/85eaa50bce25715fa7d4c41eac6c870ba74fc067#diff-d564e3b448efd8866f9bbbf1ea2d0a900d0892e6a38c5d73785218a3906ddd2eR196

With 1050f we'll be drawing outside the image.

seanfreiburg:
@yschimke Any chance you could get this merged? We are interested in using this at Netflix.

yschimke:
@jrodbx any appetite for this? We like it and use it for all screen.

But I guess it depends on the bigger plugin/interceptor strategy?

yschimke:
Added a test.  Although it fails if Text is used.

yschimke:
The Graphics2D code in A11ySnapshotHandler seems platform specific.

image

```
app.cash.paparazzi.gradle.PaparazziPluginTest > composeA11yWear FAILED
    java.lang.AssertionError: Expected % diff less than 0.01, but was: 0.1777610294117647
        at org.junit.Assert.fail(Assert.java:89)
        at app.cash.paparazzi.gradle.ImageSubject$ImageAssert.withThreshold(ImageSubject.kt:54)
        at app.cash.paparazzi.gradle.ImageSubject$ImageAssert.withDefaultThreshold(ImageSubject.kt:40)
        at app.cash.paparazzi.gradle.PaparazziPluginTest.composeA11yWear(PaparazziPluginTest.kt:984)
```

yschimke:
I'm very happy to follow up with this on looking at how to avoid platform differences. from fonts + Graphics2D.  In practice this change as is, works for us since we have a consistent CI environment.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -32,7 +32,7 @@ import app.cash.paparazzi.accessibility.RenderSettings.getColor
 import app.cash.paparazzi.accessibility.RenderSettings.toColorInt
 import app.cash.paparazzi.accessibility.RenderSettings.withAlpha
 
-class AccessibilityRenderExtension : RenderExtension {
+public class AccessibilityRenderExtension : RenderExtension {
```
geoff-powell, line 35: nit: not sure why `public` is needed in context of this pr.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/A11ySnapshotHandler.kt
```diff
@@ -0,0 +1,231 @@
...
+import java.awt.Composite
+import java.awt.Graphics2D
+import java.awt.image.BufferedImage
+import kotlin.math.max
+
+/**
+ * A Paparazzi SnapshotHandler that renders the snapshot, with a light coloured overlay,
+ * and adjacent to a legend with matching colours.
+ */
+class A11ySnapshotHandler(
```
geoff-powell, line 33: we should probably avoid modifying the snapshot handler. Ideally, a11y render extension should handle all the logic of fetching metadata and displaying it. Also when building the initial render extension, we moved away from using Graphics2D apis for drawing for 2 reasons.

1. Platform consistency across Mac, Linux, and Windows rendering differently.
2. It caused build issues when depending on Paparazzi plugin where java.awt.* classes couldn't be found.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/A11ySnapshotHandler.kt
```diff
@@ -0,0 +1,231 @@
...
+          drawRect(10, start * 28 - 21, modifiedImage.width - 20, (end - start) * 28)
+
+          index++
+        }
+      }
+    }
+
+    private fun colorForIndex(i: Int): Color {
+      return colors[i % colors.size]
+    }
```
geoff-powell, line 229: nit: We had an issue deciding colors this way as it's not consistent across updating UI thereby bumping the index for everything. Instead, we opted for using the view id to ensure the most consistent color for views.

paparazzi/paparazzi-gradle-plugin/src/test/projects/compose-a11y-wear/src/test/java/app/cash/paparazzi/plugin/test/ComposeA11yWearTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import app.cash.paparazzi.Paparazzi
+import app.cash.paparazzi.Paparazzi.Companion.determineHandler
+import app.cash.paparazzi.accessibility.A11ySnapshotHandler
+import app.cash.paparazzi.accessibility.ComposeA11yExtension
+import org.junit.Rule
+import org.junit.Test
+
+class ComposeA11yWearTest {
+  private val composeA11yExtension = ComposeA11yExtension()
+  private val maxPercentDifference = 1.0
```
geoff-powell, line 13: nit: just upping the max difference here isn't a good solution to the images not being exact on different arch/platforms.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/ComposeA11yExtension.kt
```diff
@@ -0,0 +1,127 @@
...
+      )
+    }
+
+    p0.children.forEach {
+      processAccessibleChildren(it, fn)
+    }
+  }
+
+  override fun renderView(contentView: View): View {
+    val composeView = (contentView as ViewGroup).getChildAt(0) as ComposeView
```
geoff-powell, line 102: this implementation expects the root view to be a `ComposeView`. It doesn't fix issues where you have a view hierarchy with mixed legacy views and compose views.

geoff-powell (changes requested):
Thanks for taking a look into this. There are some concerns/questions I added but I think that this is the right direction for supporting compose for a11y

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/ComposeA11yExtension.kt
yschimke: I'll test with this scenario.  I'm not sure we can easily support this, I might end up making it clear what is supported.

The A11y model for Compose seems different from Views, so unless I changed this to work as an accessibility service, it might not be possible to mix (I'll confirm in a test). I think that still captures the final state of most apps.  Either completely views, or completely compose.

paparazzi/paparazzi-gradle-plugin/src/test/projects/compose-a11y-wear/src/test/java/app/cash/paparazzi/plugin/test/ComposeA11yWearTest.kt
yschimke: I'll remove those things that cause differences from the default implementation, but allow enough extension to allow me as a user of this library, to put them back in my app code.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/A11ySnapshotHandler.kt
yschimke: I will take a look at what to do here. We don't have ids, but I'm sure I can find something similar.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/A11ySnapshotHandler.kt
yschimke: OK, I'll take a look at this.  I looked at https://github.com/cashapp/paparazzi/blob/master/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt, which works in the world of Views. 

It seemed to be a nice thing, that instead of mutating the View hierarchy, this was able to work in two passes, 

1) Collecting the a11y nodes
2) decorating the untouched Compose snapshot, with accessibility overlays and text. 

But it would potentially solve/simplify the task of matching, as we would avoid Graphics2D.

My one nagging doubt is related to the lifecycle of `ViewRootForTest.onViewCreatedCallback` and `composeView.viewTreeObserver.addOnPreDrawListener`, I'll see what I can do here once I addressed other feedback.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/ComposeA11yExtension.kt
saket: > I think that still captures the final state of most apps. Either completely views, or completely compose.

Doesn't this exclude apps that are incrementally migrating to compose ui?

yschimke:
Changes required for 1.2 release

https://github.com/google/horologist/pull/938/files#diff-97977a5620e4bc79f81f77977b54e2c278325e531919ce7ea5b8668791914006

https://github.com/google/horologist/pull/938/files#diff-0b30e0de6183d1fe1fbcff8e9ffa085866c455593ac080cbffeba349e1b32e52

yschimke:
@geoff-powell @saket I'm stuck.  The A11y approach we use for Views and Compose seem incompatible.  I don't know how to bridge them.

**Views**

Works on enumerating important Views.

```
fun View.findAccessibilityViews(): List<View> {
  val accessibilityViews = mutableListOf<View>()
  if (isImportantForAccessibility && !iterableTextForAccessibility.isNullOrBlank()) {
    accessibilityViews.add(this)
  }
```

This happens synchronously in AccessibilityRenderExtension.renderView

**Compose (this)**

Registers the ViewRootForTest.onViewCreatedCallback (used for createComposeRule testing) during test setup.
This let's us grab androidx.compose.ui.node.RootForTest during the test run.

ComposeA11yExtension.renderView can't access state synchronously, instead registers `composeView.viewTreeObserver.addOnPreDrawListener` to grab rootForTest.semanticsOwner.rootSemanticsNode later.

**Common Approaches**

I don't have any representation of Semantic Nodes that is common for Views and Compose.

Maybe writing an a11y service?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/ComposeA11yExtension.kt
saket: I unfortunately don't have a good answer

saket: I unfortunately don't have a good answer, sorry!

geoff-powell: Ya, I was able to handle this case in this pr. https://github.com/cashapp/paparazzi/pull/702 
It takes a lot from the work in this PR.

yschimke: A big thank you for taking this over.

Issue #553, opened 2022-08-26
Coil images not rendering correctly


JulianEggers:
**Description**
Paparazzi does not render Coil-Images under special circumstances:
- Using coils `AsyncImage` within a `LazyColumn` works with paparazzi.
- However, it stops working whenever prior to that another paparazzi test is launched.

**Steps to Reproduce**

Some notes regarding the code sample: 
- I  assume this behavior is influenced by the order in which tests are executed and it seems like the order in my case is: `a()` then `aa()` then `aaa()`. (?)
- Removing the a-function results in 'aa' being the first snapshot that is being created. -> Rendering the image works.
- Re-adding the a-function results in 'a' being the first snapshot and 'aa' the second. -> Rendering the 'aa'-snapshot stops working (image missing). 
- aaa-function always works
- aaa-function also works when wrapping the image into a Column.

In short:
- This bug only occurs when working with coil images (not normal compose images) wrapped into a LazyLayout (like LazyColumn) and only when prior to that another paparazzi test is launched.

```kotlin
class ExampleUnitTest {

    @get:Rule
    val paparazzi = Paparazzi( )

    /**
    * Running a() first makes aa() not render the image.
    */
    @Test
    fun a() {
        paparazzi.snapshot {
        }
    }

    /**
    * aa() does not work when run second.
    * aa() does work when run as first test.
    */
    @Test
    fun aa() {
        paparazzi.snapshot {
            LazyColumn {
                item {
                    val imageRequest = ImageRequest.Builder(LocalContext.current)
                        .fallback(ColorDrawable(Color.CYAN))
                        .build()

                    AsyncImage(
                        model = imageRequest,
                        contentDescription = null,
                        contentScale = ContentScale.Crop,
                        modifier = Modifier.height(180.dp)
                    )
                }
            }
        }
    }

    /**
    * aaa() always works.
    */
    @Test
    fun aaa() {
        paparazzi.snapshot {
            val imageRequest = ImageRequest.Builder(LocalContext.current)
                .fallback(ColorDrawable(Color.CYAN))
                .build()

            AsyncImage(
                model = imageRequest,
                contentDescription = null,
                contentScale = ContentScale.Crop,
                modifier = Modifier.height(180.dp)
            )
        }
    }
}
```
(Using `implementation "com.google.accompanist:accompanist-drawablepainter:0.24.13-rc"` to create drawables)

**Expected behavior**
Paparazzi should always render coil images within a LazyColumn.

**Additional information:**
[...]

PR #557, opened 2022-08-29
Reset AndroidUiDispatcher between compose snapshots

gamepro65:
AndroidUiDispatcher schedules a callback for both a Handler and Choreographer using [scheduledTrampolineDispatch and scheduledFrameDispatch](https://android.googlesource.com/platform//frameworks/support/+/c354fa29007af485641665604612df30199a644a/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidUiDispatcher.android.kt#52) to prevent queuing more than once. forceReleaseComposeReferenceLeaks() clears out any pending operations but we need to also reset these two fields else AndroidUiDispatcher won't dispatch any jobs after the first test.

I believe this will resolve https://github.com/cashapp/paparazzi/issues/553 and fixes an issue with infiniteAnimation not running after the first test.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -564,6 +564,16 @@ class Paparazzi @JvmOverloads constructor(
       .apply { isAccessible = true }
       .get(dispatcher) as ArrayDeque<*>
     toRunTrampolined.clear()
+    // Upon reference leaks being fixed, verify we don't need to reset these values for
+    // AndroidUiDispatcher to continue dispatching between tests.
+    dispatcher.javaClass
+      .getDeclaredField("scheduledTrampolineDispatch")
+      .apply { isAccessible = true }
+      .set(dispatcher, false)
+    dispatcher.javaClass
+      .getDeclaredField("scheduledFrameDispatch")
+      .apply { isAccessible = true }
+      .set(dispatcher, false)
   }
 
   private class PaparazziComposeOwner private constructor() : LifecycleOwner, SavedStateRegistryOwner {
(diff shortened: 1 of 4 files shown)


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -564,6 +564,14 @@ class Paparazzi @JvmOverloads constructor(
...
       .get(dispatcher) as ArrayDeque<*>
     toRunTrampolined.clear()
+    dispatcher.javaClass
+      .getDeclaredField("scheduledTrampolineDispatch")
+      .apply { isAccessible = true }
+      .set(dispatcher, false)
+    dispatcher.javaClass
+      .getDeclaredField("scheduledFrameDispatch")
+      .apply { isAccessible = true }
+      .set(dispatcher, false)
```
jrodbx, line 574: One day the Compose Runtime will probably fix these memory leaks.  At the point, will we still need this?  

If so, maybe add a comment for future contributors to remember not to remove this?

jrodbx (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
gamepro65: I'd imagine if those leaks were fixed that this would go as well since us clearing the callback lists without reseting these values is whats getting us into this situation in the first place. They would have to handle resetting them as part of that larger fix else they would run into the situation we are today.

The included plugin test should catch any regression's if/when we remove this function, but can leave a comment to double check.

adonislapin:
Hello guys, Im seeing this error today, im Using in Android (id 'app.cash.paparazzi' version '1.1.0-SNAPSHOT'), last friday it worked ok

java.lang.NoSuchFieldError: Companion
	at app.cash.paparazzi.Paparazzi$PaparazziComposeOwner.<init>(Paparazzi.kt:581)
	at app.cash.paparazzi.Paparazzi$PaparazziComposeOwner.<init>(Paparazzi.kt)
	at app.cash.paparazzi.Paparazzi$PaparazziComposeOwner$Companion.register(Paparazzi.kt:588)
	at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:198)
	at app.cash.paparazzi.Paparazzi.snapshot$default(Paparazzi.kt:191)

Issue #558, opened 2022-08-29
Layoutlib crash on API 33


ember-rose:
**Description**
As noted in https://github.com/cashapp/paparazzi/issues/489#issuecomment-1195673126, the version of layoutlib packaged with Paparazzi does not support API 33. I'm opening this issue to track API 33 support and document the failure case I encountered.

**Steps to Reproduce**
Execute a Paparazzi test with target and compile SDK 33. 

**Expected behavior**
Paparazzi would initialize layoutlib and take snaphots.

**Additional information:**
- Paparazzi Version: 1.0.0 (layoutlib: 2021.1.1-573f070)
- OS: Linux, macOS
- Compile SDK: 33
- Gradle Version: 7.5
- Android Gradle Plugin Version: 7.4.0-alpha08

**Stacktraces**
<details>
<summary>
<code>java.lang.IllegalArgumentException: value of system property 'ro.build.version.known_codenames' is longer than 91 characters</code>
</summary>

```
java.lang.IllegalArgumentException: value of system property 'ro.build.version.known_codenames' is longer than 91 characters: Base,Base11,Cupcake,Donut,Eclair,Eclair01,EclairMr1,Froyo,Gingerbread,GingerbreadMr1,Honeycomb,HoneycombMr1,HoneycombMr2,IceCreamSandwich,IceCreamSandwichMr1,JellyBean,JellyBeanMr1,JellyBeanMr2,Kitkat,KitkatWatch,Lollipop,LollipopMr1,M,N,NMr1,O,OMr1,P,Q,R,S,Sv2,Tiramisu
	at android.os.SystemProperties.set(SystemProperties.java:229)
	at com.android.layoutlib.bridge.Bridge.setSystemProperties(Bridge.java:319)
	at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
	at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2445)
	at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2501)
	at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2700)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2630)
	at java.base/java.lang.Runtime.load0(Runtime.java:768)
	at java.base/java.lang.System.load(System.java:1837)
	at com.android.layoutlib.bridge.Bridge.loadNativeLibraries(Bridge.java:720)
[...]

PR #559, opened 2022-09-01
Upgrade to a newer version of layoutlib

liutikas:
- Removes use of FLAG_RENDER_HIGH_QUALITY_SHADOW and FLAG_ENABLE_SHADOW
  as these flags are no-op. See https://cs.android.com/android-studio/platform/tools/adt/idea/+/b9847d5dced4bf81f39976c1d3aced85beb6c13c
- Removes use of FLAG_KEY_APPLICATION_PACKAGE https://cs.android.com/android-studio/platform/tools/adt/idea/+/ce1c92f23003bdc48df1621a7e24cbdc0702fc8e
- This makes it compatible with Android API 33
- Updates 4 golden images as rendering changed a bit with a newer layoutlib.

Test: ran existing tests, still passing

Fixes: #558


gradle/dependencies.gradle
```diff
@@ -1,14 +1,14 @@
...
+  // maps to this commit: https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/f5f9f71
   // which runs on JDK 11
-  ext.layoutLibPrebuiltSha = 'fa3aa65'
+  ext.layoutLibPrebuiltSha = 'f5f9f71'
 
   ext.versions = [
     kotlin: '1.6.10',
     agp: '7.1.2',
-    layoutlib: "2021.2.1-patch1-$layoutLibPrebuiltSha",
+    layoutlib: "2022.1.1-canary-$layoutLibPrebuiltSha",
```
jrodbx, line 11: As discussed in our call a few weeks back, pointing to Electric Eel canary builds would be undesirable since these aren't considered stable yet (Is that a fair statement?).  Dolphin betas might be fine though.

I understand the desire for API 33 support, but any chance the Google Tools team can consider moving that support up to Dolphin release instead?

Or if the desire here is to simply to push a canary repackaged layoutlib to Maven Central, we can certainly do that!

ZacSweers: Not privy to the mentioned call, but since layoutlib isn't a public API in general wouldn't it be better to just generally target latest? That's sorta what we do with other tools built on unstable APIs (compiler plugins, AGP-oriented gradle plugins, etc).

jrodbx: Not everyone feels comfortable with using beta or canary versions (I *do*, but trying to keep the wider community in mind).

liutikas: I think we should land it as is. I believe the risk is very low in doing so. Looking at the [prebuilts repo](https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+log/refs/heads/mirror-goog-studio-main) there has only been [one more drop of prebuilts](https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/2aee3a801ce0b7597a0b72753059d4a8e122019f) with minor fixes. It is using Android T as a base that has shipped as stable to devices and AOSP.

The ship has sailed for AS D, AS EE is close to beta now.  I think we should pick this up.

jrodbx: Maybe this canary was relative low change, but as a library maintainer, I wouldn't want to depend on future Google layoutlib canaries being as low change.

Secondly, I don't want to manage several Paparazzi releases in parallel, for each of the AS release channels.

Thirdly, the application package flag change is replaced by a new method -- i believe, in layoutlib-api? -- which is blocked by an sdk-common bump, which in turn is blocked by the resource repository work.

I'll push a new layoutlib if you would like to maintain a fork for a little longer.

hrach:
Why closed?

chomamatus:
> Why closed?

thats a very good question...🤔

PR #562, opened 2022-09-01
Start using gradle-build-action

liutikas:
Enables the use of Github Action cache to restore relevant Gradle caches to speed up builds.

Details:
https://github.com/marketplace/actions/gradle-build-action


jrodbx (approved):
Awesome!

PR #563, opened 2022-09-01
Enable parallel execution in paparazzi build

liutikas:


rharter (commented):
Lgtm!

paparazzi/gradle.properties
```diff
@@ -18,3 +18,4 @@ SONATYPE_HOST=DEFAULT
 RELEASE_SIGNING_ENABLED=true
 
 org.gradle.caching=true
+org.gradle.parallel=true
```
TWiStErRob, line 21: @JakeWharton had an interesting point about artifact uploads here:
https://github.com/JakeWharton/dagger-reflect/pull/38
(not sure if it applies here, or still a problem)

TWiStErRob: Note: if it's a problem, it can be mitigated by adding `--no-parallel` on CI, maybe.

liutikas: So this is a bug in `com.vanniktech.maven.publish` plugin? Ideally uploadArchives would just use a BuildService where only 1 can be acquired at once.

jrodbx: I'd be ok merging this if the parallel artifact publishing issue has been fixed.

TWiStErRob: I'm not using vanniktech plugin and haven't seen failures related to this on my CI (it doesn't mean it's not a problem though).

liutikas: So do we have any evidence to suggest that it is broken for this project other than a comment on an unrelated project by Jake?

JakeWharton: Which project is irrelevant here. The behavior is on Sonatype's side as concurrent uploads get staged into separate repositories. Build services didn't exist at the time of my comment, but you'll see that i linked to the issue which resulted in their creation. It would be nice to see the release plugin set it up automatically.

If you want to test that it's still a problem do a release publish of any project locally with parallel and check Sonatype. You should see 2 or more repositories created on their side with artifacts split between them. (You can drop the repos after the test)

JakeWharton: I can probably run this test later tonight. Jrod is out and about.

gabrielittner: The problem still exists.

> It would be nice to see the release plugin set it up automatically.

I've started working on adding this to the plugin so hopefully this will be a problem of the past soon

TWiStErRob: In the meantime, I think this can be merged with 2 more changed lines:
 * One for the release process to change the command with --no-parallel
 * One for the CI upload with --no-parallel

This will enable the change in most places (e.g. dev builds), then if you're done with build services impl, the PR that updates the plugin can remove the flags.

jrodbx: Adding the CI flag is a good workaround until fixed in the release plugin 👍

jrodbx: @liutikas wanna add the `--no-parallel` flag in the CI upload as a temp workaround?

jrodbx: nvm, i'll merge and add it in a separate PR.

jrodbx (approved):

PR #575, opened 2022-09-14
Explicit opt in to RTL support

jrodbx:
In production apps, RTL support is enabled from the [application manifest](https://developer.android.com/guide/topics/manifest/application-element#supportsrtl).  
In LayoutLib, RTL support is enabled in [RenderParams.setRtlSupport](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:layoutlib-api/src/main/java/com/android/ide/common/rendering/api/RenderParams.java;l=157), which was set to `true` as part of [this PR](https://github.com/dniHze/paparazzi/blob/e88a9a99a60189c32fe2684aaeb87f58053a42e1/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/SessionParamsBuilder.kt#L108).

However, while testing internally on CashApp, it looks like a few screens (primarily due to [this Contour issue](https://github.com/cashapp/contour/issues/92)) break with this change, so let's enable explicit opt-in for those using Contour, until the upstream issue is fixed.

Note the X and Search icons in the toolbar in the following screenshots from a production app, when the aforementioned manifest attribute is enabled:

<kbd>Screenshot_20220913-233354</kbd><kbd>Screenshot_20220913-233547</kbd>


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -86,7 +86,8 @@ class Paparazzi @JvmOverloads constructor(
   private val appCompatEnabled: Boolean = true,
   private val maxPercentDifference: Double = 0.1,
   private val snapshotHandler: SnapshotHandler = determineHandler(maxPercentDifference),
-  private val renderExtensions: Set<RenderExtension> = setOf()
+  private val renderExtensions: Set<RenderExtension> = setOf(),
+  private val supportsRtl: Boolean = false
```
JakeWharton, line 90:
```suggestion
  private val supportsRtl: Boolean = false,
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -158,7 +159,8 @@ class Paparazzi @JvmOverloads constructor(
       .copy(
         layoutPullParser = LayoutPullParser.createFromString(contentRoot),
         deviceConfig = deviceConfig,
-        renderingMode = renderingMode
+        renderingMode = renderingMode,
+        supportsRtl = supportsRtl
```
JakeWharton, line 163:
```suggestion
        supportsRtl = supportsRtl,
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/SessionParamsBuilder.kt
```diff
@@ -48,7 +48,8 @@ internal data class SessionParamsBuilder(
   private val layoutPullParser: LayoutPullParser? = null,
   private val projectKey: Any? = null,
   private val minSdk: Int = 0,
-  private val decor: Boolean = true
+  private val decor: Boolean = true,
+  private val supportsRtl: Boolean = false
```
JakeWharton, line 52:
```suggestion
  private val supportsRtl: Boolean = false,
```

JakeWharton (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: i tried but spotless complained :(

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/SessionParamsBuilder.kt
jrodbx: i tried but spotless complained :(

jrodbx:
> If you want you can force trailing commas with ktlint to keep diffs like this clean with only added lines instead of causing each new parameter to also change a line.

i tried but spotless complained :( cc: @fcduarte who had looked into this

JakeWharton:
It's not the default. You have to opt in so ktlint knows you want it.

JakeWharton:
If you switch to Kotlinter than you can put those in your editorconfig and both the IDE and ktlint will honor them.

PR #578, opened 2022-09-17
Fix master; upgrade test project to Compose latest

jrodbx:


paparazzi/paparazzi-gradle-plugin/src/test/projects/layout-direction/src/test/java/app/cash/paparazzi/plugin/test/LayoutDirectionTest.kt
```diff
@@ -32,7 +32,8 @@ class LayoutDirectionTest(
     deviceConfig = DeviceConfig.NEXUS_5.copy(
       layoutDirection = localeAndDirection.direction,
       locale = localeAndDirection.tag
-    )
+    ),
+    supportsRtl = true
```
jrodbx, line 36: I'm a bit perplexed why this test didn't fail per https://github.com/cashapp/paparazzi/pull/575?

luis-cortes (approved):

PR #580, opened 2022-09-19
Remove deprecated `forUseAtConfigurationTime()`

veyndan:
https://docs.gradle.org/current/userguide/upgrading_version_7.html#for_use_at_configuration_time_deprecation

> Provider#forUseAtConfigurationTime is now deprecated and scheduled for removal in Gradle 9.0. Clients should simply remove the call.


jrodbx (approved):
mind rebasing now that the build is fixed?

PR #581, opened 2022-09-20
Expose an API for frame time

swankjesse:
Allow capturing snapshots at arbitrary points in an animation. This is building on the same mechanism as our gif support, but without the complexity of choosing a video encoding format.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -208,8 +208,8 @@ class Paparazzi @JvmOverloads constructor(
   }
 
   @JvmOverloads
-  fun snapshot(view: View, name: String? = null, timeMillis: Long = 0L) {
-    takeSnapshots(view, name, TimeUnit.MILLISECONDS.toNanos(timeMillis), -1, 1)
+  fun snapshot(view: View, name: String? = null, offsetMillis: Long = 0L) {
+    takeSnapshots(view, name, TimeUnit.MILLISECONDS.toNanos(offsetMillis), -1, 1)
   }
 
   @JvmOverloads
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -208,8 +208,8 @@ class Paparazzi @JvmOverloads constructor(
   }
 
   @JvmOverloads
-  fun snapshot(view: View, name: String? = null) {
-    takeSnapshots(view, name, 0, -1, 1)
+  fun snapshot(view: View, name: String? = null, timeNanos: Long = 0L) {
+    takeSnapshots(view, name, timeNanos, -1, 1)
   }
 
   @JvmOverloads
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -208,8 +208,8 @@ class Paparazzi @JvmOverloads constructor(
   }
 
   @JvmOverloads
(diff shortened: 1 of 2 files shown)


paparazzi/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -115,6 +116,65 @@ class PaparazziTest {
...
+    animator.start()
+    assertThat(AnimationHandler.getAnimationCount()).isEqualTo(1)
+
+    paparazzi.snapshot(view, timeNanos = 0L)
+    assertThat(log).containsExactly(
+      "onDraw text=",
+    )
+    log.clear()
+
+    paparazzi.snapshot(view, timeNanos = 2_000_000_000L)
```
swankjesse, line 156: I could go either way on millis vs. nanos here. In general I prefer nanos 'cause the precision is potentially valuable (?) but I also don’t like how many annoying `0` characters I need to write here when specifying a time literally.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -208,8 +208,8 @@ class Paparazzi @JvmOverloads constructor(
   }
 
   @JvmOverloads
-  fun snapshot(view: View, name: String? = null) {
-    takeSnapshots(view, name, 0, -1, 1)
+  fun snapshot(view: View, name: String? = null, timeNanos: Long = 0L) {
```
swankjesse, line 211: I didn’t add an overload for compose, and I probably should. Does the timeNanos argument slot in before the Composable? And if so, do I need a manual overload?

swankjesse:
After a little study there’s one big hangup with this approach - it only works for exactly one snapshot after each animation is triggered. This is because Paparazzi clears out the animation handler completely after taking a snapshot, which fights our ability to snapshot specific frames.

I’d like to find an alternative solution to whatever problem this call is attempting to fix.

```
        AnimationHandler.sAnimatorHandler.set(null)
```

paparazzi/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
jrodbx: Agreed, millis would be nicer here

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: My instinct says yes, should support Compose but maybe in a separate PR because I don't think it'll "just work" without some extra work.

Yes, to slotting before composable.

Why manual overload?  Cuz major/minor compatibility?

paparazzi/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
swankjesse: Done!

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
swankjesse: Yep. I think that in order to keep the lambda argument last we can’t have defaulted params that precede it?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -208,8 +208,8 @@ class Paparazzi @JvmOverloads constructor(
   }
 
   @JvmOverloads
-  fun snapshot(view: View, name: String? = null) {
-    takeSnapshots(view, name, 0, -1, 1)
+  fun snapshot(view: View, name: String? = null, timeMillis: Long = 0L) {
```
jrodbx, line 211: ubernit: should the arg name be more telling?  maybe...

```suggestion
  fun snapshot(view: View, name: String? = null, offsetMillis: Long = 0L) {
```

jrodbx (approved):

PR #589, opened 2022-09-30
Initial code for Paparazzi interceptors

swankjesse:
I think this is how we should implement features like device frames and image resizing. It also permits other integrated annotations, such as drawing layout bounds onto the output image.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Interceptor.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Interceptor.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2022 Block, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


swankjesse:
Before I land this I think I gotta support snapshots of views only without any device bounds.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Interceptor.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import android.view.View
+import java.awt.image.BufferedImage
+
+/**
+ * Participate in Paparazzi's process of transforming a view into an image.
+ *
+ * Implementations may operate on the input view, the returned bitmap, or both.
+ */
+interface Interceptor {
```
geoff-powell, line 26: I like this api as it works similar to retrofit interceptor. This also seems as a API replacement for `RenderExtension` which is used to render on top of the layout lib output.

geoff-powell: The one issue I know I came across with RenderExtension returning `BufferedImage` is that there were issues finding `java.awt` classes for drawing when creating your own instance within your codebase.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Interceptor.kt
```diff
@@ -0,0 +1,38 @@
...
+  fun intercept(chain: Chain): BufferedImage
+
+  interface Chain {
+    val deviceConfig: DeviceConfig
+    val view: View
```
yschimke, line 31: What is the timing of events with regard to creating the View? Can I set something like a Coil ImageLoader before the view gets rendered?

PR #590, opened 2022-09-30
Apply compose hooks to all snapshot calls

gamepro65:
Add support for ViewGroups which have compose components since AbsComposeView can be added anywhere. If we detect compose on the classpath then apply our hooks instead of only applying to the Composable snapshot.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -28,7 +28,6 @@ import android.view.Choreographer
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import android.widget.FrameLayout
 import androidx.annotation.LayoutRes
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Recomposer
@@ -195,19 +194,9 @@ class Paparazzi @JvmOverloads constructor(
 
   fun snapshot(name: String? = null, composable: @Composable () -> Unit) {
     val hostView = ComposeView(context)
-    // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's
-    // CompositionContext, which requires first finding the "content view", then using that to
-    // find a root view with a ViewTreeLifecycleOwner
-    val parent = FrameLayout(context).apply { id = android.R.id.content }
-    parent.addView(hostView)
-    PaparazziComposeOwner.register(parent)
     hostView.setContent(composable)
 
-    try {
-      snapshot(parent, name)
-    } finally {
-      forceReleaseComposeReferenceLeaks()
-    }
+    snapshot(hostView, name)
   }
 
   @JvmOverloads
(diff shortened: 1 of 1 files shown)


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -622,6 +622,15 @@ class Paparazzi @JvmOverloads constructor(
     private val isVerifying: Boolean =
       System.getProperty("paparazzi.test.verify")?.toBoolean() == true
 
+    private val hasComposeRuntime: Boolean =
```
TWiStErRob, line 625: How often does this change? Should it be cached either `by lazy` or in `companion object`?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -193,20 +191,11 @@ class Paparazzi @JvmOverloads constructor(
...
-    PaparazziComposeOwner.register(parent)
-    hostView.setContent(composable)
-
-    try {
-      snapshot(parent, name)
-    } finally {
-      forceReleaseComposeReferenceLeaks()
+    val hostView = ComposeView(context).apply {
+      setContent(composable)
     }
```
jrodbx, line 196:
```suggestion
    val hostView = ComposeView(context)
    hostView.setContent(composable)
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -292,6 +281,14 @@ class Paparazzi @JvmOverloads constructor(
           // Initialize the choreographer at time=0.
         }
 
+        if (hasComposeRuntime) {
+          // AbstractComposeView will attempt to resolve its parent's CompositionContext, which
```
jrodbx, line 285:
```suggestion
          // // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's CompositionContext, which
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: Will be calculated for each new Paparazzi rule creation for each test method run.  Fine for now, unless it's quick to profile and improve via companion object.

jrodbx (approved):
Clever!

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: Oh, this is already in `companion object`.  Never mind!

TWiStErRob: Oops 🤐, need to expand context on GH 😞

PR #591, opened 2022-09-30
New feature, viewOnly, to ignore the device frame

swankjesse:
This works in two ways:

 - Just viewOnly omits the system bars
 - viewOnly with RenderingMode.EXPAND also changes the output size to match the view

Closes: https://github.com/cashapp/paparazzi/issues/37

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -80,6 +80,10 @@ import java.util.Date
 import java.util.concurrent.TimeUnit
 import kotlin.coroutines.ContinuationInterceptor
 
+/**
+ * @param viewOnly true to omit the device frame. Use this with [RenderingMode.V_SCROLL] or
+ *   [RenderingMode.FULL_EXPAND] to fit the snapshot's size to the view.
+ */
 class Paparazzi @JvmOverloads constructor(
   private val environment: Environment = detectEnvironment(),
   private val deviceConfig: DeviceConfig = DeviceConfig.NEXUS_5,
@@ -89,7 +93,8 @@ class Paparazzi @JvmOverloads constructor(
   private val maxPercentDifference: Double = 0.1,
   private val snapshotHandler: SnapshotHandler = determineHandler(maxPercentDifference),
   private val renderExtensions: Set<RenderExtension> = setOf(),
-  private val supportsRtl: Boolean = false
+  private val supportsRtl: Boolean = false,
+  private val viewOnly: Boolean = false
 ) : TestRule {
   private val logger = PaparazziLogger()
   private lateinit var renderSession: RenderSessionImpl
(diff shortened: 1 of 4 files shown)


swankjesse:
If ever we do a 2.x, it’d be nice to remove `com.android.ide.common.rendering.api.SessionParams.RenderingMode` from Paparazzi’s public API and instead add a new thing that encapsulates both `viewsOnly` and that. As-is the two features are a bit weird as there’s some combinations that don‘t do much.

swankjesse:
# Some interesting outputs!

This button renders just the pixels we need. It’s still scaled up to be 1000px wide, and that’s bad! I’m hoping to address that in the interceptors follow-up. (https://github.com/cashapp/paparazzi/pull/589)
app cash paparazzi_ViewOnlyTest_viewSmallerThanDeviceResolution_full_expand-viewonly

This button word-wraps at the width of the target device. This is using `viewsOnly = true` with `H_SCROLL`:
app cash paparazzi_ViewOnlyTest_viewLargerThanDeviceResolution_v_scroll-viewonly

This ultrawide button doesn’t word wrap at all. It’s using `viewsOnly = true` with `FULL_EXPAND`.
app cash paparazzi_ViewOnlyTest_viewLargerThanDeviceResolution_full_expand-viewonly

This button renders in a canvas as if there’s a device frame, cause we’re using `RenderMode.NORMAL`. I don’t think this is a good combination.
app cash paparazzi_ViewOnlyTest_viewSmallerThanDeviceResolution_normal-viewonly

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -89,7 +93,8 @@ class Paparazzi @JvmOverloads constructor(
   private val maxPercentDifference: Double = 0.1,
   private val snapshotHandler: SnapshotHandler = determineHandler(maxPercentDifference),
   private val renderExtensions: Set<RenderExtension> = setOf(),
-  private val supportsRtl: Boolean = false
+  private val supportsRtl: Boolean = false,
+  private val viewOnly: Boolean = false
```
saket, line 97: `viewOnly` is reminding me of file attributes such as `read only`. 

Considering that our usages of the term `View` is going to only go down as we slowly write everything as composables, can I propose using something else instead? maybe `hideDeviceFrame`?

swankjesse: Good call. Let’s workshop a new name.

Two consequences:
 - no frame
 - no device dimensions

Any other suggestions?

saket: Another suggestion is to match what Android Studio uses for emulators -- `enableDeviceFrame`

BenedictP:
Hey, I also tried to solve this problem in my PR: https://github.com/cashapp/paparazzi/pull/550
So some input from my side:
- As you said it would be better to remove `com.android.ide.common.rendering.api.SessionParams.RenderingMode` because we don't control this class and there is missing functionality. I tried to solve it with another class instead just a boolean.
- I think compose layouts are not affected by your change. `Paparazzi.kt Line 216` is still using `match_parent` for height and width.
- Our main use case is that we want to test Recycler Viewholders. Therefore we want to shrink the y-axis but not the x-axis to check how the view under test behaves on phone/tablet/fablet/foldables.

rharter:
This is also fixed by #497 and here are some of my thoughts:

* I think your usage of `RenderingMode` is backwards. `RenderingMode.SHRINK` is [documented](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:layoutlib-api/src/main/java/com/android/ide/common/rendering/api/SessionParams.java;l=31?q=RenderingMode) to "Shrink canvas to the minimum size that is needed to cover the scene", which sounds like your interpretation of `EXPAND`. I think that `KEEP` is meant to match the size of the device, `EXPAND` is meant to allow the view to be rendered larger than the device (to support scrolling), and `SHRINK` is meant to shrink the canvas to match the view.
* RenderingMode supports defining EXPAND/KEEP/SHRINK along both horizontal and vertical axes, but you seem to be conflating them here with the updated `contentRoot`.

swankjesse:
I’m using `EXPAND` but setting the initial canvas size to 1x1. That way the snapshot can exceed the bounds of the configured device.

As I see them these are the interesting constraints.

 * Layout constraints. For each x,y dimension we can fix to the device, or do unbounded.
 * System bars on or off. These work best if height and width match the device.

Issue #412, opened 2022-04-16
Hard to cleanly override SnapshotHandler


yschimke:
Attempting a temporary fix I tried to decorate the snapshot handler.

It seems awkward to do this currently, since determineHandler is private in Paparazzi.Companion.

```
    @get:Rule
    val paparazzi = Paparazzi(
        deviceConfig = GALAXY_WATCH4_CLASSIC_LARGE,
        theme = "android:ThemeOverlay.Material.Dark",
        maxPercentDifference = 0.1,
        snapshotHandler = WearSnapshotHandler(determineHandler(0.1))
    )

    private val isVerifying: Boolean =
        System.getProperty("paparazzi.test.verify")?.toBoolean() == true

    private fun determineHandler(maxPercentDifference: Double): SnapshotHandler =
        if (isVerifying) {
            SnapshotVerifier(maxPercentDifference)
        } else {
            HtmlReportWriter()
        }
```

yschimke:
I won't need this after https://github.com/cashapp/paparazzi/pull/411, but I assume this is why it is there.

yschimke:
If I want to use my own custom SnapshotHandler, a lot of the types you need to do similar are also internal.

ImageUtils, Snapshot.toFileName and so on.

Is a good way to take this forward to show examples of where I want to use these?  Or they will remain internal, and I should put PRs up to add features I'd like (nothing critical, just nice to haves).

PR #599, opened 2022-10-08
Expose isVerifying and determineHandler

yschimke:
Expose isVerifying and determineHandler. These are useful when implementing a decorating SnapshotHandler. Used for Wear and A11y reasons.

```
    snapshotHandler: SnapshotHandler = WearSnapshotHandler(
        determineHandler(maxPercentDifference),
        round = round
    ),
```

#fixes https://github.com/cashapp/paparazzi/issues/412

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-agp = "7.3.0"
+agp = "7.1.2"
 bytebuddy = "1.12.10"
 composeCompiler = "1.3.0"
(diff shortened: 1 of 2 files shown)


jrodbx:
We definitely need to improve on this customization, but not by exposing these internal details.

Maybe expose all 3 handlers in the api?  val recordHandler, verifyHandler, etc.  Let me have a think on this.

yschimke:
Yep, I'm not blocked, just wanted to get to a smaller set of local changes.

PR #616, opened 2022-10-18
DNM - closed

nak5ive:
## DO NOT MERGE

A draft PR for introducing the `@Paparazzi` annotation processor
see README: https://github.com/nak5ive/paparazzi/tree/akent/annotation-processing/paparazzi/paparazzi-annotation

This PR introduces 2 new modules to the project:
`paparazzi-annotation`: the API for the annotation and configuration
`paparazzi-annotation-processor`: the KSP code responsible for generating test classes

There is also an extension on the plugin to include KSP, the above dependencies, and wire-in the task graph.

### Open items
1. Currently, the annotation is highly configurable. In fact, it exposes pretty much [the entire configuration API](https://github.com/nak5ive/paparazzi/blob/akent/annotation-processing/paparazzi/paparazzi-annotation/src/main/java/app/cash/paparazzi/annotation/api/Paparazzi.kt) for paparazzi. Still need to determine how much configuration is appropriate in this context.

2. Since the original intention of this annotation was to leverage `@Preview` methods, the api dependency needs to be included in the main source set. The configuration landscape for the annotation [needed to be mapped](https://github.com/nak5ive/paparazzi/tree/akent/annotation-processing/paparazzi/paparazzi-annotation/src/main/java/app/cash/paparazzi/annotation/api/config) so as not to produce a slew of duplicate classpath errors. This could be avoided once the `core` / `junit` dependencies are broken apart. We could potentially expose an `api` dependency to support this use-case, but needs to be discussed.


paparazzi/paparazzi-annotation/README.md
```diff
@@ -0,0 +1,161 @@
...
+## Test Parameter Injection
+The `@Paparazzi` annotation also leverages `TestParameterInjector` to execute multiple tests on a handful of configuration parameters.
+
+### Font Scaling
+The annotation exposes the `fontScales` configuration which accepts an array of `Float` values. Setting this parameter will override the value set in the (non-plural) `fontScale` parameter.
+
+```kotlin
+@Paparazzi(
+  name = "fontScaling",
+  fontScales = [1.0f, 2.0f, 3.0f]
```
swankjesse, line 115: (These sample values are dangerous; should be 1.0, 1.1, 1.2 or something. 3x font scale is a nightmare)

PR #628, opened 2022-11-11
Add an interface for intercepting the snapshot and allowing image modifications

yvonnewong:
The use case for making this change is to allow screenshots to be handled in ways other than thumbnail scaling which is helpful for having longer screenshots and not having it scaled

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -193,16 +193,25 @@ class Paparazzi @JvmOverloads constructor(
 
   fun <V : View> inflate(@LayoutRes layoutId: Int): V = layoutInflater.inflate(layoutId, null) as V
 
-  fun snapshot(name: String? = null, composable: @Composable () -> Unit) {
+  fun snapshot(
+    name: String? = null,
+    imageModifier: SnapshotImageModifier = ThumbnailScaleImageModifier,
+    composable: @Composable () -> Unit
+  ) {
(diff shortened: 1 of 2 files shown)


yvonnewong:
@jrodbx PTAL

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -302,7 +313,8 @@ class Paparazzi @JvmOverloads constructor(
             }
 
             val image = bridgeRenderSession.image
-            frameHandler.handle(scaleImage(frameImage(image)))
+            val framedImage = frameImage(image)
```
yschimke, line 316: Should frameImage be moved to a SnapshotImageModifier also?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/SnapshotImageModifier.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Interface that allows access to the image from the bridge session for modification
+ */
+interface SnapshotImageModifier {
+
+  /**
+   * @param image The image from the bridge session to modify
+   * @return modified image
+   */
+  fun modifyImage(image: BufferedImage): BufferedImage
```
yschimke, line 15: The frameImage I mention above needs deviceConfig, should this be considered context and passed as a parameter to `modifyImage`?

jrodbx:
Obsoleted by https://github.com/cashapp/paparazzi/pull/1382

PR #642, opened 2022-12-03
Fix RecyclerView issue due to Layoutlib Dolphin change

jrodbx:
Relevant commit: https://android.googlesource.com/platform/frameworks/layoutlib/+/cc1a3b598a00b9bdbeae0e378d07ba997546d979

Android Studio implementation:  https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:android/src/org/jetbrains/android/uipreview/ViewLoader.java;l=135


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
```diff
@@ -88,11 +89,26 @@ internal class PaparazziCallback(
...
+    constructorArgs: Array<Any>
   ): Any? {
-    val viewClass = Class.forName(name)
-    val viewConstructor = viewClass.getConstructor(*constructorSignature)
-    viewConstructor.isAccessible = true
-    return viewConstructor.newInstance(*constructorArgs)
+    // RecyclerView.Adapter is an abstract class, but its instance is needed for RecyclerView to work correctly.
+    // So, when LayoutLib asks for its instance, we define a new class which extends the Adapter class.
+    // We check whether the class being loaded is the support or the androidx one and use the appropriate adapter that references to the
+    // right namespace.
```
jrodbx, line 102: @JakeWharton intercepting the call was too much of a hassle.  so this copies the AS implementation with a dummy adapter (during inflation code paths only, in BridgeInflater), until the actual adapter eventually gets wired anyway.

geoff-powell (approved):

PR #644, opened 2022-12-04
Compose 1.3 and Kotlin 1.7.20

jrodbx:


gradle/libs.versions.toml
```diff
@@ -2,10 +2,10 @@
 agp = "7.3.1"
 androidTools = "30.3.1" # == 23.0.0 + agp version
 bytebuddy = "1.12.10"
-composeCompiler = "1.3.0"
+composeCompiler = "1.3.2"
```
TWiStErRob, line 5: Does `composeUi-material` need to be updated?
https://mvnrepository.com/artifact/androidx.compose.material/material?repo=google

JakeWharton (approved):
Does this actually fix those issues? Or did the layoutlib bump fix them? Libraries don't need to track the latest versions of Compose in order to work on newer versions of Compose in a downstream project.

geoff-powell (approved):
Along with what jake said, is there a test case we can add that captures the two linked issues so we catch this issue if it happens again?

jrodbx:
> Along with what jake said, is there a test case we can add that captures the two linked issues so we catch this issue if it happens again?

A test case that ensures that the version code doesn't drop to a lower max version in the future doesn't seem useful to me; plus this wasn't a bug per se, but rather catching up with dependency version hell.

gradle/libs.versions.toml
TWiStErRob: You're right, [it is now](https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html).

> The first library to break away from the single Compose version is the Compose Compiler. Today we’re releasing the [1.2.0 stable version](https://developer.android.com/jetpack/androidx/releases/compose-compiler) that brings support for Kotlin 1.7.0! The release is both backwards and forwards compatible with the Compose UI libraries and the Compose Runtime library. This means you can upgrade your Compose Compiler to 1.2.0 stable and use Kotlin 1.7.0, while leaving your other Compose libraries on their current version, for example 1.1.0 stable.

swankjesse (approved):

PR #647, opened 2022-12-06
Gradle 7.6

jrodbx:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-all.zip
```
JakeWharton, line 3: IntelliJ can attach sources to Gradle now (the same as it would any binary dependency from Maven) so using bin means drastically smaller and faster downloads.

JakeWharton (approved):

gradle/wrapper/gradle-wrapper.properties
jrodbx: will fix in a follow up

PR #648, opened 2022-12-07
Use compileOnly for kotlin/agp plugins

jrodbx:


build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
 
   dependencies {
     classpath libs.plugin.kotlin
-    classpath libs.plugin.android
+    classpath 'com.android.tools.build:gradle:7.3.1'
```
eric-labelle, line 13: Not sure if this change was intented?  `7.4-rc1` is in the toml.
`/paparazzi/build.gradle` and your new compileOnly still uses the 7.4 so you might get overridden.

TWiStErRob: I have a feeling this was him testing the solution.

See the last `classpath` below and https://github.com/cashapp/paparazzi/blob/master/settings.gradle#L9 where that plugin is being replaced.

paparazzi/paparazzi-gradle-plugin/build.gradle
```diff
@@ -11,17 +11,28 @@ gradlePlugin {
...
+
+tasks.named('pluginUnderTestMetadata').configure {
+  getPluginClasspath().from(configurations.fixtureClasspath)
+}
+
 dependencies {
   compileOnly gradleApi()
-  implementation platform(libs.kotlin.bom)
   implementation libs.plugin.kotlin
-  implementation libs.plugin.android
```
TWiStErRob, line 18: please merge this PR, I can't use Android Studio Dolphin with 1.2.0-SNAPSHOT, because it doesn't support 7.4:
> The project is using an incompatible version (AGP 7.4.0-rc01) of the Android Gradle plugin. Latest supported version is AGP 7.3.1
> [See Android Studio & AGP compatibility options.](https://developer.android.com/studio/releases#android_gradle_plugin_and_android_studio_compatibility)

I'm not even sure of a workaround yet, because I'm using `plugins {}` so there's no `classpath(...) { exclude }`.

TWiStErRob: Tried this, but it doesn't apply transitively and I don't want to list all modules:
```gradle
buildscript {
	configurations.classpath.configure {
		resolutionStrategy { 
			force(libs.android.gradle)
			force(libs.android.lint.common)
    }
```

This crazy hack works 🤦‍♂️:
```gradle
buildscript {
	configurations.classpath.configure {
		resolutionStrategy { 
			eachDependency { 
				if (requested.version == "30.4.0-rc01") {
					useVersion("30.3.1")
				}
				if (requested.version == "7.4.0-rc01") {
					useVersion("7.3.1")
				}
			}
		}
	}
}
```

build.gradle
TWiStErRob: I think I understand this more now. As I ran into an issue with this.

The the paparazzi-gradle-plugin can use the latest version, but for testing and building the paparazzi AAR an older version can be used.

@jrodbx I suggest splitting it into 3:
 * `libs.plugin.android.build` (for AAR build)
 * `libs.plugin.android.test` (for test-fixtures)
 * `libs.plugin.android.api` (for dependency)

The `api` should be behind a bit (stable), the `build` can be the latest alpha, and the `test` is the best compatible version.

yogurtearl:
Looks like this PR is superseded by ->
- https://github.com/cashapp/paparazzi/pull/1105

BrianGardnerAtl (approved):
I've got little context here but this seems fine? Cutting out the dependencies at runtime seems like an easy win

geoff-powell (approved):

joshfriend:
@BrianGardnerAtl the context is that the strict dependency that paparazzi places on AGP and Kotlin makes upgrading any of the three rather difficult because you have to do them all at once, or add some ugly workarounds.

BrianGardnerAtl:
I think that makes sense. Is that the requirement on the consuming project since paparazzi was declaring the dependencies outside of just compile time?

joshfriend:
yeah, any build that uses papparazzi gets forced to use the same AGP/Kotlin versions as paparazzi unless you apply multiple layers of version constraints in gradle.

jrodbx:
> @BrianGardnerAtl the context is that the strict dependency that paparazzi places on AGP and Kotlin makes upgrading any of the three rather difficult because you have to do them all at once, or add some ugly workarounds.

Keep in mind that any API changes in AGP that the plugin depends on will result in the need for a strict dependency on a minimum particular version anyway.  All that means is there may be breaking surprise changes at some point in the future.  I plan to explore adding a test matrix for this at some point.

PR #650, opened 2022-12-10
Fix ClassCastException in Compose 1.4.x

### LINKED ISSUE #641: java.lang.ClassCastException with compose-runtime 1.4.0-alpha02
**Description**
Paparazzi tests fail with `java.lang.ClassCastException` when using compose-runtime 1.4.0-alpha02

**Steps to Reproduce**
- Update compose-runtime to 1.4.0-alpha02
- Try to run a Paparazzi test

**Additional information:**
The internals of `Recomposer.kt` were recently [updated](https://android-review.googlesource.com/c/platform/frameworks/support/+/2258823), which [changed](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/runtime/runtime/src/commonMain/kotlin/androidx/compose/runtime/Recomposer.kt;l=203;drc=98f2641f492c7ffc91e77e39d60b77909669f0ea) the type of `snapshotInvalidations` from a `val MutableList` to a `var MutableSet`.

Paparazzi retrieves this field by reflection and [casts](https://github.com/cashapp/paparazzi/blob/c7c19125fa05f044c5d865f42f970776f3171fd8/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt) it to a `MutableList`, which causes a `java.lang.ClassCastException` due to the update.

ZacSweers:
Resolves #641 in a backward-compatible way

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -572,7 +572,7 @@ class Paparazzi @JvmOverloads constructor(
       val snapshotInvalidations = recomposer.javaClass
         .getDeclaredField("snapshotInvalidations")
         .apply { isAccessible = true }
-        .get(recomposer) as MutableList<*>
+        .get(recomposer) as MutableCollection<*>
       compositionInvalidations.clear()
       snapshotInvalidations.clear()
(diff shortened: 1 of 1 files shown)


JakeWharton:
We need to get rid of a bunch of these CI shards...

JakeWharton (approved):

PR #655, opened 2022-12-14
Expose flag to show/hide system ui

jrodbx:
Complements #497 by taking the key bit from #591 allowing testers to render widgets without the system ui status and navigation bar.

Closes https://github.com/cashapp/paparazzi/pull/591. Closes: https://github.com/cashapp/paparazzi/issues/37

Examples from CashApp:
Screen Shot 2022-12-14 at 3 51 12 AM
Screen Shot 2022-12-14 at 3 51 26 AM


swankjesse (approved):
Love it.

I think we probably need a follow up to control how the widget sizes itself. Does wide text wrap? Are narrow components rendered in their intrinsic width or do they expand to the device width?

jrodbx:
> I think we probably need a follow up to control how the widget sizes itself. Does wide text wrap? Are narrow components rendered in their intrinsic width or do they expand to the device width?

The sizing will come from the DeviceConfig's properties which directly affect the canvas size.  This seems natural since a widget might render differently on a Nexus 5 vs a Pixel 3.

TWiStErRob:
This was already possible with `Paparazzi(deviceConfig = DeviceConfig.PIXEL_2.copy(softButtons = false))`, now we have a duplicate API surface? or is there actually a difference between the two?

Proven equivalence by recording with softButtons = false and verifying with showSystemUi = false. Same equivalence stands in both NORMAL and SHIRNK modes.

PR #656, opened 2022-12-14
Enable setting a custom parent view in compose paparazzi snapshot method

elizrus:
I'd like to be able to set my own `FrameLayout` as a parent view when snapshotting a composable.

My use case is a little niche- I need my parent view to inherit an interface based on some assumptions in my compose code.

Alternatively I could write my own little snapshot method, but a lot of the goodies like `PaparazziComposeOwner` are private so I'd have to do a lot of duplication.


elizrus:
After discussing with @saket , I realize that https://github.com/cashapp/paparazzi/pull/590 is the better solution here

PR #658, opened 2022-12-15
Update linux action to ubuntu-latest

gamepro65:
Looks like ubuntu-18.04 is deprecated and will not be available after April 1, 2023 https://github.com/actions/runner-images/issues/6002. Removed the comment since we didn't run into any issues with the update.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,10 +13,7 @@ jobs:
         os: [
           macOS-latest,
           windows-latest,
-          # Copied from SqlDelight: https://github.com/cashapp/sqldelight/blame/master/.github/workflows/PR.yml#L13-L18
-          # TL;DR looks like libraries installed on ubuntu-latest conflicts, resulting in failed builds
-          # Also, see: https://github.com/touchlab/SQLiter/pull/38#issuecomment-867171789
-          ubuntu-18.04
+          ubuntu-latest
         ]
         java-version: [11, 12, 16, 18]
 


jrodbx (approved):

.github/workflows/build.yml
```diff
@@ -13,10 +13,7 @@ jobs:
         os: [
           macOS-latest,
           windows-latest,
-          # Copied from SqlDelight: https://github.com/cashapp/sqldelight/blame/master/.github/workflows/PR.yml#L13-L18
-          # TL;DR looks like libraries installed on ubuntu-latest conflicts, resulting in failed builds
-          # Also, see: https://github.com/touchlab/SQLiter/pull/38#issuecomment-867171789
-          ubuntu-18.04
+          ubuntu-22.04
```
TWiStErRob, line 16: Considering
* the context around it (both others referencing "latest")
* `ubuntu-latest` is now 22 (it was 20 when this was copied over  
Ref 22 being latest: https://github.com/actions/runner-images/  
and announcements: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

It think:
```suggestion
          ubuntu-latest
```

TWiStErRob: (or fix others to specific numbers too?)

gamepro65: Good point, will change to latest. Thanks

PR #661, opened 2022-12-16
Remove setting LayoutParams on ComposeView

gamepro65:
We don't need to apply extra LayoutParams to ComposeView since it is a child of our `contentRoot` which already sets the expected LayoutParams for the RenderingMode.

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/test/projects/widgets/build.gradle
+++ b/paparazzi/paparazzi-gradle-plugin/src/test/projects/widgets/build.gradle
@@ -18,8 +18,15 @@ android {
   defaultConfig {
     minSdk libs.versions.minSdk.get() as int
   }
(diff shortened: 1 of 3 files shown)


paparazzi/paparazzi-gradle-plugin/src/test/projects/widgets/build.gradle
```diff
@@ -18,8 +18,15 @@ android {
...
+    compose true
+  }
+  composeOptions {
+    kotlinCompilerExtensionVersion libs.versions.composeCompiler.get()
+  }
 }
 
 dependencies {
   testImplementation libs.testParameterInjector
+  implementation libs.composeUi.material
```
jrodbx, line 31: nit: implementation above testImplementation

jrodbx (approved):

PR #664, opened 2022-12-17
Build on latest JDK, but run tests on several

jrodbx:
https://jakewharton.com/build-on-latest-java-test-through-lowest-java/

### DIFF
--- a/paparazzi/build.gradle
+++ b/paparazzi/build.gradle
@@ -95,6 +95,29 @@ subprojects {
       ])
     }
   }
+
+  // Avoid the native projects in 'libs'
+  plugins.withId('org.jetbrains.kotlin.jvm') {
+    // Create a test task for each supported JDK.
+    [11, 12, 16, 18].each { majorVersion ->
+      def jdkTest = tasks.register("testJdk$majorVersion", Test) {
+        javaLauncher = javaToolchains.launcherFor {
+          languageVersion = JavaLanguageVersion.of(majorVersion)
+        }
+
+        description = "Runs the test suite on JDK $majorVersion"
+        group = LifecycleBasePlugin.VERIFICATION_GROUP
+
+        // Copy inputs from normal Test task.
+        def testTask = tasks.getByName("test")
+        classpath = testTask.classpath
+        testClassesDirs = testTask.testClassesDirs
+      }
+      tasks.named("check").configure {
+        dependsOn(jdkTest)
+      }
+    }
+  }
 }
 
 tasks.register("clean", Delete).configure {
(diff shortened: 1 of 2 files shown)


paparazzi/build.gradle
```diff
@@ -95,6 +95,29 @@ subprojects {
...
     }
   }
+
+  // Avoid the native projects in 'libs'
+  plugins.withId('org.jetbrains.kotlin.jvm') {
+    // Create a test task for each supported JDK.
+    [11, 12, 16, 18].each { majorVersion ->
+      def jdkTest = tasks.register("testJdk$majorVersion", Test) {
+        javaLauncher = javaToolchains.launcherFor {
+          languageVersion = JavaLanguageVersion.of(majorVersion)
```
TWiStErRob, line 105: Will this also change GradleRunner's JVM?

TWiStErRob, line 105: Re CI failure, these might come in handy:
https://github.com/actions/setup-java#install-multiple-jdks
https://github.com/gradle/gradle/issues/14903#issuecomment-943421254

paparazzi/build.gradle
```diff
@@ -95,6 +95,29 @@ subprojects {
       ])
     }
   }
+
+  // Avoid the native projects in 'libs'
+  plugins.withId('org.jetbrains.kotlin.jvm') {
+    // Create a test task for each supported JDK.
+    [11, 12, 16, 18].each { majorVersion ->
```
JakeWharton, line 102: Should really include 17 as some vendors declared that an LTS release so the probability that someone will have it is high.

JakeWharton (approved):

paparazzi/build.gradle
jrodbx: Good question, I'm not sure.  @JakeWharton ?

TWiStErRob: When I was having "fun" with Java and Gradle versions, I added this println to always see what's up:
https://github.com/TWiStErRob/net.twisterrob.gradle/blob/master/test/src/main/kotlin/net/twisterrob/gradle/test/GradleRunnerRule.kt#L159

(I didn't yet do what Jake suggested in his blog in my Gradle setup.)

JakeWharton:
I assume the failure is because AdoptOpenJDK (which doesn't even exist anymore) deletes versions which they don't provide as LTS. You're much better off with Zulu since they never delete any builds. All major versions, all minor versions, all patch versions, all architectures. They're all there.

You can get Gradle to provision Azul Zulu via https://github.com/gradle/foojay-toolchains and specifying "Zulu" as the vendor.

PR #668, opened 2022-12-22
Register LifecycleOwner and SavedStateRegistryOwner for all Views

saket:
`LifecycleOwner` and `SavedStateRegistryOwner` aren't specific to Compose UI. They can also be used by Views.


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
```diff
@@ -0,0 +1,48 @@
...
+      val lifecycleOwner = PaparazziLifecycleOwner()
+      ViewTreeLifecycleOwner.set(view, lifecycleOwner)
+
+      if (hasSavedStateRegistryOwnerRuntime) {
+        view.setViewTreeSavedStateRegistryOwner(PaparazziSavedStateRegistryOwner(lifecycleOwner))
+      }
+    }
+  }
+
+  private fun isPresentInClassPath(className: String): Boolean {
```
jrodbx, line 28:
```suggestion
  private fun isPresentInClasspath(className: String): Boolean {
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.LifecycleRegistry
+import androidx.lifecycle.ViewTreeLifecycleOwner
+import androidx.savedstate.SavedStateRegistry
+import androidx.savedstate.SavedStateRegistryController
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+
+internal object PaparazziViewOwners {
```
jrodbx, line 13: can we get rid of this and have separate owners?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
jrodbx: move this out to Paparazzi, since it's the consumer and therefore the classpath checker

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+private class PaparazziLifecycleOwner : LifecycleOwner {
+  private val lifecycleRegistry = LifecycleRegistry(this).apply { currentState = Lifecycle.State.CREATED }
+  override fun getLifecycle(): Lifecycle = lifecycleRegistry
+}
+
+private class PaparazziSavedStateRegistryOwner(
+  private val lifecycleOwner: LifecycleOwner
```
jrodbx, line 44: does this need the lifecycle owner?

saket: Yes. `SavedStateRegistryOwner` extends `LifecycleOwner` so it is dependent. `OnBackPressedDispatcher` will have the same dependency.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
saket: Should I inline `PaparazziViewOwners.register()` into its call site at `Paparazzi#L284`? That'll also let me move `isPresentInClasspath()` to Paparazzi.

https://github.com/cashapp/paparazzi/blob/19d11c148b10599eb8b8f6cc2570227eac19dd8c/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt#L284

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
saket: continuing this discussion in https://github.com/cashapp/paparazzi/pull/668#discussion_r1055912569 as I believe these two are inter-related.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
saket: The owners are already separate btw. I am using `PaparazziViewOwners` only to aggregate them.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
```diff
@@ -0,0 +1,50 @@
...
+  fun register(view: View) {
+    if (hasLifecycleOwnerRuntime) {
+      val lifecycleOwner = PaparazziLifecycleOwner()
+      ViewTreeLifecycleOwner.set(view, lifecycleOwner)
+
+      if (hasSavedStateRegistryOwnerRuntime) {
+        view.setViewTreeSavedStateRegistryOwner(PaparazziSavedStateRegistryOwner(lifecycleOwner))
+      }
+
+      lifecycleOwner.registry.currentState = Lifecycle.State.CREATED
```
jrodbx, line 26: nit: should this move to before line 20 (maintaining the previous setup/execution order) or does it not matter?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
jrodbx: Yea, I think PaparazziViewOwners can be inlined.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -288,7 +281,7 @@ class Paparazzi @JvmOverloads constructor(
           // CompositionContext, which requires first finding the "content view", then using that
           // to find a root view with a ViewTreeLifecycleOwner
           viewGroup.id = android.R.id.content
-          PaparazziComposeOwner.register(viewGroup)
+          PaparazziViewOwners.register(viewGroup)
```
jrodbx, line 284: should this be moved out of this block to support all views, as suggested by the commit message?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
saket: The lifecycle state can't be changed from `INITIALIZED` until `SavedStateRegistryOwner` has finished restoring all state. I've added a comment. Let me know if you think that our failing tests were sufficient?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
saket: Whoops thanks for catching! I lost that change in a git undo.

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
saket: Done!

jrodbx (approved):

PR #670, opened 2022-12-28
Change the project name when building from Intellij or Android Studio

aaalaniz:
This change set updates the `paparazzi/settings.gradle` root project name to `paparrazzi` when the project is built from Intellij or Android Studio. Previously when attempting to create run configurations, the IDE would set the task to `paparazzi-libs:` in accordance with the root project name which would fail to run because the task did not match the directory structure. This change provides a workaround for the following [gradle issue](https://github.com/gradle/gradle/issues/16608) by checking a property `idea.active` which is only set when building the project from Intellij or Android Studio. This approach is documented in this [Stack Overflow post](https://stackoverflow.com/questions/63409895/configure-gradle-plugin-based-on-future-tasks/63411337#63411337).

Screenshot 2022-12-28 at 7 53 38 AM

### DIFF
--- a/paparazzi/settings.gradle
+++ b/paparazzi/settings.gradle
@@ -1,4 +1,13 @@
-rootProject.name = 'paparazzi-libs'
+/**
+ * Set the root project name to `paparazzi` when building from Intellij or Android Studio so that run executions created
+ * via context actions resolve to the task that matches the actual project directory structure. This is a workaround for
+ * https://github.com/gradle/gradle/issues/16608.
+ */
+if (System.getProperty("idea.active")) {
+  rootProject.name = 'paparazzi'
+} else {
+  rootProject.name = 'paparazzi-libs'
+}
 
 include ':paparazzi'
 include ':paparazzi-agent'


aaalaniz:
Yikes, this completely breaks CI. I will take a look and see if there is a better alternative to enable seamless testing from IDE context actions.

paparazzi/settings.gradle
```diff
@@ -1,5 +1,3 @@
-rootProject.name = 'paparazzi-libs'
```
TWiStErRob, line 1: > Project accessors enabled, but root project name not explicitly set for 'paparazzi'. Checking out the project in different folders will impact the generated code and implicitly the buildscript classpath, breaking caching.

TWiStErRob, line 1: Looks like it gets confused on the root project and the subproject both being called paparazzi. IDEA has had a long issue where included builds are misconfigured and don't work as well. I think you might be experiencing that issue. Which version of AS/IJ are you using?

TWiStErRob: Even if you manage to use `paparazzi` as the name, make it explicit.

aaalaniz: Agreed. I'll try. Thanks!

aaalaniz: IntelliJ IDEA 2022.3.1 (Community Edition)
Build #IC-223.8214.52, built on December 20, 2022
Runtime version: 17.0.5+1-b653.23 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.1
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 8
Metal Rendering is ON
Registry:
    debugger.new.tool.window.layout=true
    ide.experimental.ui=true

Non-Bundled Plugins:
    IdeaVIM (2.0.0)

Kotlin: 223-1.7.21-release-272-IJ8214.52

aaalaniz: Android Studio Flamingo | 2022.2.1 Canary 10
Build #AI-222.4459.24.2221.9409768, built on December 15, 2022
Runtime version: 17.0.4.1+0-17.0.4.1b469.62-9127311 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 13.1
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 8
Metal Rendering is ON
Registry:
    external.system.auto.import.disabled=true
    ide.text.editor.with.preview.show.floating.toolbar=false
    gradle.version.catalogs.dynamic.support=true

Non-Bundled Plugins:
    IdeaVIM (1.11.1)

TWiStErRob: Hmm, these are way too new, unless the issue regressed. It's interesting that the compilation fails as it does. I think that might be a Gradle bug:
repro: name root project same as one of the submodules and enable project accessors. I'll try to repro and report.

TWiStErRob: Well, no good news on the Gradle front... it is simply not possible: https://github.com/gradle/gradle/issues/16608

aaalaniz: Interesting 🤔 

Is the solution to rename the `paparazzi` directory to `paparazzi-libs`? Alternatively, all the modules under `paparazzi/` could be moved to the root directory.

TWiStErRob: It needs to be in a subdir, so that the top level can include it. Rename might work. OR rename :paparazzi to something else, but that would affect artifact names I guess or make the dir-artifact naming inconsistent.

TWiStErRob: Due to the above, it'll never compile with name = paparazzi :(. Not even in IDEA sync/run, because that just executes Gradle. With your recent change CI will be green, but locally it should fail sync/build/run.

aaalaniz: Indeed. This is a tricky issue 😅. I'll keep trying though!

aaalaniz: Perhaps the only way is to add a system gradle property that updates the root project name for local development. I was hoping the experience would be more seamless though.

TWiStErRob: It doesn't matter how you set it, this "active" thing is also a property. Did you try syncing with the flag, does it work locally?

TWiStErRob: I think maybe we can take a step back. I have a feeling it'll work even without a code change. Are you importing the repository root? If so, try importing only the paparazzi subfolder. Try this on master.

aaalaniz: Actually, after checking this a little closer, I was able to build with these changes from both IntelliJ and Android Studio. Maybe this is working as intended?

aaalaniz: Perhaps I should create a separate issue that demonstrates the problem I'm trying to address. Essentially, without these changes, when you import the project and attempt to run a test using context actions from a file or from a test directory, the generated configuration will contain `paparazzi-libs` as the root task name instead of `paparazzi`. After these changes, I am able to run tests just fine from the AS and IntelliJ

TWiStErRob: That makes sense, but that sounds like a bug in the IJ core (create issue in youtrack.com for sure):
Based on OP and your description above IJ confuses the `rootProject.name` and the included build's name. It's trying to be too smart and uses the included build's `rootProject.name` rather than the name that it actually has in at the point of include.

What you're trying to do is a workaround for this bug, but it's not compatible with Gradle. I'm surprised that it works, I'll have a go when I get back after holidays.

What I proposed above is another workaround: rather than importing "repo/" in IJ, import "repo/paparazzi/" folder. At this point your only editing the "libs" project and it should be listed exactly as is, because there's no included build in play. My base assumption around included build might be wrong.

I'm curious if everyone experiences what you're seeing. Definitely have a look at YouTrack as well because this behaviour sounds buggy.

paparazzi/settings.gradle
```diff
@@ -1,4 +1,13 @@
...
+/**
+ * Set the root project name to `paparazzi` when building from Intellij or Android Studio so that run executions created
+ * via context actions resolve to the task that matches the actual project directory structure. This is a workaround for
+ * https://github.com/gradle/gradle/issues/16608.
+ */
+if (System.getProperty("idea.active")) {
```
TWiStErRob, line 6: Is this a documented prop? Where did you find it? (Just curious, because I can't seem to find good description)

jrodbx:
Apologies, but I don't understand the proposed change here. This repo is set up in order for the sample to dynamically use the plugin at build time via a nested build, as opposed to first deploying the artifact to a repository.  As a result, running the plugin tests requires setting the project dir (via -p) explicitly.  Does this not work via a run configuration?

TWiStErRob:
It probably does, but it should "just work" when pressing the play button ad-hoc on any Test annotated method. It is an IDEA v Included build issue, we were trying to find a workaround so it "just works".

aaalaniz:
@TWiStErRob is correct. Run configurations do work, but not when create one from the IDE say from a single test or test suite. The auto-generated root module is `paparazzi-libs` when created from the IDE, so as a user, you have to navigate to the run configuration and manually change the root path to `paparazzi` to match the directory structure and then you can run. I don't think this change should be merged in. I think the actual solution if you want it to "just work" is to import the `paparazzi` directory into the IDE and not use the composite build or you can just manually change the path when you create run configurations.

PR #681, opened 2023-01-15
Update to AGP 7.4.0 stable

joshfriend:
Just wanted to make sure this happens before the next release

Also noted that layoutlib did not change between beta4 and the stable tag for EE, both point to the same commit https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/refs/tags/studio-2022.1.1


TWiStErRob:
> Just wanted to make sure this happens before the next release

Do you have an idea when that will be?

JakeWharton (approved):

jrodbx:
Thanks @joshfriend !

Issue #675, opened 2023-01-06
Compose 1.3.0 & Row loop cause OutOfMemoryError


Masaki-U:
**Description**
Failed recordPaparazzi command include specified layout.

**Steps to Reproduce**

1. Add 'Bug Sample Specified Layout'
2. Run recordPaparazziHoge
3. Failed OutOfMemoryError

**Expected behavior**
Success take a Screen Shot.

**Additional information:**
- Paparazzi Version: 1.1.0
- OS: Mac OSX
- Compile SDK: 33
- Gradle Version: 7.6
- Android Gradle Plugin Version: 7.4.0-rc01

Bug Sample Specified Layout

```
@Preview
@Composable
fun Preview() {
    Column {
        listOf(
            "test1",
            "test2",
        ).forEach {
            Row {
                Text(it)
            }
        }
    }
}
```

ErrorLog

```
      java.lang.OutOfMemoryError: Java heap space
        at java.desktop/java.awt.image.DataBufferInt.<init>(DataBufferInt.java:75)
        at java.desktop/java.awt.image.Raster.createPackedRaster(Raster.java:467)
        at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(DirectColorModel.java:1032)
        at java.desktop/java.awt.image.BufferedImage.<init>(BufferedImage.java:333)
        at app.cash.paparazzi.internal.ImageUtils.scale(ImageUtils.kt:300)
        at app.cash.paparazzi.Paparazzi.scaleImage(Paparazzi.kt:397)
        at app.cash.paparazzi.Paparazzi.access$scaleImage(Paparazzi.kt:83)
        at app.cash.paparazzi.Paparazzi$takeSnapshots$1$2.invoke(Paparazzi.kt:313)
        at app.cash.paparazzi.Paparazzi$takeSnapshots$1$2.invoke(Paparazzi.kt:306)
        at app.cash.paparazzi.Paparazzi.withTime(Paparazzi.kt:347)
        at app.cash.paparazzi.Paparazzi.takeSnapshots(Paparazzi.kt:306)
        at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:222)
        at app.cash.paparazzi.Paparazzi.snapshot$default(Paparazzi.kt:221)
        at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:214)
[...]

PR #682, opened 2023-01-17
Execute Handler callbacks after snapshots to clean up compose references

gamepro65:
Setup a repro case with a long running test suite of all compose snapshots and verified that with the current release function we'd leak a reference inside WindowRecomposer's LifecycleEventObserver where the call to `recomposer.runRecomposeAndApplyChanges()` would hang preventing its job from cleaning up after being cancelled. By running Handler_Delegate.executeCallbacks() an additional time after removing ComposeView from the hierarchy we give AndroidUiDispatcher another tick to run the subsequent cleanup it needs to release its compose references.

Unfortunately RenderSessionImpl holds a strong reference to its last snapshot view inside its `mViewInfoList` and `mSystemViewInfoList` until we teardown Paparazzi without a way to properly clear the reference between snapshot calls unless we teardown / recreate the RenderSessionImpl. This is just temporary as subsequent snapshots in the same test clear out the previous. Given the teardown issue the included test is to at least ensure we are not leaking ComposeView between class tests.

This should resolve https://github.com/cashapp/paparazzi/issues/675

Heap dump comparisons while running the long standing test

Before showing 400+ ComposeView's hanging around
Screenshot 2023-01-17 at 4 33 07 PM

After showing just 2 (one is the current test, other is the one left in memory in RenderSessionImpl until teardown)
Screenshot 2023-01-17 at 4 33 46 PM


paparazzi/paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeReferenceLeakTest.kt
```diff
@@ -0,0 +1,41 @@
...
+          HelloPaparazzi()
+        }
+
+        paparazzi.snapshot(this)
+      }
+    )
+  }
+
+  companion object {
+
```
jrodbx, line 28:
```suggestion
```

jrodbx (approved):

jrodbx:
awesome stuff!  thank you!

paparazzi/paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeReferenceLeakTest.kt
```diff
@@ -0,0 +1,40 @@
...
+    @AfterClass
+    @JvmStatic
+    fun teardown() {
+      assert(weakComposeView.get() != null)
+
+      System.gc()
+
+      assert(weakComposeView.get() == null)
+    }
```
TWiStErRob, line 38: @gamepro65 This is flaky, it could be freed up any time depending on when GC executes. We can't assert that it's non-null because it's a weak reference!

I got a random failure: https://github.com/TWiStErRob/paparazzi/actions/runs/3967102579/jobs/6798680398
```
java.lang.AssertionError: Assertion failed
	at app.cash.paparazzi.plugin.test.ComposeReferenceLeakTest$Companion.teardown(ComposeReferenceLeakTest.kt:33)
	at app.cash.paparazzi.plugin.test.ComposeReferenceLeakTest.teardown(ComposeReferenceLeakTest.kt)
```

TWiStErRob: It is possible to keep this assertion but then it's necessary to fully control the references:
```kotlin
class ComposeReferenceLeakTest {
  @get:Rule
  val paparazzi = Paparazzi()

  @Test
  fun cleansUpComposeReferences() {
    composeView = ComposeView(paparazzi.context).apply {
      setContent {
        HelloPaparazzi()
      }
      paparazzi.snapshot(this)
    }
    weakComposeView = WeakReference(composeView)
  }

  companion object {
    private var composeView: ComposeView? = null
    private lateinit var weakComposeView: WeakReference<ComposeView>

    @AfterClass
    @JvmStatic
    fun teardown() {
      assert(weakComposeView.get() != null)

      composeView = null
      System.gc()

      assert(weakComposeView.get() == null)
    }
```
this would still fail if there's another reference held somewhere.

gamepro65: 😅  Will update the test, thanks for calling out!

TWiStErRob: Nice response, thanks for fixing.

PR #683, opened 2023-01-17
Ensure LifecycleOwner is available for ComposeView

gamepro65:
Compose installs its WindowRecomposer on the contentChild

`
The content child is the view that is either a direct child of the view with id android.R.id.content (and was therefore set as a content view into an activity or dialog window) or the root view of the window.
`

When using renderer extensions `view` becomes a child of `modifiedView` which prevents compose from finding the LifecycleOwner.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -291,13 +291,13 @@ class Paparazzi @JvmOverloads constructor(
 
         if (hasLifecycleOwnerRuntime) {
           val lifecycleOwner = PaparazziLifecycleOwner()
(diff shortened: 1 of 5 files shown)


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -291,13 +291,13 @@ class Paparazzi @JvmOverloads constructor(
 
         if (hasLifecycleOwnerRuntime) {
           val lifecycleOwner = PaparazziLifecycleOwner()
-          ViewTreeLifecycleOwner.set(view, lifecycleOwner)
+          ViewTreeLifecycleOwner.set(modifiedView, lifecycleOwner)
```
saket, line 294: Can we leverage name shadowing to prevent these errors?

saket (approved):

jrodbx (approved):

PR #684, opened 2023-01-17
Check for newer classes in classpath for setting view owners

saket:


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -633,10 +633,10 @@ class Paparazzi @JvmOverloads constructor(
       "androidx.lifecycle.ViewTreeLifecycleOwner"
     )
     private val hasSavedStateRegistryOwnerRuntime = isPresentInClasspath(
-      "androidx.savedstate.SavedStateRegistryOwner"
+      "androidx.savedstate.ViewTreeSavedStateRegistryOwner"
```
saket, line 636: `View#setViewTreeSavedStateRegistryOwner(owner)` is contained in `ViewTreeSavedStateRegistryOwner` and was introduced in [androidx.savedstate.*:1.2.0](https://developer.android.com/jetpack/androidx/releases/savedstate#1.2.0). 

Should we instead support a slightly older version of `1.1.0` by using the old `ViewTreeSavedStateRegistryOwner.set(view, owner)` API?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -633,10 +633,10 @@ class Paparazzi @JvmOverloads constructor(
...
     )
     private val hasAndroidxActivityRuntime = isPresentInClasspath(
-      "androidx.activity.OnBackPressedDispatcherOwner"
+      "androidx.activity.ViewTreeOnBackPressedDispatcherOwner"
```
saket, line 639: `View#setViewTreeOnBackPressedDispatcherOwner(owner)` is contained in `ViewTreeOnBackPressedDispatcherOwner` and was introduced in [androidx.activity.*:1.5.0](https://developer.android.com/jetpack/androidx/releases/activity#1.5.0).

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: in general, we lean towards only supporting later versions of dependencies.  would it be too aggressive to expect paparazzi users to be on 1.2.0+?

jrodbx (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
saket: I… don't have a good answer. Will requiring 1.2.0+ and calling it out in release notes be sufficient?

FWIW `1.2.0` was released in June, 2022 whereas `1.1.0` was released in Feb, 2022.

jrodbx: > Will requiring 1.2.0+ and calling it out in release notes be sufficient?

I think so.  :shipit:

saket: 🚀

joshfriend: Seems fine to me, this version is already bumping the minimum kotlin version and requiring AGP 7.4.0. Can't wait forever for consumers to catch up and definitely don't want to hold anyone back from using savedstate 1.2.0 either.

PR #687, opened 2023-01-19
Add plugin test results to be included in GHA artifacts

jrodbx:
JVM module test reports are in **/build/reports/tests/test/
But, plugin test reports are in **/build/reports/tests/testDebugUnitTest/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           name: test-failures
           path: |
-            **/build/reports/tests/test/
+            **/build/reports/tests/*/
             **/out/failures/
             paparazzi/paparazzi-gradle-plugin/src/test/projects/**/build/reports/paparazzi/images/
 


.github/workflows/build.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         with:
           name: test-failures
           path: |
-            **/build/reports/tests/test/
+            **/build/reports/tests/*/
             **/out/failures/
```
TWiStErRob, line 51: We'll need the reports from individual tests too, they're Gradle projects with Android tests ezecuted, does this ** include those?

.github/workflows/build.yml
```diff
@@ -47,7 +47,7 @@ jobs:
...
             paparazzi/paparazzi-gradle-plugin/src/test/projects/**/build/reports/paparazzi/images/
```
TWiStErRob, line 52: We'll need the reports from individual tests too, they're Gradle projects with Android tests ezecuted, does this ** you modified above include those?

JakeWharton (approved):

.github/workflows/build.yml
jrodbx: yes

PR #690, opened 2023-01-19
Make sure record/verify/properties actually affect test tasks

### LINKED ISSUE #747: verify does not take in to account resource changes
**Description**
When a Composable is using a string resource the verify tasks does not take into account changes that could have happened to that resources.

**Steps to Reproduce**
```diff
Index: sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt b/sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt
--- a/sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt	(revision 9e4684a5a180a52cef543efbe0a6d26af3aaea1e)
+++ b/sample/src/test/java/app/cash/paparazzi/sample/HelloComposeTest.kt	(date 1680258820694)
@@ -8,6 +8,7 @@
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
@@ -37,7 +38,7 @@
       .wrapContentSize()
   ) {
     Text(text)
-    Text(text, style = TextStyle(fontFamily = FontFamily.Cursive))
+    Text(stringResource(id = R.string.test), style = TextStyle(fontFamily = FontFamily.Cursive))
     Text(
       text = text,
       style = TextStyle(textDecoration = TextDecoration.LineThrough)
Index: sample/src/main/res/values/strings.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/sample/src/main/res/values/strings.xml b/sample/src/main/res/values/strings.xml
new file mode 100644
--- /dev/null	(date 1680258846345)
+++ b/sample/src/main/res/values/strings.xml	(date 1680258846345)
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="test">Hello World</string>
+</resources>
```

Run `./gradlew recordPaparazziDebug`
A screenshot is generated

app cash paparazzi sample_HelloComposeTest_compose

Change the string resource value

```diff
Index: sample/src/main/res/values/strings.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/sample/src/main/res/values/strings.xml b/sample/src/main/res/values/strings.xml
--- a/sample/src/main/res/values/strings.xml	(revision 3ee2d41e6c3889c9cd4e1b8964735f484f8fa956)
+++ b/sample/src/main/res/values/strings.xml	(date 1680259023916)
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <string name="test">Hello World</string>
+  <string name="test">Hello Paparazzi</string>
 </resources>
```

run `./gradlew verifyPaparazziDebug`

The build succeeds without any issue.

> BUILD SUCCESSFUL in 9s
36 actionable tasks: 1 executed, 3 from cache, 32 up-to-date

**Expected behavior**
Build should fail.

**Additional information:**
- Paparazzi Version: 1.2.0
- OS: MacOS Ventura 13.1
- Compile SDK: (Sample project) 33
- Gradle Version: (Sample project) 7.6
- Android Gradle Plugin Version: (Sample project) 7.4.0

**Screenshots**
n/a

Thanks in advance

TWiStErRob:
See https://github.com/cashapp/paparazzi/pull/206#issuecomment-1397011673

Fixes #747


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -163,6 +163,14 @@ class PaparazziPlugin : Plugin<Project> {
         test.outputs.dir(snapshotOutputDir)
 
         val paparazziProperties = project.properties.filterKeys { it.startsWith("app.cash.paparazzi") }
+        test.systemProperties.putAll(paparazziProperties)
+
+        // Explicitly register these as inputs so that they are considered when determining up-to-date.
+        // They have to be added later to test.systemProperties,
+        // because at this point in the configuration, the properties are not resolvable yet.
+        // They become resolvable after the last afterEvaluate runs.
```
jrodbx, line 171:
```suggestion
        // Explicitly register these as inputs so that they are considered when determining up-to-date.
        // The properties become resolvable after the last afterEvaluate runs.
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -163,6 +163,14 @@ class PaparazziPlugin : Plugin<Project> {
         test.outputs.dir(snapshotOutputDir)
 
         val paparazziProperties = project.properties.filterKeys { it.startsWith("app.cash.paparazzi") }
+        test.systemProperties.putAll(paparazziProperties)
```
jrodbx, line 166: why did this have to move?  looks irrelevant to the change?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -163,6 +163,14 @@ class PaparazziPlugin : Plugin<Project> {
...
+
+        // Explicitly register these as inputs so that they are considered when determining up-to-date.
+        // They have to be added later to test.systemProperties,
+        // because at this point in the configuration, the properties are not resolvable yet.
+        // They become resolvable after the last afterEvaluate runs.
+        test.inputs.property("paparazzi.test.record", isRecordRun)
+        test.inputs.property("paparazzi.test.verify", isVerifyRun)
```
jrodbx, line 173: note to future us: should probably move this to an enum type, since these values are mutually exclusive to each other.

jrodbx (approved):
thanks!

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: Because it's irrelevant to the doFirst hack (didn't look at history but possible that it was added here because doFirst was already setting properties). Now that I had full understanding I could confidently move it outside the doFirst block. It has the same problem as the other two, if it's set AFTER up to date check, it breaks Gradle. see https://github.com/cashapp/paparazzi/pull/690/commits/8ab5ee9f38b288bbdb41324d08b913cf7639fd9b
I can't prove this, because I don't know what those properties can be, but I'm almost certain they would come from `-Papp.cash.paparazzi.foo` if you point me to how/where/why this is used, I can add another test proving it was broken.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: Yeah, I had a similar thought, however mind you that the calculation of it could be very complicated because of the laziness needed in "when graph ready". Anyway, it might be doable.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Sorry, still not following.  Rewording my question: does moving this line have any effect on fixing the "make sure record/verify actually affect test tasks" original premise that linked to the comment?   it doesn't appear so, so should probably be undone?

> because I don't know what those properties can be

I think the only example at the moment is the property that enables turning off verbose logging?  https://github.com/cashapp/paparazzi/blob/master/paparazzi/paparazzi-gradle-plugin/src/test/projects/flag-debug-linked-objects-on/gradle.properties#L2

And that wouldn't/shouldn't affect test task cacheability?

TWiStErRob: Yes, the original issue @alashow reported was that record/verify was not re-executing if test was executed before. Actually he only reported record, I read the code and added verify because I saw it was broken the same way. I also added properties to the scope because it was broken the same way.

Everything that modifies a tasks behaviour must cause the task to re-execute. If it doesn't, it's going to cause incorrect builds and/or confuse users, because that's just how Gradle tasks are expected to work. This is especially true for something that changes logging, because imagine I run:
```
gradlew recordPaparazziDebug
```
and then I realize I want to see the logs for it, so I run:
```
gradlew recordPaparazziDebug -Papp.cash.paparazzi.debug.linked.objects=true
```
and I get `> :testDebugUnitTest UP-TO-DATE`, that's a big wtf moment. Because I changed a build parameter! I won't change dependencies, because I want to see the current version's logs. I won't change my code, because I want to see the current code's behaviour. I won't change the tests, because I want to see my current test behaviour. I **will** add the property, because the change is that before I didn't want to see logs, and now I want to see them.

Add on top of this if the user hopefully has caching enabled and because the task executed unexpectedly UP-TO-DATE, now the next reasonable thing is to do:
```
gradlew clean recordPaparazziDebug -Papp.cash.paparazzi.debug.linked.objects=true
```
(which one should never do, btw.)
...and now I get `> :testDebugUnitTest FROM-CACHE`, there's just no winning trying to execute my tests with logs. User has to hack around with even more advanced Gradle options or change their code for no reason to get their logs from Paparazzi.

With this PR neither of this will happen it will "just work", this PR essentially changes the cache key, which is used for both build cache and up to date checks. If the user removed the flag, the old cache key will be used and tests will execute from cache.

So, yes, even for a logging altering behaviour, cache key should change and affected tasks re-execute.

Note that enabling logs will affect the console output, but will also affect the generated report (Standard Output) tab. So it seriously changes the `@OutputDirectory` contents, it is therefore an `@Input` to the task. `systemProperties` is automatically an input, so this works.

Note that changing a value in `gradle.properties` and `-P` and `ORG_GRADLE_PROJECT_prop` env var are all almost equivalent. They all show up in the same way in `Project.properties`, just with different precedence and override order.

TWiStErRob:
Related Gradle feature: https://github.com/gradle/gradle/issues/9092

TWiStErRob:
Is there anything missing here @jrodbx?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Sorry for the delay in getting back to this!

Re: how Gradle works, I understand the caching bit, but I think the part that I was missing (and now understand) is this: 

> With this PR neither of this will happen it will "just work", this PR essentially changes the cache key, which is used for both build cache and up to date checks. If the user removed the flag, the old cache key will be used and tests will execute from cache.

TL;DR: setting the system properties in the task action is too late in terms of having those properties considered in calculation of the cache key!  and I agree, changing the logging property should result in showing/hiding the logs which would require a re-run of the task.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -163,6 +163,12 @@ class PaparazziPlugin : Plugin<Project> {
         test.outputs.dir(snapshotOutputDir)
 
         val paparazziProperties = project.properties.filterKeys { it.startsWith("app.cash.paparazzi") }
+        test.systemProperties.putAll(paparazziProperties)
+
+        // Explicitly register these as inputs so that they are considered when determining up-to-date.
+        // The properties become resolvable after the last afterEvaluate runs.
+        test.inputs.property("paparazzi.test.record", isRecordRun)
+        test.inputs.property("paparazzi.test.verify", isVerifyRun)
```
jrodbx, line 171: style nit: can we move this under the other "inputs" (just after "paparazzi.nativePlatform")?  also, let's remove the comment, since this is standard Gradle knowledge.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -163,6 +163,12 @@ class PaparazziPlugin : Plugin<Project> {
         test.outputs.dir(snapshotOutputDir)
 
         val paparazziProperties = project.properties.filterKeys { it.startsWith("app.cash.paparazzi") }
+        test.systemProperties.putAll(paparazziProperties)
```
jrodbx, line 166: similar style nit: can we group/move these two lines above the other "systemProperties" (just above "paparazzi.test.resources")?

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -476,6 +476,60 @@ class PaparazziPluginTest {
...
+      .runFixture(fixtureRoot) { build() }
+
+    with(secondRunResult.task(":testDebugUnitTest")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isEqualTo(SUCCESS) // not UP-TO-DATE
+    }
+  }
+
+  @Test
+  fun rerunTestsOnVerifyPropertyChange() {
```
jrodbx, line 507: can we group this into one test `rerunTestsOnPropertyChange` with one project `rerun-property-change` merging the 3 takes, i.e., testDebugUnitTest, recordPaparazziDebug, verifyPaparazziDebug?

jrodbx (approved):
Approved again with updated comments.  Also, please rebase and squash.  Thanks for fixing this!

TWiStErRob:
Did everything so far, squashing now.

TWiStErRob:
Addressed all the comments in separate commits, when merging select squash merge please.

jrodbx (approved):

PR #691, opened 2023-01-19
Add .gitattributes to make spotlessApply work on Windows

TWiStErRob:
Original behavior: `gradlew spotlessApply` -> all files changed

Steps:
 * look up why -> https://github.com/diffplug/spotless/issues/39
 * add .gitattributes
 * reset working copy -> https://stackoverflow.com/q/48642692/253468

New behavior: If I run spotless again, it changes nothing, unless there's actually something wrong.

Please someone on Mac/linux check out the branch test this (on an existing and a new clone):
 * change a file in `paparazzi/paparazzi-gradle-plugin/src/main/kotlin`
 * `git status` (observe changed file)
 * run `./gradlew -p paparazzi spotlessApply`
 * `git status` (observe no changed files)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 * text eol=lf
 gradlew.bat text eol=crlf
+*.png binary
+*.jar binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,17 +1,5 @@
-# Source code
-*.kt            text eol=lf
-*.java          text eol=lf
-*.xml           text eol=lf
+* text=auto eol=lf
 
-# Gradle files
-*.gradle        text eol=lf
-*.properties    text eol=lf
-*.bat           text eol=crlf
-
-# GitHub stuff
-*.md            text eol=lf
-*.yml           text eol=lf
-
-# Binaries
-*.png           binary
-*.jar           binary
+*.bat text eol=crlf
+*.jar binary
+*.png binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-* text eol=lf
-gradlew.bat text eol=crlf
 *.png binary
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,17 @@
+# Source code
+*.kt            text eol=lf
```
JakeWharton, line 2: Make the default LF and only add exemptions. Like https://github.com/square/retrofit/blob/master/.gitattributes. Less whack-a-mole.

.gitattributes
```diff
@@ -0,0 +1,17 @@
...
+*.java          text eol=lf
+*.xml           text eol=lf
+
+# Gradle files
+*.gradle        text eol=lf
+*.properties    text eol=lf
+*.bat           text eol=crlf
```
TWiStErRob, line 9: Something is fishy still, even after resetting gradlew.bat keeps changing, any ideas welcome.

I tried autocrlf=true and false, and then reset, but no luck.

build.gradle
```diff
@@ -68,7 +68,3 @@ subprojects {
 tasks.register("clean", Delete).configure {
   delete rootProject.buildDir
 }
-
-tasks.named("wrapper").configure {
-  distributionType = Wrapper.DistributionType.ALL
```
JakeWharton, line 73: Big fan. But should be in a separate PR.

TWiStErRob: k

.gitattributes
TWiStErRob: Re-cloned and all good, but I was able to reproduce it with the code here: https://stackoverflow.com/q/48642692/253468

This however fixes it: https://stackoverflow.com/a/19588910/253468

Right now it's above my git abilities, but it should be fine considering retrofit uses the same file.

jrodbx (approved):

.gitattributes
TWiStErRob: It was solved in: https://github.com/cashapp/paparazzi/pull/721

PR #695, opened 2023-01-20
Repeatable test fixes

TWiStErRob:
See individual commit messages for more details (rationale, fixes, reasons).

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -117,6 +117,7 @@ class PaparazziPluginTest {
   @Test
   fun customBuildDir() {
     val fixtureRoot = File("src/test/projects/custom-build-dir")
+    fixtureRoot.resolve("custom").deleteRecursively()
 
     val result = gradleRunner
       .withArguments("testDebug", "--stacktrace")
@@ -130,8 +131,6 @@ class PaparazziPluginTest {
 
     val snapshotsDir = File(fixtureRoot, "custom/reports/paparazzi/images")
     assertThat(snapshotsDir.exists()).isTrue()
-
-    fixtureRoot.resolve("custom").deleteRecursively()
   }
 
   @Test
@@ -223,6 +222,8 @@ class PaparazziPluginTest {
   @Test
   fun record() {
     val fixtureRoot = File("src/test/projects/record-mode")
+    val snapshotsDir = File(fixtureRoot, "src/test/snapshots")
+    snapshotsDir.deleteRecursively()
 
     val result = gradleRunner
       .withArguments("recordPaparazziDebug", "--stacktrace")
@@ -230,20 +231,17 @@ class PaparazziPluginTest {
 
     assertThat(result.task(":testDebugUnitTest")).isNotNull()
 
-    val snapshotsDir = File(fixtureRoot, "src/test/snapshots")
-
     val snapshot = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_RecordTest_record.png")
     assertThat(snapshot.exists()).isTrue()
 
     val snapshotWithLabel = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_RecordTest_record_label.png")
     assertThat(snapshotWithLabel.exists()).isTrue()
-
-    snapshotsDir.deleteRecursively()
   }
 
   @Test
   fun recordAllVariants() {
     val fixtureRoot = File("src/test/projects/record-mode")
+    fixtureRoot.resolve("src/test/snapshots").deleteRecursively()
 
     val result = gradleRunner
       .withArguments("recordPaparazzi", "--stacktrace")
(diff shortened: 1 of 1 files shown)


gamepro65 (approved):
Nice! No more manually deleting build folders when these tests fail 🥂

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -117,6 +117,7 @@ class PaparazziPluginTest {
   @Test
   fun customBuildDir() {
     val fixtureRoot = File("src/test/projects/custom-build-dir")
+    fixtureRoot.resolve("custom").deleteRecursively()
```
jrodbx, line 120: will the delete fail if an exception is thrown and the custom directory doesn't yet exist?

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -151,6 +150,7 @@ class PaparazziPluginTest {
   @Test
   fun flagDebugLinkedObjectsIsOff() {
     val fixtureRoot = File("src/test/projects/flag-debug-linked-objects-off")
+    fixtureRoot.resolve("build").deleteRecursively()
```
jrodbx, line 153: similar, what if build doesn't exist yet?

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -162,6 +162,7 @@ class PaparazziPluginTest {
   @Test
   fun flagDebugLinkedObjectsIsOn() {
     val fixtureRoot = File("src/test/projects/flag-debug-linked-objects-on")
+    fixtureRoot.resolve("build").deleteRecursively()
```
jrodbx, line 165: similar, what if build doesn't exist yet?

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -174,6 +175,9 @@ class PaparazziPluginTest {
   fun cacheable() {
     val fixtureRoot = File("src/test/projects/cacheable")
 
+    fixtureRoot.resolve("build").deleteRecursively()
+    fixtureRoot.resolve("build-cache").deleteRecursively()
```
jrodbx, line 179: similar, what if build and build-cache doesn't exist yet?

jrodbx (commented):
I like the idea, as I too have been annoyed by this, but I think this trades one issue for another.

I conservatively called delete where the directory was ensured to exist, but early test failures don't cleanup properly.  Here, if a test fails before the directory exists, the cleanup fails and distracts from the original test failure. 

A better way might be to use TemporaryDirectory rules + custom directory paths in the tests, to ensure paths exist before deleting?

TWiStErRob:
The answer is the same for all, `deleteRecursively` returns a boolean. It is idempotent, if it was deleted, nothing happens, and moves on. This why it "just works" on CI right now in this state of the PR. 

It could be made more clever, e.g. `assertTrue(!dir.exists() || dir.deleteRecursively())` but I didn't want to deviate too much and introduce new constructs. Maybe I aimed too low.

My aim was to get a consistent pass when executing from more states than right now (more stable tests in dev env). If you run the tests twice with a whitespace change between the two, some tests just fail. This makes it really hard to change anything, because we cannot/don't trust the tests. Or as @gamepro65 said, it's annoying to work with (against) them.

So this is a quick fix for this specific case as a first step in a line of future improvements. I agree about somehow always starting from scratch in a temp dir therefore ensuring a clean slate. I use that approach in my project with a ton of tests in a huge matrix and it has served me well. Wanted to keep changes to a minimum here. I might propose another change after to start a discussion how to improve the experience.

> Here, if a test fails before the directory exists, the cleanup fails and distracts from the original test failure.

That won't happen if you pre-clean: if the test fails before preparatory cleanup, then cleanup won't happen. There's no try-finally. And the state that's left from previous execution (if any) will be cleaned up by the next by ensuring the cleanup is called asap (before doing anything really).

This distractive cleanup failure could happen with a test rule, if it too eagerly cleans up. I think the whole notion of preparatory cleanup in tests is wrong, because it implies shared state. If a temp dir is used, the only reason to (best effort) delete the resulting files is to save space on a shared machine. And on failures it's worth not doing cleanup so diagnosis is possible.

> I think this trades one issue for another.

Based on this I don't see the new issue we are trading into. Maybe the only thing is that we could ensure in this PR additionally is that the cleanup was successful, so the precondition of empty folders is actually and always there. If the existing folder cannot be deleted that's an error, because the OS/process is locking some files, which is a problem in our code or usage of frameworks. The more strict the better, unless it's flaky.

TWiStErRob:
Is there anything else need to be done here? @jrodbx

PR #673, opened 2023-01-05
Register a default OnBackPressedDispatcherOwner if its present in classpath

saket:
Follow-up of https://github.com/cashapp/paparazzi/pull/668.

PR #696, opened 2023-01-20
Add tests to verify installation of LifecycleOwner and SavedStateRegistryOwner

saket:
Follow-up of https://github.com/cashapp/paparazzi/pull/673#discussion_r1061993765

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -600,7 +600,7 @@ class PaparazziPluginTest {
 
     val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/images")
     val snapshots = snapshotsDir.listFiles()
-    assertThat(snapshots!!).hasLength(1)
+    assertThat(snapshots!!).hasLength(3)
   }
 
   @Test
--- a/paparazzi/paparazzi-gradle-plugin/src/test/projects/lifecycle-usages/build.gradle
+++ b/paparazzi/paparazzi-gradle-plugin/src/test/projects/lifecycle-usages/build.gradle
@@ -22,4 +22,5 @@ android {
 
 dependencies {
   testImplementation 'androidx.activity:activity-compose:1.6.1'
+  testImplementation libs.truth
 }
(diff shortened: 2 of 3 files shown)


paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -593,14 +593,14 @@ class PaparazziPluginTest {
...
+    snapshotsRunsDir.deleteRecursively()
 
     gradleRunner
       .withArguments("testDebug", "--stacktrace")
       .runFixture(fixtureRoot) { build() }
 
-    val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/images")
-    val snapshots = snapshotsDir.listFiles()
-    assertThat(snapshots!!).hasLength(1)
+    assertThat(snapshotsRunsDir.listFiles()!!).hasLength(3)
```
saket, line 603: Thoughts on this? I wasn't able to assert the presence of 3 images because the 3 tests generate identical screenshots resulting in a single image saved on disk.

jrodbx: ah right, this is why i've kept a loose convention of 1 test per test project.  remember setting different background colors to force 3 images?  while looking into the images dir is peeking into implementation details, peeking into the runs dir is kinda moreso, so it'd be nice to avoid that.

paparazzi/paparazzi-gradle-plugin/src/test/projects/lifecycle-usages/build.gradle
```diff
@@ -22,4 +22,5 @@ android {
 
 dependencies {
   testImplementation 'androidx.activity:activity-compose:1.6.1'
+  testImplementation libs.truth
```
jrodbx, line 25: can the assertions part of these tests be moved out of the plugin integration tests and into Paparazzi tests?

oh wait nvm, probably not, cuz androidx savedstate and friends are aars, and not jars....drat.

jrodbx (approved):

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
saket: I can do that

saket: done!

PR #710, opened 2023-01-30
fix NPE when unit test variant is disabled

damianw:
When used in a project with configuration like the following:
```
androidComponents {
  beforeVariants(selector().all()) {
    enableUnitTest = buildType == "debug"
  }
}
```

The Paparazzi Gradle plugin will throw an NPE:

```
java.lang.NullPointerException: Cannot invoke "com.android.build.gradle.api.UnitTestVariant.getName()" because the return value of "com.android.build.gradle.api.LibraryVariant.getUnitTestVariant()" is null
	at app.cash.paparazzi.gradle.PaparazziPlugin.setupPaparazzi$lambda-29(PaparazziPlugin.kt:111)
```

This skips such variants with unit tests disabled and adds a test for it.

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -74,6 +74,8 @@ class PaparazziPlugin : Plugin<Project> {
       .libraryVariants
     variants.all { variant ->
       val variantSlug = variant.name.capitalize(Locale.US)
+      val testVariantSlug = variant.unitTestVariant?.name?.capitalize(Locale.US)
+        ?: return@all
 
       val mergeResourcesOutputDir = variant.mergeResourcesProvider.flatMap { it.outputDir }
       val mergeAssetsProvider =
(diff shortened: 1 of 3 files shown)


dniHze:
This PR should resolve #711

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -74,6 +74,8 @@ class PaparazziPlugin : Plugin<Project> {
       .libraryVariants
     variants.all { variant ->
       val variantSlug = variant.name.capitalize(Locale.US)
+      val testVariantSlug = variant.unitTestVariant?.name?.capitalize(Locale.US)
+        ?: return@all
```
jrodbx, line 78:
```suggestion
      val testVariant = variant.unitTestVariant ?: return@all
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -108,8 +110,6 @@ class PaparazziPlugin : Plugin<Project> {
         task.paparazziResources.set(project.layout.buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
       }
 
-      val testVariantSlug = variant.unitTestVariant.name.capitalize(Locale.US)
```
jrodbx, line 111: val testVariantSlug = testVariant.name.capitalize(Locale.US)

jrodbx (approved):

jrodbx:
Thanks!

PR #716, opened 2023-02-15
update publishing library

gabrielittner:
Wit this the plugin will explicitly create a staging repository before publishing and close it and the end which means that the `--no-parallel` isn't needed anymore and publishing is a lot more reliant. The `SONATYPE_AUTOMATIC_RELEASE=true` means that the repository will also be released automatically at the end of the build, so you'll have 0 interactions with the Sonatype Web UI.

I needed to adjust the code that injects the plugin version into the code because the publish plugin does not send `project.version` anymore.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -69,6 +69,6 @@ plugin-android = { module = "com.android.tools.build:gradle", version.ref = "agp
 plugin-dokka = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version.ref = "kotlin" }
 plugin-grgit = { module = "org.ajoberstar.grgit:grgit-gradle", version = "5.0.0" }
 plugin-kotlin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
-plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.21.0" }
+plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.24.0" }
 plugin-spotless = { module = "com.diffplug.spotless:spotless-plugin-gradle", version = "6.12.0" }
 plugin-versions = { module = "com.github.ben-manes:gradle-versions-plugin", version = "0.44.0" }
--- a/paparazzi/gradle.properties
+++ b/paparazzi/gradle.properties
@@ -16,6 +16,7 @@ POM_DEVELOPER_URL=https://github.com/cashapp/
 
 SONATYPE_HOST=DEFAULT
 RELEASE_SIGNING_ENABLED=true
+SONATYPE_AUTOMATIC_RELEASE=true
 
 org.gradle.caching=true
 org.gradle.parallel=true
--- a/paparazzi/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi/paparazzi-gradle-plugin/build.gradle
@@ -31,7 +31,7 @@ sourceSets {
 def generateVersion = tasks.register("pluginVersion") {
   def outputDir = file('src/generated/kotlin')
 
-  inputs.property 'version', version
+  inputs.property 'version', project.property("VERSION_NAME")
   inputs.property 'nativeLibVersion', libs.versions.layoutlib
   outputs.dir outputDir
 
@@ -40,7 +40,7 @@ def generateVersion = tasks.register("pluginVersion") {
     versionFile.parentFile.mkdirs()
     versionFile.text = """// Generated file. Do not edit!
 package app.cash.paparazzi
-const val VERSION = "${project.version}"
+const val VERSION = "${project.property("VERSION_NAME")}"
 const val NATIVE_LIB_VERSION = "${libs.versions.layoutlib.get()}"
 """
   }


paparazzi/paparazzi-gradle-plugin/build.gradle
```diff
@@ -31,7 +31,7 @@ sourceSets {
 def generateVersion = tasks.register("pluginVersion") {
   def outputDir = file('src/generated/kotlin')
 
-  inputs.property 'version', version
+  inputs.property 'version', project.property("VERSION_NAME")
```
TWiStErRob, line 34: This looks fishy. This could mean that the POM will have missing metadata, if the `project.version` is not set up correctly. I would keep these as `version`, and set `project.version = project.property("VERSION_NAME")` at the top somewhere.

gabrielittner: The property is hardcoded in the committed gradle.properties file and `project.property` will throw an exception if the property does not exist. The config here will not in any way determine what ends up in the pom, this is handled by the publish plugin which does this based on the same property and has been using this property for that purpose since the beginning of its existence

TWiStErRob: > I needed to adjust the code that injects the plugin version into the code because the publish plugin does not send project.version anymore.

so, earlier project.version was set, which was propagated to many places, now you're saying it's not done. `project.version` is the central place to define the version of the module, its artifacts, metadata, components, publications etc.

This `inputs.property` is only used for doing up-to-date checks (and cache key generation), it won't actually be an input. The one lower down will generate the Kotlin code so that **at runtime** consumers can see which version is there.

I did a quick diff and the version is actually correctly present on the branch, but many other things are wrong, will report them separately.

gradle/libs.versions.toml
```diff
@@ -69,6 +69,6 @@ plugin-android = { module = "com.android.tools.build:gradle", version.ref = "agp
 plugin-dokka = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version.ref = "kotlin" }
 plugin-grgit = { module = "org.ajoberstar.grgit:grgit-gradle", version = "5.0.0" }
 plugin-kotlin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
-plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.21.0" }
+plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.24.0" }
```
TWiStErRob, line 72: The `paparazzi-javadoc.jar` is empty.

TWiStErRob, line 72: The `paparazzi-sources.jar` contents are relocated into `main/` and only after that the package structure starts: `main/app/cash`. This is unusual I think.

`paparazzi-gradle.plugin-sources.jar` is starts with `app/cash` on both main and this PR.

TWiStErRob: The additional index.html and paparazzi.js files are now missing from the `-sources.jar`.

TWiStErRob: Same applies to `paparazzi-agent-sources.jar`

TWiStErRob (commented):
I tested the output with
```
gradlew :paparazzi:publishToMavenLocal :paparazzi-gradle-plugin:publishToMavenLocal :paparazzi-agent:publishToMavenLocal
```
and found some interesting differences in `~/.m2/app/cash`.

paparazzi/paparazzi-gradle-plugin/build.gradle
gabrielittner: >  so, earlier project.version was set, which was propagated to many places, now you're saying it's not done. 

Yes.

> project.version is the central place to define the version of the module, its artifacts, metadata, components, publications etc.

It's the default used for publications but the publish plugin is giving higher priority to the `VERSION_NAME` property if that is set and that is how paparazzi was already setting the version before.

gradle/libs.versions.toml
gabrielittner: >The paparazzi-sources.jar contents are relocated into main/ and only after that the package structure starts: main/app/cash. This is unusual I think.
>paparazzi-gradle.plugin-sources.jar is starts with app/cash on both main and this PR.

This is how the Kotlin plugin creates the sources jar (previously we had a custom task for creating sources in the publish plugin), I guess they share the `kotlinSourcesJar` task with multiplatform where you need to differentiate between source sets because otherwise you might have the same file twice. IntelliJ/Studio seem to be fine in handling this. The reason why the Gradle plugin is different is that Gradle is creating the sources jar there.

> The additional index.html and paparazzi.js files are now missing from the -sources.jar.

Also seems to come from now using `kotlinSourcesJar`. Personally I don't think this isn't a problem because resources are included in the main jar already but if it is I can see if we can workaround this.

gabrielittner: This is a bug in the publish plugin will fix that over there.

gabrielittner: Will be resolved by https://github.com/vanniktech/gradle-maven-publish-plugin/pull/559, will do a 0.25.2 release after this is merged

gabrielittner: [92d0956](https://github.com/cashapp/paparazzi/pull/716/commits/92d09562056786a7ba1dd71465c9f13b450a2e4d)

paparazzi/paparazzi-gradle-plugin/build.gradle
TWiStErRob: I see, so now the Paparazzi code-gen is matching the publishing plugin's behavior, which is different than Gradle's, but at least it's consistent within this build.

paparazzi/paparazzi-gradle-plugin/build.gradle
```diff
@@ -40,7 +40,7 @@ def generateVersion = tasks.register("pluginVersion") {
     versionFile.parentFile.mkdirs()
     versionFile.text = """// Generated file. Do not edit!
 package app.cash.paparazzi
-const val VERSION = "${project.version}"
+const val VERSION = "${project.property("VERSION_NAME")}"
```
jrodbx, line 43: mind moving this instead to the paparazzi/build.gradle line 23 as:

```
version = property("VERSION_NAME") as String
```

keeps things closer to how they were, and prevents possible new modules from needing to remember this (if they need it).

paparazzi/gradle.properties
```diff
@@ -16,6 +16,7 @@ POM_DEVELOPER_URL=https://github.com/cashapp/
 
 SONATYPE_HOST=DEFAULT
 RELEASE_SIGNING_ENABLED=true
+SONATYPE_AUTOMATIC_RELEASE=true
```
jrodbx, line 19: is this only on snapshot builds or release builds too?  

i eventually want this, when a new git tag is pushed (similar to how some other OSS repos do it) but not sure if this repo is ready for this yet?  @JakeWharton thoughts?

jrodbx (approved):

paparazzi/paparazzi-gradle-plugin/build.gradle
gabrielittner: Done a2b31a4

paparazzi/gradle.properties
gabrielittner: This is for release builds, snapshots are automatically available after publishing either way. I can set it to false for now if you want to but from what I can see everything you should need for it is in place (I've also used this to publish a fork from master for us a few times, although that was using a different github actions workflow).

jrodbx: let's see how it goes!

Issue #717, opened 2023-02-17
layoutlib module clones "the world"


TWiStErRob:
As a contributor I wanted to quickly test something in relation to a PR in this repo.
I ran `gradlew publishToMavenLocal` and that started `cloneLayoutLib`, which took 1.5 hours, only to end with:
```
* What went wrong:
Execution failed for task ':libs:layoutlib:cloneLayoutlib'.
> There is not enough space on the disk
```

I double-checked and the `paparazzi\build\prebuilts\studio\layoutlib\.git` folder is 10G. Is it really necessary to pull everything from the server to only get the 300M of `paparazzi\build\prebuilts\studio\layoutlib\data` we need?

_After writing this report I realized that I should've been more specific about which module I want to pubish..._ the issue still stands, we could achieve a 97% bandwidth and disk space saving. 😅

### Investigation

Out of curiosity how hard this is, I did the following:
```shell
$ git clone https://android.googlesource.com/platform/prebuilts/studio/layoutlib . --depth=1 --single-branch
Cloning into '.'...
remote: Finding sources: 100% (2/2)
remote: Total 2 (delta 0), reused 2 (delta 0)
Receiving objects: 100% (2/2), done.

$ git status
On branch master
Your branch is up to date with 'origin/master'.

$ git fetch origin f5f9f71
fatal: couldn't find remote ref f5f9f71

$ git fetch origin studio-2022.1.1-beta4 --depth 1
remote: Finding sources: 100% (10480/10480)
Receiving objects: 100% (10480/10480), 181.67 MiB | 348.00 KiB/s, done.

Resolving deltas: 100% (2160/2160), done.
From https://android.googlesource.com/platform/prebuilts/studio/layoutlib
 * tag               studio-2022.1.1-beta4 -> FETCH_HEAD
```

That said, this (ab)uses the fact that the repo's `master` just happens to be empty, I tried again and this one command does it all:
```shell
$ git clone https://android.googlesource.com/platform/prebuilts/studio/layoutlib --branch studio-2022.1.1-beta4 --single-branch --depth 1 .
Cloning into '.'...
remote: Finding sources: 100% (10484/10484)
Receiving objects: 100% (10484/10484), 181.67 MiB | 344.00 KiB/s, done.

Resolving deltas: 100% (2160/2160), done.
Note: switching to 'f5f9f71ebf7658a26af18684242501e987968ddb'.

Updating files: 100% (10647/10647), done.
```

[...]

PR #718, opened 2023-02-17
Rewire Grgit usage

TWiStErRob:
1. `apply plugin: 'org.ajoberstar.grgit'` is not necessary, because it exists for a performance improvement where it eagerly opens **the current folder** as a Git repo, but we need to clone a new folder, so that "improvement" is actually not necessary and locks the .git folder in the Gradle daemon.
2. `grgit` is an instance from the project extension, but we need just need `org.ajoberstar.grgit.Grgit` which is on the classpath already even without applying the plugin.
3. `grgit.clone()` is a static method call on an instance of Grgit.
4. `grgit` instance created by `open`/`clone` is never closed, so therefore the `paparazzi\build\prebuilts\studio\layoutlib\.git` folder is locked inside the Gradle daemon (e.g. cannot delete without `gradlew --stop`)
5. There's an operation missing when an existing repo is `open`ed, the hash might not exist yet, so a `fetch()` is necessary to get the new updates. When the repo was just cloned this is essentially a no-op.

### DIFF
--- a/paparazzi/libs/layoutlib/build.gradle
+++ b/paparazzi/libs/layoutlib/build.gradle
@@ -33,6 +33,7 @@ tasks.register('cloneLayoutlib') {
       }
     }
     grgit.withCloseable { repo ->
+      repo.fetch()
       logger.warn("Checking out SHA ${libs.versions.layoutlibPrebuiltSha.get()}")
       repo.checkout {
         branch = libs.versions.layoutlibPrebuiltSha.get()
--- a/paparazzi/libs/layoutlib/build.gradle
+++ b/paparazzi/libs/layoutlib/build.gradle
@@ -1,4 +1,5 @@
-apply plugin: 'org.ajoberstar.grgit'
+import org.ajoberstar.grgit.Grgit
+
 apply plugin: 'com.vanniktech.maven.publish'
 
 version = libs.versions.layoutlib.get()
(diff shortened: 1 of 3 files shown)


AfzalivE:
I recently added the support for shallow clone in [grgit 5.2.0](https://github.com/ajoberstar/grgit/releases) (using the `depth` argument). Will specifying this also help speed up anything?

TWiStErRob:
Awesome @AfzalivE! Thanks for that! Replied in the performance issue (#717), as this PR is just a side effect of that investigation in #717.

paparazzi/libs/layoutlib/build.gradle
```diff
@@ -17,22 +18,26 @@ tasks.register('cloneLayoutlib') {
     if (repoDir.exists() && !repoDir.list()) {
       repoDir.delete()
     }
+    Grgit grgit
     if (!repoDir.exists()) {
       logger.warn('Cloning prebuilt layoutlib: this make take a few minutes...')
```
jrodbx, line 23:
```suggestion
      logger.warn('Cloning prebuilt layoutlib: this may take a few minutes...')
```

unrelated, but apparently future me catches past me typos 😄

paparazzi/libs/layoutlib/build.gradle
```diff
@@ -1,4 +1,5 @@
-apply plugin: 'org.ajoberstar.grgit'
+import org.ajoberstar.grgit.Grgit
```
jrodbx, line 1: should we also remove the plugin and put `grgit-core` on the classpath instead? 
 https://search.maven.org/search?q=a:grgit-core

jrodbx (approved):
finally got around to this.  lgtm!

TWiStErRob:
Force push is a clean rebase

paparazzi/libs/layoutlib/build.gradle
TWiStErRob: Makes sense, looks a bit strange in code, but did it anyway, because the grgit-gradle classes are not used. c2cacaed

TWiStErRob:
For testing after rebase I only got as far as
```
> Task :libs:layoutlib:cloneLayoutlib
Cloning prebuilt layoutlib: this may take a few minutes...
```
I don't have hours to wait for gigabytes to download at the moment.

TWiStErRob:
While developing #884 I only changed significantly change the inside of the clone/fetch/checkout, so in essence I tested this PR as well.

Issue #560, opened 2022-09-01
The inputs of PrepareResourcesTask are misconfigured


cesar1000:
**Description**
`mergeResourcesOutput`, `mergeAssetsOutput` and `platformDataRoot` are declared as file inputs with relative path sensitivity, which causes Gradle to compute those inputs based on the file contents. The task, however, consumes the paths, not the contents. As a result, the task will continue to be up-to-date if files are moves to a new location, making the resulting resources file stale.

**Steps to Reproduce**
Run Paparazzi tests on a project, move the project to a new location, run tests again. Resources can no longer be found, and tests will fail with an obscure exception.

**Expected behavior**
All three inputs should be of type string, representing the paths to the files, not the files themselves.

**Additional information:**
- Paparazzi Version: 1.0.0

liutikas:
https://github.com/cashapp/paparazzi/pull/536 fixes the platformDataRoot part. It will become a dependency on the test task directly. The contents do matter for the `Test` task.

I agree that `mergeResourcesOutput` and `mergeAssetsOutput` should be treated as string inputs given that the `Test` task already takes these as inputs, although i'd argue that https://github.com/cashapp/paparazzi/blob/master/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L156 needs to be updated to make it not path sensitive.

cesar1000:
One alternative to consider is to not use a resource file at all, and instead just pass in all configuration in system properties. Splitting the configuration to a separate task feels unnecessarily and, and the end of the day, the path to the resource file is passed in as a system property already, so it makes no significant difference.

PR #707, opened 2023-01-28
Add project to test resource loading issue.

geoff-powell:
Attempt to reproduce and fix font and resource loading issue on linux. See sample stacktraces below:

```
 java.lang.NullPointerException
        at android.graphics.fonts.Font_Builder_Delegate.createBuffer(Font_Builder_Delegate.java:52)
        at android.graphics.fonts.Font$Builder.createBuffer(Font.java:264)
        at android.graphics.fonts.Font$Builder.<init>(Font.java:200)
        at android.graphics.Typeface$Builder.<init>(Typeface.java:517)
        at android.graphics.Typeface_Delegate.createFromDisk(Typeface_Delegate.java:91)
        at com.android.layoutlib.bridge.impl.ResourceHelper.getFont(ResourceHelper.java:495)
        at com.android.layoutlib.bridge.impl.ResourceHelper.getFont(ResourceHelper.java:507)
        at android.content.res.Resources_Delegate.getFont(Resources_Delegate.java:876)
        at android.content.res.Resources.getFont(Resources.java:469)
        at app.cash.paparazzi.internal.ResourcesInterceptor.intercept(ResourcesInterceptor.kt:12)
        at androidx.core.content.res.ResourcesCompat.getFont(ResourcesCompat.java)
        at com.squareup.util.android.widget.ContextsKt.getFont(Contexts.kt:23)
....
```

```
java.lang.IllegalStateException: Could not load font
    	at androidx.compose.ui.text.font.TypefaceRequestCache.runCached(FontFamilyResolver.kt:205)
    	at androidx.compose.ui.text.font.FontFamilyResolverImpl.resolve(FontFamilyResolver.kt:94)
    	at androidx.compose.ui.text.font.FontFamilyResolverImpl.resolve-DPcqOEQ(FontFamilyResolver.kt:81)
    	at androidx.compose.ui.text.platform.AndroidParagraphIntrinsics$resolveTypeface$1.invoke-DPcqOEQ(AndroidParagraphIntrinsics.android.kt:79)
    	at androidx.compose.ui.text.platform.AndroidParagraphIntrinsics$resolveTypeface$1.invoke(AndroidParagraphIntrinsics.android.kt:77)
    	at androidx.compose.ui.text.platform.extensions.TextPaintExtensions_androidKt.applySpanStyle(TextPaintExtensions.android.kt:63)
    	at androidx.compose.ui.text.platform.AndroidParagraphIntrinsics.<init>(AndroidParagraphIntrinsics.android.kt:90)
[...]

PR #720, opened 2023-02-21
Fixes issues with preparePaparazzi{Variant}Resources task that is caching incorrectly for different variants.

geoff-powell:
Fixes the issue [this pr](https://github.com/cashapp/paparazzi/pull/707) attempted to reproduce/solve.

[Investigation doc](https://docs.google.com/document/d/17jlBWeuWV0M68uYUALX1HAGcLpA6EWTcQv8l3FJ-skc/edit#) on this issue for Cash.

TL;DR
> Looking into the generated cache key it seems that it creates a build cache for the same generated key. So to recap preparePaparazziDebugResources runs around 6m into the build process and generates the correct debug resource file for paparazzi. At 9m into the build preparePaparazziReleaseResources runs to generate the correct release resource file for paparazzi but stores the output for the same cache key 8c39bb.
> So this validates why the output for the debug task contains references to release paths.

I added a test gradle plugin project here to simulate the exact issue in [the first commit](https://github.com/cashapp/paparazzi/pull/720/commits/d15ef7b348374644fbc04295d3ed449f1c00404a) - [Sample Build Failure](https://github.com/cashapp/paparazzi/actions/runs/4235006678/jobs/7358079374#step:6:446)

[The last commit](https://github.com/cashapp/paparazzi/pull/720/commits/5d4aefabe0b686677558a25f6cd76e1a86832bf9) adds variantName as a build input and causes the failing test to pass.

cc @mohsin-motorway this may fix the issue you were having as well.

Fixes issue https://github.com/cashapp/paparazzi/issues/560 as well.

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -99,6 +99,7 @@ class PaparazziPlugin : Plugin<Project> {
         val nonTransitiveRClassEnabled =
           (project.findProperty("android.nonTransitiveRClass") as? String).toBoolean()
 
+        task.variantName.set(variant.name)
         task.packageName.set(android.packageName())
         task.artifactFiles.from(packageAwareArtifacts.artifactFiles)
         task.nonTransitiveRClassEnabled.set(nonTransitiveRClassEnabled)
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -28,10 +28,13 @@ import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition
 import org.gradle.api.attributes.Attribute
+import org.gradle.api.file.Directory
+import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.FileCollection
 import org.gradle.api.internal.artifacts.transform.UnzipTransform
 import org.gradle.api.logging.LogLevel.LIFECYCLE
 import org.gradle.api.plugins.JavaBasePlugin
+import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.options.Option
@@ -43,6 +46,7 @@ import org.jetbrains.kotlin.gradle.plugin.KotlinAndroidPluginWrapper
 import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinAndroidTarget
 import java.util.Locale
+import kotlin.io.path.invariantSeparatorsPathString
 
 @Suppress("unused")
 class PaparazziPlugin : Plugin<Project> {
@@ -95,18 +99,24 @@ class PaparazziPlugin : Plugin<Project> {
         "preparePaparazzi${variantSlug}Resources",
         PrepareResourcesTask::class.java
       ) { task ->
+        fun DirectoryProperty.asRelativePathString(childDirectory: Provider<Directory>): Provider<String> =
+          flatMap { buildDir ->
+            childDirectory.map { childDir ->
+              buildDir.asFile.toPath().relativize(childDir.asFile.toPath()).invariantSeparatorsPathString
+            }
+          }
+
         val android = project.extensions.getByType(BaseExtension::class.java)
         val nonTransitiveRClassEnabled =
           (project.findProperty("android.nonTransitiveRClass") as? String).toBoolean()
 
-        task.variantName.set(variant.name)
         task.packageName.set(android.packageName())
         task.artifactFiles.from(packageAwareArtifacts.artifactFiles)
         task.nonTransitiveRClassEnabled.set(nonTransitiveRClassEnabled)
-        task.mergeResourcesOutput.set(mergeResourcesOutputDir)
+        task.mergeResourcesOutput.set(project.layout.buildDirectory.asRelativePathString(mergeResourcesOutputDir))
         task.targetSdkVersion.set(android.targetSdkVersion())
         task.compileSdkVersion.set(android.compileSdkVersion())
-        task.mergeAssetsOutput.set(mergeAssetsOutputDir)
+        task.mergeAssetsOutput.set(project.layout.buildDirectory.asRelativePathString(mergeAssetsOutputDir))
         task.paparazziResources.set(project.layout.buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
       }
 
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -84,7 +84,8 @@ class PaparazziPlugin : Plugin<Project> {
       val mergeAssetsProvider =
         project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
       val mergeAssetsOutputDir = mergeAssetsProvider.flatMap { it.outputDir }
-      val reportOutputDir = project.layout.buildDirectory.dir("reports/paparazzi")
+      val buildDirectory = project.layout.buildDirectory
+      val reportOutputDir = buildDirectory.dir("reports/paparazzi")
       val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
 
       val packageAwareArtifacts = project.configurations
@@ -99,13 +100,6 @@ class PaparazziPlugin : Plugin<Project> {
         "preparePaparazzi${variantSlug}Resources",
         PrepareResourcesTask::class.java
       ) { task ->
-        fun DirectoryProperty.asRelativePathString(childDirectory: Provider<Directory>): Provider<String> =
-          flatMap { buildDir ->
-            childDirectory.map { childDir ->
-              buildDir.asFile.toPath().relativize(childDir.asFile.toPath()).invariantSeparatorsPathString
-            }
-          }
-
         val android = project.extensions.getByType(BaseExtension::class.java)
         val nonTransitiveRClassEnabled =
           (project.findProperty("android.nonTransitiveRClass") as? String).toBoolean()
@@ -113,11 +107,11 @@ class PaparazziPlugin : Plugin<Project> {
         task.packageName.set(android.packageName())
         task.artifactFiles.from(packageAwareArtifacts.artifactFiles)
         task.nonTransitiveRClassEnabled.set(nonTransitiveRClassEnabled)
-        task.mergeResourcesOutput.set(project.layout.buildDirectory.asRelativePathString(mergeResourcesOutputDir))
+        task.mergeResourcesOutputDir.set(buildDirectory.asRelativePathString(mergeResourcesOutputDir))
         task.targetSdkVersion.set(android.targetSdkVersion())
         task.compileSdkVersion.set(android.compileSdkVersion())
-        task.mergeAssetsOutput.set(project.layout.buildDirectory.asRelativePathString(mergeAssetsOutputDir))
-        task.paparazziResources.set(project.layout.buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
+        task.mergeAssetsOutputDir.set(buildDirectory.asRelativePathString(mergeAssetsOutputDir))
+        task.paparazziResources.set(buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
       }
 
       val testVariantSlug = testVariant.name.capitalize(Locale.US)
@@ -163,7 +157,7 @@ class PaparazziPlugin : Plugin<Project> {
         test.systemProperties["paparazzi.test.resources"] =
           writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path
         test.systemProperties["paparazzi.build.dir"] =
-          project.layout.buildDirectory.get().toString()
+          buildDirectory.get().toString()
 
         test.inputs.dir(mergeResourcesOutputDir)
         test.inputs.dir(mergeAssetsOutputDir)
@@ -274,6 +268,13 @@ class PaparazziPlugin : Plugin<Project> {
       ?: DEFAULT_COMPILE_SDK_VERSION.toString()
   }
 }
+
+private fun Directory.relativize(child: Directory): String {
+  return asFile.toPath().relativize(child.asFile.toPath()).invariantSeparatorsPathString
+}
+
+private fun DirectoryProperty.asRelativePathString(child: Provider<Directory>): Provider<String> =
+  flatMap { root -> child.map { root.relativize(it) } }
 
 // TODO: Migrate to ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE when Gradle 7.3 is
 //  acceptable as the minimum supported version
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -17,13 +17,10 @@ package app.cash.paparazzi.gradle
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileCollection
-import org.gradle.api.file.Directory
-import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.InputFiles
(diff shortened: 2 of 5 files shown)


gradlew.bat
```diff
@@ -1,92 +1,92 @@
...
-set EXIT_CODE=%ERRORLEVEL%
-if %EXIT_CODE% equ 0 set EXIT_CODE=1
-if not ""=="%GRADLE_EXIT_CONSOLE%" exit %EXIT_CODE%
-exit /b %EXIT_CODE%
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
TWiStErRob, line 1: my bad https://github.com/cashapp/paparazzi/pull/721

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -114,6 +114,53 @@ class PaparazziPluginTest {
...
+
+    with(firstRun.task(":preparePaparazziDebugResources")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isNotEqualTo(FROM_CACHE)
+    }
+
+    with(firstRun.task(":preparePaparazziReleaseResources")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isNotEqualTo(FROM_CACHE)
+    }
```
geoff-powell, line 135: Added these checks here as they should never use cache on the first run. This will fail due to the task inputs being treated by gradle as the same and therefore the `preparePaparazziReleaseResources` is cached and the `preparePaparazziDebugResources` uses that cache instead of running.

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -114,6 +114,53 @@ class PaparazziPluginTest {
...
+
+    var resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    assertThat(resourcesFile.exists()).isTrue()
+    var resourceFileContents = resourcesFile.readLines()
+    assertThat(resourceFileContents.any { it.contains("release") }).isFalse()
+
+    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/release/resources.txt")
+    assertThat(resourcesFile.exists()).isTrue()
+    resourceFileContents = resourcesFile.readLines()
+    assertThat(resourceFileContents.any { it.contains("debug") }).isFalse()
```
geoff-powell, line 145: These will verify that the resource file shouldn't have any paths with `debug` for release or vice-versa. The first commit doesn't verify this but it does fail if you comment out the task outcome assertions above.

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -114,6 +114,53 @@ class PaparazziPluginTest {
...
+
+    with(secondRun.task(":preparePaparazziDebugResources")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isEqualTo(FROM_CACHE)
+    }
+
+    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    assertThat(resourcesFile.exists()).isTrue()
+    resourceFileContents = resourcesFile.readLines()
+    assertThat(resourceFileContents.any { it.contains("release") }).isFalse()
```
geoff-powell, line 161: this will also fail for the same reason above. Uses release resource output for `preparePaparazziDebugResources` as it is cached with the same key as the previous `preparePaparazziDebugResources`

gradlew.bat
geoff-powell: rebasing off of master as this seems to be fixed

geoff-powell:
After talking with @jrodbx yesterday, the fix I had here works but may not be the final answer. The main issue is that the paths for inputs `mergeResourcesOutput` and `mergeAssetsOutput` are incorrectly being treated as the same, even though the paths contain debug or release folders. I began investigating [@PathSensitity annotation](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/PathSensitivity.html) used to see why these are treated as the same.
 
#### PathSensitivity.RELATIVE (Issue)
When using this path sensitivity, the “root” directory is /paparazzi/sample/build/intermediates/merged_res/{variant}/. This means that the task sees the input files of the directory as `ADDED: drawable/camera.png`

#### PathSensitivity.NAME_ONLY
This reports the filenames only in the directory ADDED: camera.png which again is not what we want.

#### PathSensitivity.ABSOLUTE
This reports the absolute file path which is great for separation but doesn’t scale to different machines and environments. ADDED: /{sourceCodeRoot}/Development/paparazzi/sample/build/intermediates/merged_res/debug/drawable/camera.png

#### PathSensitivity.NONE
This should not be used here as the files in the directory are ignored.


[Very important note in gradle documentation](https://docs.gradle.org/current/userguide/build_cache_concepts.html#reuse_of_outputs_between_different_tasks) that task caching is that Gradle will try and reuse outputs for the different task given the inputs. This is basically the root cause of the issue as we are using relative paths for the directory inputs, which are identical for different task variants.

The only way to fix this issue is to do one of the following options:
1. Add a new input with the variant of the task to ensure they are different
2. Remove `@CacheableTask` annotation of the task to remove the ability to cache it and require it to run every time

I suggest the first option (the one in the PR below) as the best option as it only provides extra metadata as input and allows for the task to still be cachable.

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -79,27 +73,22 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+        it.write(mergeAssetsOutput.get())
         it.newLine()
         it.write(resourcePackageNames)
         it.newLine()
       }
   }
-
-  private fun Directory.relativize(child: Directory): String {
-    return asFile.toPath().relativize(child.asFile.toPath()).toFile().invariantSeparatorsPath
-  }
```
geoff-powell, line 104: Moved this function to the plugin to relativize the path input for mergeAssetsOutput and mergeResourcesOutput

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -95,17 +99,24 @@ class PaparazziPlugin : Plugin<Project> {
         "preparePaparazzi${variantSlug}Resources",
         PrepareResourcesTask::class.java
       ) { task ->
+        fun DirectoryProperty.asRelativePathString(childDirectory: Provider<Directory>): Provider<String> =
+          flatMap { buildDir ->
+            childDirectory.map { childDir ->
+              buildDir.asFile.toPath().relativize(childDir.asFile.toPath()).invariantSeparatorsPathString
+            }
+          }
+
```
geoff-powell, line 108: Using mappings here to keep this property lazy for configuration perf

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -32,22 +29,21 @@ import org.gradle.api.tasks.TaskAction
...
 abstract class PrepareResourcesTask : DefaultTask() {
+
   @get:Input
   abstract val packageName: Property<String>
 
-  @get:InputDirectory
-  @get:PathSensitive(PathSensitivity.RELATIVE)
-  abstract val mergeResourcesOutput: DirectoryProperty
+  @get:Input
+  abstract val mergeResourcesOutput: Property<String>
```
swankjesse, line 37: It has to be a String? Gradle is so annoying

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -114,6 +114,53 @@ class PaparazziPluginTest {
...
 
+  @Test
+  fun prepareResourcesCaching() {
+    val fixtureRoot = File("src/test/projects/prepare-resources-task-caching")
+
+    val firstRun = gradleRunner
+      .withArguments("testRelease", "testDebug", "--build-cache", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    println(firstRun.output)
```
swankjesse, line 125: nit: remove?

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -114,6 +114,53 @@ class PaparazziPluginTest {
...
+
+    with(firstRun.task(":preparePaparazziReleaseResources")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isNotEqualTo(FROM_CACHE)
+    }
+
+    var resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    assertThat(resourcesFile.exists()).isTrue()
+    var resourceFileContents = resourcesFile.readLines()
+    assertThat(resourceFileContents.any { it.contains("release") }).isFalse()
```
swankjesse, line 140: Nice

paparazzi/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
swankjesse: Nice test strategy!

paparazzi/paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+  id 'kotlin-android'
+  id 'app.cash.paparazzi'
+}
+
+repositories {
+  maven {
+    url "file://${projectDir.absolutePath}/../../../../../build/localMaven"
+  }
+  mavenCentral()
+  //mavenLocal()
```
swankjesse, line 12: nit: remove?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
geoff-powell: Yep, basically `DirectoryProperty` with `@PathSensitive(PathSensitivity.RELATIVE)` is used for a directory where you want an sub folder or file changes to run the task. Since its relative and the base directory is `.../debug` and `.../release` gradle sees the underlying files as the same.

The suggestion for using Property<String> here is if you only want the path to be used as an input and the contents don't matter. This was our exact usecase as we just use the path in the generated `resources.txt` file.

geoff-powell:
> Very much looking forward to having these fixes. THANK YOU!

Ya no problem! Fixed up your suggested changes. Will merge once CI passes 👍🏽

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -32,22 +29,21 @@ import org.gradle.api.tasks.TaskAction
 
 @CacheableTask
 abstract class PrepareResourcesTask : DefaultTask() {
+
```
jrodbx, line 32:
```suggestion
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
jrodbx: can we move this function as is to the plugin as a private extension method?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -59,8 +55,6 @@ abstract class PrepareResourcesTask : DefaultTask() {
   @get:OutputFile
   abstract val paparazziResources: RegularFileProperty
 
-  private val buildDirectory = project.layout.buildDirectory
```
jrodbx, line 62: can we move this to the plugin inside the task configuration block?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -95,17 +99,24 @@ class PaparazziPlugin : Plugin<Project> {
...
         val android = project.extensions.getByType(BaseExtension::class.java)
         val nonTransitiveRClassEnabled =
           (project.findProperty("android.nonTransitiveRClass") as? String).toBoolean()
 
         task.packageName.set(android.packageName())
         task.artifactFiles.from(packageAwareArtifacts.artifactFiles)
         task.nonTransitiveRClassEnabled.set(nonTransitiveRClassEnabled)
-        task.mergeResourcesOutput.set(mergeResourcesOutputDir)
+        task.mergeResourcesOutput.set(project.layout.buildDirectory.asRelativePathString(mergeResourcesOutputDir))
```
jrodbx, line 116:
```suggestion
        task.mergeResourcesOutput.set(buildDirectory.asRelativePathString(mergeResourcesOutputDir))
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -95,17 +99,24 @@ class PaparazziPlugin : Plugin<Project> {
...
         task.targetSdkVersion.set(android.targetSdkVersion())
         task.compileSdkVersion.set(android.compileSdkVersion())
-        task.mergeAssetsOutput.set(mergeAssetsOutputDir)
+        task.mergeAssetsOutput.set(project.layout.buildDirectory.asRelativePathString(mergeAssetsOutputDir))
```
jrodbx, line 119:
```suggestion
        task.mergeAssetsOutput.set(buildDirectory.asRelativePathString(mergeAssetsOutputDir))
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
jrodbx, line 37:
```suggestion
  abstract val mergeResourcesOutputDir: Property<String>
```

...since it's now a String property

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -32,22 +29,21 @@ import org.gradle.api.tasks.TaskAction
...
   abstract val targetSdkVersion: Property<String>
 
   @get:Input
   abstract val compileSdkVersion: Property<String>
 
-  @get:InputDirectory
-  @get:PathSensitive(PathSensitivity.RELATIVE)
-  abstract val mergeAssetsOutput: DirectoryProperty
+  @get:Input
+  abstract val mergeAssetsOutput: Property<String>
```
jrodbx, line 46:
```suggestion
  abstract val mergeAssetsOutputDir: Property<String>
```

...since it's now a String property

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -95,17 +99,24 @@ class PaparazziPlugin : Plugin<Project> {
...
         task.paparazziResources.set(project.layout.buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
```
jrodbx, line 120:
```suggestion
        task.paparazziResources.set(buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
```

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx, line 108: Let's keep this but also:
-  copy over the previous relativize function
- move both out of local scope and to the bottom of the plugin class declaration

i.e., 
```
  ...

  private fun Directory.relativize(child: Directory): String {
    return asFile.toPath().relativize(child.asFile.toPath()).invariantSeparatorsPathString
  }

  private fun DirectoryProperty.asRelativePathString(child: Provider<Directory>): Provider<String> =
    flatMap { root -> child.map { root.relativize(it) } }
}

// TODO: Migrate to ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE when Gradle 7.3 is
```

jrodbx (approved):
Nice work, def learned a lot more about Input properties as a result of this PR!

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
oldergod: The worst is when Gradle simply ignores your parameter without telling you it's not the right type and that what you expect to happen is never gonna be!

PR #721, opened 2023-02-21
Renormalize gradlew.bat

TWiStErRob:
Follow-up on #691, I [just learned](https://github.com/gradle-nexus/publish-plugin/pull/180) we need to commit the .bat file with `LF` line endings, and on checkout `git` will make it `CRLF` based on `.gitattributes`.


geoff-powell (approved):

jrodbx (approved):

jrodbx:
Was wondering about this, when cloning on my new m1.  Thanks!

PR #727, opened 2023-03-02
Do not log report file URI if it doesn't exist

atomgomba:
This is something I've found confusing when I was looking for the file and it wasn't actually there.

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -43,6 +43,7 @@ import org.jetbrains.kotlin.gradle.plugin.KotlinAndroidPluginWrapper
 import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinAndroidTarget
 import java.util.Locale
+import kotlin.io.path.exists
 
 @Suppress("unused")
 class PaparazziPlugin : Plugin<Project> {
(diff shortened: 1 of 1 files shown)


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -186,8 +187,12 @@ class PaparazziPlugin : Plugin<Project> {
...
         test.doLast(object : Action<Task> {
           override fun execute(t: Task) {
-            val uri = reportOutputDir.get().asFile.toPath().resolve("index.html").toUri()
-            test.logger.log(LIFECYCLE, "See the Paparazzi report at: $uri")
+            val path = reportOutputDir.get().asFile.toPath().resolve("index.html")
+            if (path.exists()) {
+              test.logger.log(LIFECYCLE, "See the Paparazzi report at: ${path.toUri()}")
+            } else {
+              test.logger.log(LIFECYCLE, "Paparazzi failure report not generated.")
+            }
```
jrodbx, line 195: Can you elaborate more what was the executed scenario that did not result in a report being generated?

atomgomba: When there was no failure that file will not exist at that path

TWiStErRob: I would say in that case just don't output anything? When unit tests pass there's no output either.

jrodbx: > When there was no failure that file will not exist at that path

This isn't a failure path, but rather a report that should always be generated.  Did you mean, when there *is* a failure?

Issue #732, opened 2023-03-08
When I run the tests I am getting the Unknown field type in R class: class error


jd-alexander:
**Description**

I have integrated Paparazzi into a repository at work and I am getting this error: 
```
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
Mar 08, 2023 5:42:02 PM app.cash.paparazzi.internal.PaparazziLogger error
SEVERE: Unknown field type in R class: class [Z
[...]

PR #733, opened 2023-03-09
Improve error message to help diagnose #732

TWiStErRob:
Improve error message to help diagnose #732. `java.lang.reflect.Field` has a beautiful toString including the full signature. This should help pinpoint which field is the problem (name), in which `R` class is the problem (package in FQCN) and give a nicer human-readable type message on top of the existing one (`type`). I left the original in there for future-proofing: if the code changes and the `type` variable gets calculated differently, the error message will be still correct and help figure out the problem.

See https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Field.html#toString--

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
@@ -74,7 +74,7 @@ internal class PaparazziCallback(
             } else if (type.isArray && type.componentType == Int::class.javaPrimitiveType) {
               // Ignore.
             } else {
-              logger.error(null, "Unknown field type in R class: $type")
+              logger.error(null, "Unknown field type ($type) in R class: $field")
             }
           } catch (e: IllegalAccessException) {
             logger.error(e, "Malformed R class: %1\$s", "$rPackageName.R")


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
```diff
@@ -74,7 +74,7 @@ internal class PaparazziCallback(
             } else if (type.isArray && type.componentType == Int::class.javaPrimitiveType) {
               // Ignore.
             } else {
-              logger.error(null, "Unknown field type in R class: $type")
+              logger.error(null, "Unknown field type ($type) in R class: $field")
```
jrodbx, line 77:
```suggestion
              logger.error(null, "Unknown field type ($type) for field (${field.name}) in class: ${field.declaringClass.name}")

```

jrodbx (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
jrodbx: Original: `Unknown field type in R class: int`
PR: `Unknown type (int) in R class: public static int app.cash.paparazzi.sample.R$color.cameraBody`
Suggestion:  `Unknown type (int) for field (cameraBody) in R class: app.cash.paparazzi.sample.R$color`

TWiStErRob: I would rather not go more specific, the original had the same problem, it hides info. `private` and `transient` might give more clues, as seen in the original issue. And we don't know what else we would be hiding by not using toString directly from Java. Since this is not a normal behaviour, but an edge case, the more info we get, the better, and the format doesn't really matter.

I would be happy to settle on a compromise, which reads better, but doesn't hide:
```suggestion
logger.error(null, "Unknown field type $type for field ${field.name} in class: ${field.declaringClass.name}, full signature: ${field}.")
```

jrodbx, line 77: Fair point, then l suggest updating the message text like so:
```suggestion
              logger.error(null, "Unknown type ($type) in R class field: $field")
```

Issue #513, opened 2022-07-23
Compose async effects are halted


yschimke:
**Description**

Compose async effects like LaunchedEffect, don't appear to run naturally.

**Steps to Reproduce**

```
    @Test
    fun delay() {
        val start = System.currentTimeMillis()
        println(0)
        paparazzi.gif(start = 2500L, end = 2500L, fps = 1) {
            var state by remember { mutableStateOf(0) }
            LaunchedEffect(Unit) {
                println("${System.currentTimeMillis() - start} Launched Effect")
                state++
                while (true) {
                    delay(10)
                    state++
                    println("${System.currentTimeMillis() - start} Delayed")
                }
            }
            Box(modifier = Modifier.fillMaxSize(), contentAlignment = Center) {
                Text("Value: $state")
            }
        }
        println("" + (System.currentTimeMillis() - start) + " end")
    }
```

Text output is like

```
0
Jul. 23, 2022 4:19:10 PM app.cash.paparazzi.internal.PaparazziLogger logAndroidFramework
INFO: Settings [3]: Can't get key animator_duration_scale from content://settings/global
470 Launched Effect
485 Delayed
812 end
```

Screen Output is "Value: 2"

**Expected behavior**

Output is some high number like 51?

**Additional information:**
- Paparazzi Version: 1.0

yschimke:
Switching onto Dispatchers.Default fixes it, so possibly it is contention on the main thread?

yschimke:
As another example

```
    @Test
    fun mediaPlayerScreen() {
        paparazzi.gif(start = 500L, end = 5000L, fps = 1) {
            val state = rememberScalingLazyListState()
            ScalingLazyColumn(
                horizontalAlignment = CenterHorizontally,
                state = state
            ) {
                items(10) {
                    Text(text = "Item $it")
                }
            }
        }
    }
```    

This video https://user-images.githubusercontent.com/231923/180612278-93e83b60-2a49-4b37-8d1e-94620cc16c7f.mov

Shows the delay in rendering.

yschimke:
From reading the code, my impression is that there is some missing logic with how Paparazzi simulates time.  But I'm out of my depth.

```
  private fun withTime(
    timeNanos: Long,
    block: () -> Unit
  ) {
    val frameNanos = TIME_OFFSET_NANOS + timeNanos
```

Does the existing notion of frame time integrate with Compose currently?

BenTilbrook:
[...]

PR #739, opened 2023-03-22
Use Dispatchers.Main for `delay` functionality

gamepro65:
When using `delay` with LaunchedEffect, ex. to delay animation, the default implementation uses a separate DefaultExecutor for tracking time causing delay to be tied to unit test time not Paparazzi frame time. By taking advantage of this system property we force `delay` to use Main which is driven by Paparazzi's frame time.

System Property: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt#L11-L25

Resolves the timing issues reported as part of https://github.com/cashapp/paparazzi/issues/513

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -173,6 +173,7 @@ class PaparazziPlugin : Plugin<Project> {
               nativePlatformFileCollection.singleFile.absolutePath
             test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
             test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+            test.systemProperties["kotlinx.coroutines.main.delay"] = true
             test.systemProperties.putAll(paparazziProperties)
           }
         })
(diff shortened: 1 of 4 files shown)


saket (approved):

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -173,6 +173,7 @@ class PaparazziPlugin : Plugin<Project> {
               nativePlatformFileCollection.singleFile.absolutePath
             test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
             test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+            test.systemProperties["kotlinx.coroutines.main.delay"] = true
```
jrodbx, line 176: spoke offline, slight concern about how to catch if this property key changes other than the backing test (which will fail, but might not obvious to why?)

jrodbx (approved):

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: Looks like it is a documented property in coroutines 1.6.0:
https://github.com/Kotlin/kotlinx.coroutines/blob/master/CHANGES.md#version-160
Strangely, according to that wording it should be defaulting to true.

PR #740, opened 2023-03-22
Recomposition does not happen unless lifecycle is RESUMED

gamepro65:
By making this change we will end up with 2 compositions before the first frame, one on view attach and the other when executing handler callbacks before taking the snapshot. I've confirmed this logic is the same behavior we see when executing on device in that we have an initial composition followed by a recomposition then finally the first frame draw.


saket (approved):

geoff-powell (approved):

jrodbx (approved):
nice!

PR #748, opened 2023-04-05
Add regression tests for rerunning Paparazzi

TWiStErRob:
This expected to fail since #747 is a problem and it'll be only fixed when #690 is merged.


colinmarsch:
This PR was superseded by the merged PR https://github.com/cashapp/paparazzi/pull/2089

Issue #524, opened 2022-07-30
Rewrite resource loading mechanism


jrodbx:
During our hack week, we used `com.android.ide.common.resources.deprecated.ResourceRepository` mainly due its simplicity -- pass it a file path to a resource source set directory, and it will parse and load the contents for later retrieval and resolution for free. However, this left a bit of tech debt.

First, this ResourceRepository definition has since [been deleted](https://android.googlesource.com/platform/tools/base/+/f870b9e262973c0742380e8a5fbb74977b44fa14) from `sdk-common` and replaced with a different [interface](https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/sdk-common/src/main/java/com/android/ide/common/resources/ResourceRepository.java) implemented in the [Android Studio project](https://android.googlesource.com/platform/tools/adt/idea/+/refs/heads/mirror-goog-studio-main/android/src/com/android/tools/idea/res/).  As a result, we can't bump some of LayoutLib's transitive dependencies (`sdk-common`, `layoutlib-api`).

Second, this also couples us slightly to AGP's merged resources dir, which contains all the transitive resources that drive Paparazzi.  Following the new ResourceRepository approach would instead reference and parse the project resource files in `/src/main/res` (and variants), artifact resource files in aar/res, as well as the framework resources, provided by layoutlib, directly.

Since Paparazzi tests should also run on the CLI, the new ResourceRepository implementation cannot use a similar approach and use Intellij's PSI model to reference the resource source set directories.  Instead, the implementation would likely have to rely on AGP bits to obtain them (heads up, Bazel folk, this would be another thing to implements) which is still better than coupling to AGP's intermediary output directories.

Fixing this should automatically also provide support for application modules (since we wouldn't be looking into `build/intermediates/merged_res` anymore, which is in a [binary format for application modules](https://github.com/cashapp/paparazzi/issues/107)) and fix the issue with [HTML string resources](https://github.com/cashapp/paparazzi/issues/316) since a different resource parsing path is used altogether.

I've been putting this off for a while, but it's about time to tackle this since we'll be rewriting/refactoring the core of Paparazzi on the march to 2.0.

jrodbx:
This should now be fixed, and will be promoted in the upcoming (1.3.1) release.

jisungbin:
Thanks for the awesome work! 🫶🫶 (@jrodbx @kevinzheng-ap and other Paparazzi contributors!)

PR #755, opened 2023-04-19
Allow paparazzi on android application plugin for compose

Burtan:
This pull requests removes the check that forces paparazzi to be used with the android library plugin only. It also adds a test for the android application plugin for compose.


jrodbx:
Support for this is coming soon via https://github.com/cashapp/paparazzi/issues/524, which one of our internal teams has been working on.

Issue #758, opened 2023-04-22
Tests crash if Build.VERSION.CODENAME is read


amatkovsky:
**Description**
If a code path that reads `Build.VERSION.CODENAME` is hit, the test crashes. `Build.VERSION.CODENAME` is not set during environment setup.

In this particular case, it reproduces if `compileSdkVersion` is set to 32.

From `androidx.core.os.BuildCompat`:
```java
    public static boolean isAtLeastT() {
        return VERSION.SDK_INT >= 33
                || (VERSION.SDK_INT >= 32
                && isAtLeastPreReleaseCodename("Tiramisu", VERSION.CODENAME));
    }
```

Setting `compileSdkVersion` to 31 or 33 does not enter this code branch.
```
java.lang.NullPointerException: Cannot invoke "String.toUpperCase(java.util.Locale)" because "buildCodename" is null
	at androidx.core.os.BuildCompat.isAtLeastPreReleaseCodename(BuildCompat.java:64)
	at androidx.core.os.BuildCompat.isAtLeastT(BuildCompat.java:215)
	at androidx.compose.ui.text.android.BoringLayoutFactory.measure(BoringLayoutFactory.kt:49)
	at androidx.compose.ui.text.android.LayoutIntrinsics$boringMetrics$2.invoke(LayoutIntrinsics.kt:46)
	at androidx.compose.ui.text.android.LayoutIntrinsics$boringMetrics$2.invoke(LayoutIntrinsics.kt:44)
	at kotlin.UnsafeLazyImpl.getValue(Lazy.kt:81)
	at androidx.compose.ui.text.android.LayoutIntrinsics.getBoringMetrics(LayoutIntrinsics.kt:44)
	at androidx.compose.ui.text.android.LayoutIntrinsics$maxIntrinsicWidth$2.invoke(LayoutIntrinsics.kt:63)
	at androidx.compose.ui.text.android.LayoutIntrinsics$maxIntrinsicWidth$2.invoke(LayoutIntrinsics.kt:62)
```

**Steps to Reproduce**
Code to reproduce:
```kotlin
class ReproduceCodenameCrashTest {

    @get:Rule
    val paparazzi = Paparazzi(
        deviceConfig = PIXEL_5,
        theme = "android:Theme.Material.Light.NoActionBar",
        environment = detectEnvironment().copy(
            platformDir = "${androidHome()}/platforms/android-32",
            compileSdkVersion = 32
        )
    )

    @Test
[...]

PR #759, opened 2023-04-22
Set Build.VERSION.CODENAME to fix codename checks (mainly in Compose)

amatkovsky:
This aims to fix https://github.com/cashapp/paparazzi/issues/758

**Issue:**
`CODENAME` checks like below crash snapshot tests (this is `androidx.core.os.BuildCompat`):
```java
    public static boolean isAtLeastT() {
        return VERSION.SDK_INT >= 33
                || (VERSION.SDK_INT >= 32
                && isAtLeastPreReleaseCodename("Tiramisu", VERSION.CODENAME));
    }
```

**Cause:**
`Build.VERSION.CODENAME` is `null`.

**Solution:**
Set `CODENAME` reflectively during Paparazzi environment setup. 
Most of the times the value should be `REL` (for release builds).

Still, I'd like to leave a way to set it to another value (like `UpsideDownCake` for Android 14 that is being developed right now) if someone wants to run their snapshot tests against preview version of the platform.

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
@@ -29,6 +29,7 @@ data class Environment(
   val assetsDir: String,
   val packageName: String,
   val compileSdkVersion: Int,
+  val platformCodename: String = "REL",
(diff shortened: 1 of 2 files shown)


amatkovsky:
Fixed.

PR #771, opened 2023-05-05
Fix multi line content description rendering for AccessibilityRenderExtension

geoff-powell:


paparazzi/paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/test/resources/compose_a11y.png
geoff-powell, line 1: example of how multi line content description should look

jrodbx: idea: should we instead render this with the newlines, i.e. "multi\nline\ntext"?  otherwise, multiple textviews with multi-line text input might start to distort the height of the snapshot panes?

jrodbx (approved):

paparazzi/paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/test/resources/compose_a11y.png
geoff-powell: done

PR #772, opened 2023-05-05
Update GRADLE_USER_HOME on CI to avoid potential fs conflicts

jrodbx:
Per post-merge feedback from @TWiStErRob: https://github.com/cashapp/paparazzi/pull/770#discussion_r1185889434

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
-  GRADLE_USER_HOME: ${{ github.workspace }} # required for Windows due to resource relative paths
+  GRADLE_USER_HOME: ${{ github.workspace }}/gradle-home # required for Windows due to resource relative paths
 
 jobs:
   build:


.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
-  GRADLE_USER_HOME: ${{ github.workspace }} # required for Windows due to resource relative paths
+  GRADLE_USER_HOME: ${{ github.workspace }}/.gradle # required for Windows due to resource relative paths
```
TWiStErRob, line 7: I think the checkout action puts the root of the repository to the same folder, which means the `.gradle` folder will conflict with `gradlew`'s local working folder. Luckily there's no overlap with User home, that I'm aware, here's an example listing from Detekt's build:
```
8.0.1\
8.0.2\
buildOutputCleanup\
configuration-cache\
kotlin\
vcs-1\
file-system.probe
workspace-id.txt
workspace-id.txt.lock
```
vs the [Gradle home contents](https://github.com/cashapp/paparazzi/pull/770#discussion_r1185889434).

jrodbx: I think the clone would create a new folder, but nonethless I've updated to `gradle-home` to avoid any potential conflict between 
`${{ github.workspace }}/.gradle` 
and any of
`${{ github.workspace }}/.gradle/`
`${{ github.workspace }}/gradle/`
`${{ github.workspace }}/gradlew`

TWiStErRob: Did a quick test: https://github.com/TWiStErRob/github-actions-test/actions/runs/4900931461/jobs/8751909992#step:3:9

The checkout action is equivalent to `cd ${{ github.workspace }}; git clone repourl .`

So `gradlew`'s output would create `${{ github.workspace }}/.gradle/`

geoff-powell (approved):

PR #789, opened 2023-05-12
Adds tests for ResourceFolderRepository

kevinzheng-ap:

Add tests for ResourceFolderRepository. Tests are ignored now and should be used for verification when ResourceFolderRepository integrates with new resource loading

cc @jrodbx


jrodbx (approved):
nice

PR #792, opened 2023-05-14
Update external library paths to be relative to gradle user home not project root directory

jrodbx:
This should eliminate the constraint for Windows users having to have the Gradle user home and the project root on the same drive.

cc: @kevinzheng-ap

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -65,6 +65,8 @@ abstract class PrepareResourcesTask : DefaultTask() {
 
   private val projectDirectory = project.layout.projectDirectory
 
+  private val gradleUserHomeDirectory = projectDirectory.dir(project.gradle.gradleUserHomeDir.path)
+
   @TaskAction
   // TODO: figure out why this can't be removed as of Kotlin 1.6+
   @OptIn(ExperimentalStdlibApi::class)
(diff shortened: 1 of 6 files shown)


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -65,6 +65,9 @@ abstract class PrepareResourcesTask : DefaultTask() {
 
   private val projectDirectory = project.layout.projectDirectory
 
+  private val gradleUserHomeDirectory =
+    project.layout.dir(project.provider { project.gradle.gradleUserHomeDir }).get()
```
JakeWharton, line 69: Does going through the provider buy you anything? There's an overload which takes a path string directly.
```suggestion
    project.layout.dir(project.gradle.gradleUserHomeDir.path)
```

JakeWharton (approved):

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
jrodbx: Oh nice, when reading the docs, it seemed like dir(String) would resolve the path relative to the project root directory, whereas the Provider wouldn't.  But if this works, I like this much better!

jrodbx: actually, I don't see an overload for `ProjectLayout.dir(String file)`; did you mean `Project.getProjectDirectory().dir(String path)`?

PR #793, opened 2023-05-14
Migrate from AssertJ to Truth

jrodbx:


JakeWharton:
Skip Truth. Go right to https://github.com/willowtreeapps/assertk/.

JakeWharton (approved):

PR #823, opened 2023-05-14
Avoid building code on mkdoc dependency updates

jrodbx:


JakeWharton (approved):
This doesn't preclude the other actions from running

jrodbx:
Closing for now, since it looks like mkdocs dependency updates are over with....for now!

PR #765, opened 2023-05-03
Override andorid.os.Build class fields to android.os_Original_Build

geoff-powell:
fixes #380, #758, #631

Currently using reflection to map these classes. As @JerryLocke [points out, LayoutLib is using a class loader](https://github.com/cashapp/paparazzi/issues/91#issuecomment-545724167) to swap `_Original_Build` for `Build` since the `SystemProperties` are loaded from `Bridge.init()` and static fields of `Build` are configured at application start. 

The final solution _should_ involve [using a custom class loader like AndroidStudio](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:layoutlib/src/com/android/layoutlib/LayoutlibClassLoader.java;bpv=0) uses or class replacement via bytecode manipulation with [ByteBuddy](https://bytebuddy.net/#/) or direct [Java ASM work](https://asm.ow2.io/).

The added test here mainly verifies issue #380 where Material TextInputLayout couldn't be used since it does a `Build.*` check when attached to the window.

PR #845, opened 2023-05-15
Use truth for BuildClass plugin tests instead of plain assert

geoff-powell:
Per[ this comment](https://github.com/cashapp/paparazzi/pull/765#discussion_r1184687187)

Migrated `asset` to truth `assertThat`


JakeWharton:
Tests always run with assertions enabled so that's not really a concern. Still a good idea to migrate for better failure messages, though.

JakeWharton:
I think it's somewhat an every-build-system feature but merely by convention. Obviously nothing forces you to do this, but of the 3 or 4 build systems for Java/the JVM that I've used they all have done it.

JakeWharton (approved):

jrodbx (approved):

PR #849, opened 2023-05-15
Switch to BuildConfig plugin to generate plugin version info

jrodbx:


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -15,8 +15,6 @@
  */
 package app.cash.paparazzi.gradle
 
-import app.cash.paparazzi.NATIVE_LIB_VERSION
-import app.cash.paparazzi.VERSION
```
jrodbx, line 19: gone, because I also subtlety switched the namespace of the generated vars from `app.cash.paparazzi` to `app.cash.paparazzi.gradle`.

JakeWharton (approved):

PR #852, opened 2023-05-15
Snapshot output directory should be a task input not output

jrodbx:
We want Gradle's task invalidation logic to rerun tests when the snapshot output directory is changed, which was the motivation behind https://github.com/cashapp/paparazzi/pull/206.  However, by adding it as an _output_ property, Gradle 8.0 now throws the following error:

```
A problem was found with the configuration of task ':sample:testReleaseUnitTest' (type 'AndroidUnitTest').
  - Gradle detected a problem with the following location: '/home/runner/work/paparazzi/paparazzi/sample/src/test/snapshots'.
    
    Reason: Task ':sample:spotlessKotlin' uses this output of task ':sample:testReleaseUnitTest' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':sample:testReleaseUnitTest' as an input of ':sample:spotlessKotlin'.
      2. Declare an explicit dependency on ':sample:testReleaseUnitTest' from ':sample:spotlessKotlin' using Task#dependsOn.
      3. Declare an explicit dependency on ':sample:testReleaseUnitTest' from ':sample:spotlessKotlin' using Task#mustRunAfter.
    
    Please refer to https://docs.gradle.org/8.1.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

Spotless doesn't care about snapshots, and output side effects of a task don't necessarily == task outputs, so this fixes both.

Side note: This change specifically uses `inputs.files(...)`, instead of `inputs.dir(...)`, since the latter requires the folder to exist per Gradle conventions: https://docs.gradle.org/7.6.1/userguide/validation_problems.html#input_file_does_not_exist


paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -179,9 +179,11 @@ class PaparazziPlugin : Plugin<Project> {
         test.inputs.files(nativePlatformFileCollection)
           .withPropertyName("paparazzi.nativePlatform")
           .withPathSensitivity(PathSensitivity.NONE)
+        test.inputs.files(snapshotOutputDir)
```
TWiStErRob, line 182: Add "side note" as a comment here to prevent inadvertent future "optimizations"?

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -179,9 +179,11 @@ class PaparazziPlugin : Plugin<Project> {
...
+          .withPropertyName("paparazzi.snapshot.output.dir")
+          .withPathSensitivity(PathSensitivity.RELATIVE)
 
         test.outputs.dir(reportOutputDir)
-        test.outputs.dir(snapshotOutputDir)
```
TWiStErRob, line 184: I have a feeling this ambiguity comes from the fact that the same task can run in record and verify modes. In record mode the output **is** snapshots, and in verify mode snapshots are input and the output is reports. Does this mean that adding `snapshotOutputDir` as input/output should be based on the `isRecordRun`/`isVerifyRun` properties? (e.g. something like `isRecordRun.map { if (it) snapshotOutputDir else "dummyNonExistentFolder" }`)?

geoff-powell (approved):

geoff-powell (approved):

paparazzi/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -179,9 +179,11 @@ class PaparazziPlugin : Plugin<Project> {
         test.inputs.files(nativePlatformFileCollection)
           .withPropertyName("paparazzi.nativePlatform")
           .withPathSensitivity(PathSensitivity.NONE)
+        test.inputs.files(snapshotOutputDir)
+          .withPropertyName("paparazzi.snapshot.output.dir")
+          .withPathSensitivity(PathSensitivity.RELATIVE)
```
ribafish, line 184:
```suggestion
        if (isVerifyRun.getOrElse(false)) {
          test.inputs.dir(snapshotOutputDir)
            .withPropertyName("paparazzi.snapshot.output.dir")
            .withPathSensitivity(PathSensitivity.RELATIVE)
        }
        if (isRecordRun.getOrElse(false)) {
          test.outputs.dir(snapshotOutputDir)
        }
```

Similar to @TWiStErRob, I would suggest wrapping this in ifs, as the task can run in different modes, which causes it to actually have different inputs & outputs. In verify mode, `snapshotOutputDir` is an input, as the captured snapshots are compared to the ones stored there. In record mode `snapshotOutputDir` is an output, as that's where the captured snapshots are copied. If in neither (running just `testDebug` for example), the `snapshotOutputDir` doesn't have anything to do with it, so it also shouldn't be tracked. When done like suggested, the only case where Gradle will still be confused is if one tries to run both `verifyPaparazzi` and `recordPaparazzi` tasks <- this should be solved by the copying from `build/reports/paparazzi` to `src/snapshots` to actually happen in the `recordPaparazzi` task and not in the test task, however that is bigger change. For now, I'd either crash the build if both are requested, or perhaps print a warning about it - the build will still go through, just caching will be disabled in this case.

jrodbx:
Superceded by https://github.com/cashapp/paparazzi/pull/1522.

PR #858, opened 2023-05-16
Use accessibilityText as an identifier instead of id as it's backed by AtomicInteger and isn't consistent across runs

geoff-powell:


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -26,6 +26,7 @@ import androidx.compose.ui.platform.ViewRootForTest
 import androidx.compose.ui.semantics.SemanticsActions
 import androidx.compose.ui.semantics.SemanticsNode
 import androidx.compose.ui.semantics.SemanticsProperties
+import androidx.compose.ui.semantics.contentDescription
```
jrodbx, line 29:
```suggestion
```

jrodbx (approved):

adamalyyan:
Sweet! Thank you for digging into this!

PR #860, opened 2023-05-16
Switch from kapt to ksp for Moshi codegen

jrodbx:
Unblocks Gradle 8 migration.

Specifically, fixes a Spotless error requesting that spotless depend on output from kapt


JakeWharton (approved):
Needs grouped with Kotlin in renovate JSON too.

jrodbx:
> Needs grouped with Kotlin in renovate JSON too.

Cool, will do in follow-up.

PR #865, opened 2023-05-17
Remove legacy manifest namespacing support

jrodbx:
Introduced in the AGP 4.2 to 7.0 transition; this is now the expected approach in modern Android projects


Egorand (approved):
What will the failure look like in projects that still declare the package name in the manifest?

TWiStErRob:
AGP 8 makes this the only way: https://developer.android.com/build/releases/gradle-plugin#namespace-dsl

So every project will have to migrate eventually. I guess for those who are still using the old approach the currently released Paparazzi versions are the way.

PR #868, opened 2023-05-18
Extract test project repository boilerplate into common settings.gradle

jrodbx:
Simpler than it looks.

Cut out of N-1 files, paste here: https://github.com/cashapp/paparazzi/compare/jrod/2023-05-17/extract-repository-boilerplate?expand=1#diff-9deddf6381b9e947a391a23b3234775324bc7345724db33f6d6dff6002093d7a

### DIFF
--- a/paparazzi/paparazzi-gradle-plugin/src/test/projects/test.settings.gradle
+++ b/paparazzi/paparazzi-gradle-plugin/src/test/projects/test.settings.gradle
@@ -4,4 +4,13 @@ dependencyResolutionManagement {
       from(files("../../../../../gradle/libs.versions.toml"))
     }
   }
+
+  repositories {
+    maven {
+      url "file://${rootDir}/../../../../../build/localMaven"
+    }
+    mavenCentral()
+    //mavenLocal()
+    google()
+  }
 }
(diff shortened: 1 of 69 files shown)


paparazzi/paparazzi-gradle-plugin/src/test/projects/test.settings.gradle
```diff
@@ -4,4 +4,13 @@ dependencyResolutionManagement {
       from(files("../../../../../gradle/libs.versions.toml"))
     }
   }
+
+  repositories {
```
TWiStErRob, line 8: What do you think about adding "fail on project repos" or "prefer ..." to ensure that there's no mixing between settings.gradle repos and build.gradle repos, it's either-or.

geoff-powell (approved):
Nice 👍

paparazzi/paparazzi-gradle-plugin/src/test/projects/test.settings.gradle
jrodbx: default is [PREFER_PROJECT](https://github.com/gradle/gradle/blob/d62830774ae867342c1b2b0e73f7575df953aa26/subprojects/dependency-management/src/main/java/org/gradle/internal/management/DefaultDependencyResolutionManagement.java#L85), which seems fine to me, as some test projects may want custom libraries to assert expected behaviors.

PR #871, opened 2023-05-18
Port  BasicArrayResourceItem from tools/base/resource-repository

kevinzheng-ap:

cc @jrodbx

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.paparazzi.internal.resources.base
+
+import app.cash.paparazzi.internal.resources.ResourceSourceFile
+import com.android.ide.common.rendering.api.ArrayResourceValue
+import com.android.resources.ResourceType
+import com.android.resources.ResourceVisibility
+import java.util.Collections
+
+/**
+ * Ported from: [BasicArrayResourceItem.java](https://cs.android.com/android-studio/platform/tools/base/+/18047faf69512736b8ddb1f6a6785f58d47c893f:resource-repository/main/java/com/android/resources/base/BasicArrayResourceItem.java)
+ *
+ * Resource item representing an array resource.
+ */
+class BasicArrayResourceItem(
+  name: String,
+  sourceFile: ResourceSourceFile,
+  visibility: ResourceVisibility,
+  private val elements: List<String>,
+  private val defaultIndex: Int
+) : BasicValueResourceItemBase(ResourceType.ARRAY, name, sourceFile, visibility), ArrayResourceValue {
+  init {
+    assert(elements.isEmpty() || defaultIndex < elements.size)
+  }
+
+  override fun getElementCount() = elements.size
+
+  override fun getElement(index: Int) = elements[index]
+
+  override fun iterator() = Collections.unmodifiableList(elements).iterator()
+
+  override fun getValue(): String? = if (elements.isEmpty()) null else elements[defaultIndex]
+
+  override fun equals(obj: Any?): Boolean {
+    if (this === obj) return true
+    if (!super.equals(obj)) return false
+    val other = obj as BasicArrayResourceItem
+    return elements == other.elements
+  }
+}


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
```diff
@@ -0,0 +1,54 @@
...
+  init {
+    assert(elements.isEmpty() || defaultIndex < elements.size)
+  }
+
+  override fun getElementCount() = elements.count()
+
+  override fun getElement(index: Int) = elements[index]
+
+  override fun iterator(): MutableIterator<String> = elements.toMutableList().iterator()
+
```
kevinzheng-ap, line 45: @jrodbx The `iterator()` from interface `java.lang.Iterable` is returning `java.util.Iterator`, so the return type would be MutableIterator after converting to Kotlin. Is there any good way to convert it to immutable Iterator in Kotlin?

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
```diff
@@ -0,0 +1,54 @@
...
+  visibility: ResourceVisibility,
+  private val elements: List<String>,
+  private val defaultIndex: Int
+) : BasicValueResourceItemBase(ResourceType.ARRAY, name, sourceFile, visibility), ArrayResourceValue {
+
+  init {
+    assert(elements.isEmpty() || defaultIndex < elements.size)
+  }
+
+  override fun getElementCount() = elements.count()
```
jrodbx, line 46:
```suggestion
  override fun getElementCount() = elements.size
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Resource item representing an array resource.
+ */
+class BasicArrayResourceItem(
+  name: String,
+  sourceFile: ResourceSourceFile,
+  visibility: ResourceVisibility,
+  private val elements: List<String>,
+  private val defaultIndex: Int
+) : BasicValueResourceItemBase(ResourceType.ARRAY, name, sourceFile, visibility), ArrayResourceValue {
+
```
jrodbx, line 41:
```suggestion
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Ported from: [BasicArrayResourceItem.java](https://cs.android.com/android-studio/platform/tools/base/+/18047faf69512736b8ddb1f6a6785f58d47c893f:resource-repository/main/java/com/android/resources/base/BasicArrayResourceItem.java)
+ *
+ * Resource item representing an array resource.
+ *
+ * @param name the name of the resource
+ * @param sourceFile the source file containing definition of the resource
+ * @param visibility the visibility of the resource
+ * @param elements the elements  or the array
+ * @param defaultIndex the default index for the {@link #getValue()} method
```
jrodbx, line 32:
```suggestion
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
jrodbx: Doesn't look like that's possible, due to Java interop:

https://stackoverflow.com/a/61917100
https://kotlinlang.org/docs/java-interop.html#notation-for-platform-types

jrodbx (approved):

jrodbx (approved):

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicArrayResourceItem.kt
TWiStErRob: Tip: maybe use `= java.util.Collections.unmodifiableList(elements).iterator()`, it still returns `MutableIterator`, but it'll catch if anyone is trying to modify the returned object as soon as it happens, because all the modifications would just go to a black-hole copied collection otherwise, and potential issues would be either hidden or come out in a totally unrelated part of the code.

TWiStErRob: Alternatively this would work:
```kotlin
	@Suppress("RETURN_TYPE_MISMATCH_ON_OVERRIDE") // This is suppressing a compile error https://github.com/JetBrains/kotlin/blob/master/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java#L950
	override fun iterator(): Iterator<String> = elements.iterator()
```
because it'll compile to the right bytecode:
```java
   @NotNull
   public Iterator iterator() {
      return this.elements.iterator();
   }
```
but this makes the object actually mutable from Java.

kevinzheng-ap: > Tip: maybe use `= java.util.Collections.unmodifiableList(elements).iterator()`, it still returns `MutableIterator`, but it'll catch if anyone is trying to modify the returned object as soon as it happens, because all the modifications would just go to a black-hole copied collection otherwise, and potential issues would be either hidden or come out in a totally unrelated part of the code. Additionally it won't create an extra copy array, but I guess this would be negligible anyway for Android Array Resources.

@TWiStErRob this solution looks good, will update the PR

kevinzheng-ap: > Tip: maybe use `= java.util.Collections.unmodifiableList(elements).iterator()`, it still returns `MutableIterator`, but it'll catch if anyone is trying to modify the returned object as soon as it happens, because all the modifications would just go to a black-hole copied collection otherwise, and potential issues would be either hidden or come out in a totally unrelated part of the code. Additionally it won't create an extra copy array, but I guess this would be negligible anyway for Android Array Resources.

@TWiStErRob this solution looks good, will update the PR

PR #878, opened 2023-05-18
Port BasicPluralsResourceItem from tools/base/resource-repository

kevinzheng-ap:

cc @jrodbx

### DIFF
--- a/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicPluralsResourceItem.kt
+++ b/paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicPluralsResourceItem.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicPluralsResourceItem.kt
```diff
@@ -0,0 +1,88 @@
...
+    return arities contentEquals other.arities && values contentEquals other.values
+  }
+
+  companion object {
+    private fun getIndex(arity: Arity?, arities: Collection<Arity>): Int {
+      if (arity == null || arities.isEmpty()) {
+        return 0
+      }
+
+      return arities.indexOfFirst { it == arity }.takeIf { it != -1 } ?: throw IllegalArgumentException()
```
jrodbx, line 85:
```suggestion
      val index = arities.indexOf(arity)
      return if (index != -1) index else throw IllegalArgumentException()
```

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/base/BasicPluralsResourceItem.kt
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  override fun getPluralsCount() = arities.size
+
+  override fun getQuantity(index: Int): String = arities[index].getName()
+
+  override fun getValue(index: Int): String = values[index]
+
+  override fun getValue(quantity: String): String? {
+    return arities.indexOfFirst { quantity == it.getName() }.takeIf { it != -1 }?.let { values[it] }
```
jrodbx, line 65:
```suggestion
    val index = arities.indexOfFirst { it.name == quantity }
    return if (index != -1) values[index] else null
```

jrodbx: Would be nice if the stdlib provided, something like:
```
Array<out T>.indexOfOrNull(element: T): Int?
```

But in the meantime, the proposed feels more readable vs`takeIf`.

jrodbx (approved):

PR #886, opened 2023-05-22
Handle labels when mergeDescendants is true with children setting custom contentDescriptions

adamalyyan:
This addresses an edge case when a parent element is sets `mergeDescendants` to `true` and one or more child elements have a custom `contentDescription` set. Previously, only the children with a custom `contentDescription` will show up on the snapshots label and any children with nothing set will be clobbered (for instance a Text field with no `contentDescription` set. 

This change updates the logic where we process the semantic tree which uses the merged semantic nodes to traverse all the nodes and generate their labels. In the edge case above however, in the merged tree, when a root has `mergeDescendants` set to `true` it also merges all the properties of the children into 1 map including all text fields, contentDescriptions, etc. As a result, when we generate the label, we look at the contentDescription property and see that it has a value and we use that, but then any children that didn't need a custom contentDescription won't appear in the final label. To remedy this, when traversing the merged tree, if we find a node that has `mergeDescendants` set to `true` we will look up the unmerged version of that root and generate the label from the children of that node instead since it has all the properties on each specific child, allowing us to generate the label that TalkBack would generate.

Below is an example from using the ComposeA11yTest file with `mergeDescendants` set to `true` with 2 children and 1 having a custom `contentDescription`

| Before | After |
| ------ | ----- |
| app cash paparazzi sample_ComposeA11yTest_compositeItems | app cash paparazzi sample_ComposeA11yTest_compositeItems


geoff-powell (approved):
LGTM, can you run spotless and also regenerate the golden image for `ComposeA11y` test in paparazzi-gradle-plugin?

sample/src/test/java/app/cash/paparazzi/sample/ComposeA11yTest.kt
```diff
@@ -56,7 +57,13 @@ class ComposeA11yTest {
...
           Column {
             Text("Text")
-            Text("more text")
+            Text(
+              text = "more text",
+              modifier = Modifier.semantics { contentDescription = "custom description" })
+            Row(modifier = Modifier.semantics(mergeDescendants = true) {}) {
+              Text("Merged descendant text")
+              Text(text = "more text", modifier = Modifier.semantics { contentDescription = "custom description" })
+            }
```
geoff-powell, line 66: nit: maybe we can add the same support in the plugin test `composeA11y` project?

geoff-powell: try doing it here
https://github.com/cashapp/paparazzi/blob/master/paparazzi/paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/main/java/app/cash/paparazzi/plugin/test/CompositeComposable.kt#L48-L57

adamalyyan: Sounds good, will do!

adamalyyan:
@geoff-powell How do I run the tests in the paparazzi-gradle-plugin? I can only seem to run tasks for things under :sample

geoff-powell:
> @geoff-powell How do I run the tests in the paparazzi-gradle-plugin? I can only seem to run tasks for things under :sample

you can run `./gradlew -p paparazzi paparazzi-gradle-plugin:test`

geoff-powell (approved):

PR #912, opened 2023-05-31
Fix renovate/com.diffplug.spotless branch

TWiStErRob:
After looking at the error message I think this will make https://github.com/cashapp/paparazzi/pull/889 work. 

Note: this is based on #889, but if the base branch is changed to `master` it can be merged directly, up to maintainers, feel free to change base branch and title to supersede #889 instead of amend it.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,7 +6,7 @@ composeCompiler = "1.4.7"
 javaTarget = "11"
 jcodec = "0.2.5"
 kotlin = "1.8.21"
-ktlint = "0.46.1"
+ktlint = "0.49.1"
 moshi = "1.15.0"
 
 minSdk = "25"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -38,6 +38,8 @@ jcodec-javase = { module = "org.jcodec:jcodec-javase", version.ref = "jcodec" }
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.7.1" }
 
+ktlint = { module = "com.pinterest.ktlint:ktlint-rule-engine", version.ref = "ktlint" }
+
 kxml2 = { module = "kxml2:kxml2", version = "2.3.0" }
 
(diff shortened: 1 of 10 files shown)


gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ composeCompiler = "1.4.7"
 javaTarget = "11"
 jcodec = "0.2.5"
 kotlin = "1.8.21"
-ktlint = "0.46.1"
+ktlint = "0.49.1"
```
TWiStErRob, line 9: [Ktlint 0.49.0](https://github.com/pinterest/ktlint/releases/tag/0.49.0) contains some breaking changes to the API, and Paparazzi is behind on 0.46.1 which is incompatible with latest Spotless.

gradle/libs.versions.toml
```diff
@@ -38,6 +38,8 @@ jcodec-javase = { module = "org.jcodec:jcodec-javase", version.ref = "jcodec" }
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.7.1" }
 
+ktlint = { module = "com.pinterest.ktlint:ktlint-rule-engine", version.ref = "ktlint" }
```
TWiStErRob, line 41: This is to help Renovate understand the "ktlint" version so it can upgrade it from mavenCentral.

Another, potentially better, long term solution is to remove the `libs.versions.ktlint` constant fully and rely on the default built-in version in Spotless. It's pretty hard to get incompatible if there's no override. _This might become a non-issue, because it looks like ktlint is preparing for a 1.0 release (as far as I understand they pre-broke users, so 1.0 doesn't break users 🙄)._

jrodbx (approved):
thank you!

paparazzi/paparazzi/src/main/java/app/cash/paparazzi/internal/SessionParamsBuilder.kt
```diff
@@ -103,7 +103,7 @@ internal data class SessionParamsBuilder(
     )
 
     val result = SessionParams(
-      layoutPullParser, renderingMode, projectKey /* for caching */,
+      layoutPullParser, renderingMode, projectKey,
```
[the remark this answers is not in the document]
TWiStErRob: Considering there are probably hundreds of other function calls not having named arguments due to the same reason, it's probably fine to remove these two. I didn't look too deep as to why it was failing though, an error for this sounds a bit harsh on the ktlint side.

PR #928, opened 2023-06-08
Ignore xliff in XmlTextExtractor

kevinzheng-ap:
Ignore `xliff` in XmlTextExtractor to keep original parameter like `(0)` rather than being transformed to example like `{9}`

cc @jrodbx

### DIFF
--- a/paparazzi/src/test/resources/folders/res/values/strings.xml
+++ b/paparazzi/src/test/resources/folders/res/values/strings.xml
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
   <plurals name="plural_name">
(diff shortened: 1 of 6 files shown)


paparazzi/src/test/resources/folders/res/values/strings.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
   <plurals name="plural_name">
     <item quantity="zero">Nothing</item>
     <item quantity="one">One String</item>
   </plurals>
   <string name="string_name">Test String</string>
+  <string name="string_name_xliff">Test String <xliff:g id="number" example="9">{0}</xliff:g> with suffix</string>
```
jrodbx, line 8: nice!

jrodbx (approved):

PR #929, opened 2023-06-08
Convert task actions to lambdas; supported as of Gradle 7.6

jrodbx:
https://github.com/gradle/gradle/pull/21054


sample-library-depth-a1/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'com.android.library'
```
TWiStErRob, line 1: Did you mean to commit these samples here?

geoff-powell: yep, looks like this was some other work that snuck in.

geoff-powell (approved):
LGTM, once the sample library stuff is removed.

PR #937, opened 2023-06-09
Enable ignored tests in AarSourceResourceRepositoryTest

kevinzheng-ap:
Enable ignored tests in AarSourceResourceRepositoryTest


paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
```diff
@@ -69,8 +69,6 @@ internal class Renderer(
             }.apply { loadResources() }
           )
       } else {
-        println("New resource loading coming soon")
```
jrodbx, line 72: This should fail one of the tests....why isn't it?  Either way, let's revert this change as it's unrelated to this PR?

jrodbx (approved):

PR #939, opened 2023-06-09
Fix relative path for module

kevinzheng-ap:
In [RepositoryLoader](https://github.com/cashapp/paparazzi/blob/7e7a8486057610a00b4fba664b5aa23dce5f06f7/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt#L937), today it looks for `res` segment and get sub path if the resource file is relative path. In our resources.txt, the module path is this pattern `../module/build/intermediates/packaged_res/debug`. So current assertion fails as no exact `res` is found. 

The solution is to make locale resource paths absolute to allow it be relativized properly. I have added `sample module` in our test and confirmed it is working. 

cc @jrodbx

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -89,6 +89,7 @@ class PaparazziPlugin : Plugin<Project> {
       val mergeAssetsProvider =
         project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
       val mergeAssetsOutputDir = mergeAssetsProvider.flatMap { it.outputDir }
+      val projectDirectory = project.layout.projectDirectory
       val buildDirectory = project.layout.buildDirectory
       val gradleUserHomeDir = project.gradle.gradleUserHomeDir
       val reportOutputDir = buildDirectory.dir("reports/paparazzi")
@@ -175,6 +176,7 @@ class PaparazziPlugin : Plugin<Project> {
       val testTaskProvider = project.tasks.named("test$testVariantSlug", Test::class.java) { test ->
         test.systemProperties["paparazzi.test.resources"] =
           writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path
+        test.systemProperties["paparazzi.project.dir"] = projectDirectory.toString()
         test.systemProperties["paparazzi.build.dir"] = buildDirectory.get().toString()
         test.systemProperties["paparazzi.artifacts.cache.dir"] = gradleUserHomeDir.path
         test.systemProperties["kotlinx.coroutines.main.delay"] = true
--- a/paparazzi/build.gradle
+++ b/paparazzi/build.gradle
@@ -118,6 +118,10 @@ tasks.withType(Test).configureEach {
     generateTestConfig.map { it.outputs.files.singleFile }.get().path
   )
   systemProperty(
+    "paparazzi.project.dir",
+    project.layout.projectDirectory.toString()
+  )
+  systemProperty(
     "paparazzi.build.dir",
     project.layout.buildDirectory.get().toString()
   )
--- a/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
@@ -55,6 +55,7 @@ fun detectEnvironment(): Environment {
   val resourcesFile = File(System.getProperty("paparazzi.test.resources"))
   val configLines = resourcesFile.readLines()
 
+  val projectDir = Paths.get(System.getProperty("paparazzi.project.dir"))
   val appTestDir = Paths.get(System.getProperty("paparazzi.build.dir"))
   val artifactsCacheDir = Paths.get(System.getProperty("paparazzi.artifacts.cache.dir"))
   val androidHome = Paths.get(androidHome())
@@ -66,7 +67,7 @@ fun detectEnvironment(): Environment {
     packageName = configLines[0],
     compileSdkVersion = configLines[2].toInt(),
     resourcePackageNames = configLines[5].split(","),
-    localResourceDirs = configLines[6].split(","),
+    localResourceDirs = configLines[6].split(",").map { projectDir.resolve(it).toString() },
     libraryResourceDirs = configLines[7].split(",").map { artifactsCacheDir.resolve(it).toString() }
   )
 }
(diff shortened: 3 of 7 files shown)


settings.gradle
```diff
@@ -10,6 +10,7 @@ include ':paparazzi-agent'
 include ':paparazzi-gradle-plugin'
 
 include ':sample'
+include ':sample-module'
```
jrodbx, line 13: i prefer not to add a new sample module for this, especially with the discussed plan to overhaul the sample soon.  can we revert that part and maybe also update the existing plugin test?

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt
```diff
@@ -927,15 +927,12 @@ abstract class RepositoryLoader<T : LoadableResourceRepository>(
   }
 
   protected fun getResRelativePath(file: PathString): String {
-    if (file.isAbsolute) {
+    if (file.isAbsolute || !isLoadingFromZipArchive) {
```
jrodbx, line 930:
```suggestion
    if (file.isAbsolute) {
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt
```diff
@@ -927,15 +927,12 @@ abstract class RepositoryLoader<T : LoadableResourceRepository>(
...
       return resourceDirectoryOrFilePath.relativize(file).portablePath
     }
     assert(file.nameCount != 0)
```
jrodbx, line 933: since we're reverting back to the Android Studio way, let's include this comment.  for reference: https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:resource-repository/main/java/com/android/resources/base/RepositoryLoader.java;l=125?q=RepositoryLoader&sq=

```suggestion
    // The path is already relative, drop the first "res" segment.
    assert(file.nameCount != 0)
```

jrodbx (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt
kevinzheng-ap: `isLoadingFromZipArchive` is needed here otherwise `../module/build/intermediates/packaged_res/debug/` cannot be handled as there is no `res` segment in the path. It is tricky one. Let us discuss in next sync meeting. Or do you have better suggestion?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -177,6 +177,7 @@ class PaparazziPlugin : Plugin<Project> {
           writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path
         test.systemProperties["paparazzi.build.dir"] = buildDirectory.get().toString()
         test.systemProperties["paparazzi.artifacts.cache.dir"] = gradleUserHomeDir.path
+        test.systemProperties["paparazzi.project.dir"] = project.projectDir.path
```
jrodbx, line 180: nit: move this up two lines, above "build.dir"

paparazzi/build.gradle
```diff
@@ -125,6 +125,10 @@ tasks.withType(Test).configureEach {
     "paparazzi.artifacts.cache.dir",
     project.gradle.gradleUserHomeDir.path
   )
+  systemProperty(
+    "paparazzi.project.dir",
+    project.projectDir.path
+  )
```
jrodbx, line 131: ditto

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -57,6 +57,7 @@ fun detectEnvironment(): Environment {
 
   val appTestDir = Paths.get(System.getProperty("paparazzi.build.dir"))
   val artifactsCacheDir = Paths.get(System.getProperty("paparazzi.artifacts.cache.dir"))
+  val projectDir = System.getProperty("paparazzi.project.dir")
```
jrodbx, line 60: ditto

jrodbx, line 60:
```suggestion
  val projectDir = Paths.get(System.getProperty("paparazzi.project.dir"))
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -66,7 +67,7 @@ fun detectEnvironment(): Environment {
     packageName = configLines[0],
     compileSdkVersion = configLines[2].toInt(),
     resourcePackageNames = configLines[5].split(","),
-    localResourceDirs = configLines[6].split(","),
+    localResourceDirs = configLines[6].split(",").map { "$projectDir/$it" },
```
jrodbx, line 70:
```suggestion
    localResourceDirs = configLines[6].split(",").map { projectDir.resolve(it).toString() },
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt
```diff
@@ -930,12 +930,10 @@ abstract class RepositoryLoader<T : LoadableResourceRepository>(
     if (file.isAbsolute) {
       return resourceDirectoryOrFilePath.relativize(file).portablePath
     }
+    // The path is already relative, drop the first "res" segment.
```
jrodbx, line 933:
```suggestion

    // The path is already relative, drop the first "res" segment.
    assert(file.getNameCount() != 0)
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -175,6 +175,7 @@ class PaparazziPlugin : Plugin<Project> {
       val testTaskProvider = project.tasks.named("test$testVariantSlug", Test::class.java) { test ->
         test.systemProperties["paparazzi.test.resources"] =
           writeResourcesTask.flatMap { it.paparazziResources.asFile }.get().path
+        test.systemProperties["paparazzi.project.dir"] = project.projectDir.path
```
jrodbx, line 178: can we follow the pattern for buildDirectory and do something like this?

```suggestion
        test.systemProperties["paparazzi.project.dir"] = projectDirectory.toString()
```

This means extracting `val projectDirectory = project.layout.projectDirectory` from `snapshotOutputDir` above.

paparazzi/build.gradle
```diff
@@ -117,6 +117,10 @@ tasks.withType(Test).configureEach {
     "paparazzi.test.resources",
     generateTestConfig.map { it.outputs.files.singleFile }.get().path
   )
+  systemProperty(
+    "paparazzi.project.dir",
+    project.projectDir.path
```
jrodbx, line 122:
```suggestion
    project.layout.projectDirectory.toString()
```

jrodbx (approved):
almost there, thanks for sticking with this!

PR #942, opened 2023-06-12
Split local resources to local and module in resources.txt

kevinzheng-ap:
Split local resources to `local` and `module` in resources.txt to make sure correct orders are used in repository later

cc @jrodbx

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -741,9 +741,10 @@ class PaparazziPluginTest {
     assertThat(resourceFileContents[0]).isEqualTo("app.cash.paparazzi.plugin.test")
     assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
     assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
-    assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary")
+    assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary,app.cash.paparazzi.plugin.test.module1,app.cash.paparazzi.plugin.test.module2")
     assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
-    assertThat(resourceFileContents[7]).matches("^caches/transforms-3/[0-9a-f]{32}/transformed/external/res\$")
+    assertThat(resourceFileContents[7]).isEqualTo("module1/build/intermediates/packaged_res/debug,module2/build/intermediates/packaged_res/debug")
+    assertThat(resourceFileContents[8]).matches("^caches/transforms-3/[0-9a-f]{32}/transformed/external/res\$")
   }
 
   @Test
@@ -763,9 +764,10 @@ class PaparazziPluginTest {
     assertThat(resourceFileContents[0]).isEqualTo("app.cash.paparazzi.plugin.test")
     assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
     assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
-    assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary")
+    assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary,app.cash.paparazzi.plugin.test.module1,app.cash.paparazzi.plugin.test.module2")
     assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
-    assertThat(resourceFileContents[7]).matches("^caches/transforms-3/[0-9a-f]{32}/transformed/external/res\$")
+    assertThat(resourceFileContents[7]).isEqualTo("module1/build/intermediates/packaged_res/debug,module2/build/intermediates/packaged_res/debug")
+    assertThat(resourceFileContents[8]).matches("^caches/transforms-3/[0-9a-f]{32}/transformed/external/res\$")
   }
 
   @Test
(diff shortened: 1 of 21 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/resources/ProjectResourceRepository.kt
```diff
@@ -19,20 +19,39 @@ internal class ProjectResourceRepository private constructor(
...
+    ): List<LocalResourceRepository> {
+      val main: LocalResourceRepository = getModuleResources(resourceDirectories)
+
+      val resources: MutableList<LocalResourceRepository> = ArrayList(moduleResourceDirectories.size + 1)
+      resources += main
+      for (moduleResourceDirectory in moduleResourceDirectories) {
+        resources += getModuleResources(listOf(moduleResourceDirectory))
+      }
+
+      return resources
```
jrodbx, line 43:
```suggestion
      val resources = buildList< LocalResourceRepository>(moduleResourceDirectories.size + 1) {
        this += main
        for (moduleResourceDirectory in moduleResourceDirectories) {
          this += getModuleResources(listOf(moduleResourceDirectory))
        }
      }

      return resources
```

jrodbx: can the for loop be simplified to:

```
this += getModuleResources(moduleResourceDirectories)
```
?

kevinzheng-ap: > can the for loop be simplified to:
> 
> ```
> this += getModuleResources(moduleResourceDirectories)
> ```
> 
> ?

As discussed, `getModuleResources(moduleResourceDirectories)` will reverse the order again. Each `moduleResourceDirectory` should be in its own list and keep original order

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/ProjectResourceRepository.kt
```diff
@@ -19,20 +19,39 @@ internal class ProjectResourceRepository private constructor(
...
-        )
+        localResources = computeRepositories(resourceDirectories, moduleResourceDirectories)
       )
     }
 
+    private fun computeRepositories(
+      resourceDirectories: List<File>,
+      moduleResourceDirectories: List<File>
+    ): List<LocalResourceRepository> {
+      val main: LocalResourceRepository = getModuleResources(resourceDirectories)
```
jrodbx, line 35:
```suggestion
      val main = getModuleResources(resourceDirectories)
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -765,7 +765,27 @@ class PaparazziPluginTest {
...
+    val resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    assertThat(resourcesFile.exists()).isTrue()
+
+    val resourceFileContents = resourcesFile.readLines()
+    assertThat(resourceFileContents[0]).isEqualTo("app.cash.paparazzi.plugin.test")
+    assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
+    assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
+    assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test")
+    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
+    assertThat(resourceFileContents[7]).isEqualTo("module1/build/intermediates/packaged_res/debug,module2/build/intermediates/packaged_res/debug")
```
jrodbx, line 788: can we merge this test into the original, instead of having two separate tests testing different line number in resources.txt?

paparazzi-gradle-plugin/src/test/projects/verify-resources-modules/module2/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
jrodbx, line 2: i'm confused; shouldn't there be at least one string resource defined in each of these modules?

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -68,7 +69,8 @@ fun detectEnvironment(): Environment {
     compileSdkVersion = configLines[2].toInt(),
     resourcePackageNames = configLines[5].split(","),
     localResourceDirs = configLines[6].split(",").map { projectDir.resolve(it).toString() },
-    libraryResourceDirs = configLines[7].split(",").map { artifactsCacheDir.resolve(it).toString() }
+    moduleResourceDirs = configLines[7].split(",").filter { it.isNotBlank() }.map { projectDir.resolve(it).toString() },
```
jrodbx, line 72:
```suggestion
    moduleResourceDirs = configLines[7].split(",").filter { it.isNotBlank() }.map { projectDir.resolve(it).toString() },
```

why is the `filter { it.isNotBlank() }` needed?  can we omit?

jrodbx (approved):
approving, with comments

paparazzi-gradle-plugin/src/test/projects/verify-resources-modules/module2/src/main/res/values/strings.xml
kevinzheng-ap: As discussed, we just need dependency of the module and test the order, it does not matter if the module has actual resource or not

kevinzheng-ap: As discussed, we just need dependency of the module and test the order, it does not matter if the module has actual resource or not

kevinzheng-ap: As discussed, we just need dependency of the module and test the order, it does not matter if the module has actual resource or not

jrodbx (approved):

PR #945, opened 2023-06-13
Fix Arity name comparison in BasicPluralsResourceItem

kevinzheng-ap:
Fix the bug that in BasicPluralsResourceItem, the `name` field in enum Arity should be used to match quantity rather than the name of enum constant.

cc @jrodbx


jrodbx (approved):
Oh wow, nice catch!

Issue #191, opened 2021-02-03
Enable code completion for test projects


saket:
Adding new test projects to `/paparazzi-gradle-plugin/src/test` is slightly painful right now because code completion doesn't work. Is this something that can be enabled?

TWiStErRob:
I don't think that's possible, because build.gradle will only have completion if you synced that project. There's an implicit pluginClasspath hidden by the Gradle test kit, so IDEA has no chance of discovering where to get the dependencies from.

PR #959, opened 2023-06-27
Add test project modules to the root settings.gradle

edenman:
This allows for syntax highlight and running tests in the IDE.


settings.gradle
```diff
@@ -18,3 +18,19 @@ includeBuild('build-logic') {
...
+
+def brokenProjects = [
+  "appcompat-missing",
+  "invalid-application-plugin",
+  "missing-library-plugin",
+  "multiplatform-plugin-without-android",
+  "night-mode-xml", // TODO fix this one or remove it? Maybe left out of PaparazziPluginTest by accident?
+]
+
+new File(rootDir, "paparazzi-gradle-plugin/src/test/projects").eachDir {
```
TWiStErRob, line 30: What an idea! Nice.

I think we would want to do this conditionally, because doing `gradlew build` or `gradlew test` locally to quickly check if Paparazzi is functioning would `build`/`test` these projects as well, which is double work since the `:p-g-p:test` will run them anyway.

If we wrap this whole added block in `val isIdea = (System.getProperty("idea.version") ?: "").isNotEmpty()` (or equivalent), they would only be included in AS/IDEA and not in command line/CI.

What do you think?

edenman: love it.  @jrodbx wdyt?

jrodbx: looking at the CI logs, it looks like now all these test projects would be subject to the check task running on them, which now does added work....

so, not sure if doubling down on this, if extra script hacking, to improve IDE syntax highlighting is the right tradeoff, here.

re: running tests in the IDE, def want this, but I suspect that can be accomplished a different way.

@JakeWharton thoughts?

edenman: Presumably if we did @TWiStErRob's suggestion, those tests would no longer be run as part of ./gradlew check (since you're not including the projects unless you're in the IDE)

JakeWharton: I don't know anything about these nested builds for this project. In the projects I work on I rarely have to change anything in the nested builds. I would probably look into pursuing the ability to use `includeBuild` from the nested build so that you could open that project in the IDE and it would pull in the enclosing Paparazzi build with it. That way you could work within a single nested project at a time so that it models whatever scenario you want and then switch back to just the regular build when you're done.

edenman: I guess each of them would need a settings.gradle to get that to work?  Would still be annoying to switch between projects, especially in the case where there's additional steps to the _actual_ test (e.g. `PaparazziPluginTest.rerunOnAssetChange`)

Is it really that big of a deal to add the IDE-only inclusion of these modules into the main one?

JakeWharton: ¯\\\_(ツ)_/¯ Not my project so no idea. Was just asked for an opinion.

edenman: yeah sorry, the second question was aimed at @jrodbx

jrodbx: Ok, I like this much better.  Still a little annoying that the IDE adds the project name redundantly next to the module name, but I can live with it.

Screen Shot 2023-07-18 at 5 15 21 PM

settings.gradle
```diff
@@ -18,3 +18,27 @@ includeBuild('build-logic') {
...
+
+def isIdea = !(System.getProperty("idea.version") ?: "").isEmpty()
+if (isIdea) {
+  includeTestProjectsIntoIDE()
+}
+
+private void includeTestProjectsIntoIDE() {
+  // Some of the test projects are intentionally supposed to fail. Don't import these into the main project.
+  def brokenProjects = [
+    "appcompat-missing", // intentionally does not import
```
jrodbx, line 30:
```suggestion
```

this one appears to work for me, can you double check?

settings.gradle
```diff
@@ -18,3 +18,27 @@ includeBuild('build-logic') {
...
+    "invalid-application-plugin", // intentionally does not import
+    "missing-library-plugin", // intentionally does not import
+    "multiplatform-plugin-without-android", // intentionally does not import
+    "night-mode-xml", // TODO fix this one or remove it? Maybe left out of PaparazziPluginTest by accident?
```
jrodbx, line 34: oops, fixed in https://github.com/cashapp/paparazzi/pull/1009.  thanks!

jrodbx (approved):
mind rebasing?  we added a few more tests due to resource/asset loading rewrite, so you might have to update the denylist some more.

thanks!

jrodbx:
Also, this technically resolves https://github.com/cashapp/paparazzi/issues/191.  cc: @saket

edenman:
Ok should be good to go now, thx team!

settings.gradle
```diff
@@ -18,3 +18,28 @@ includeBuild('build-logic') {
...
+
+private void includeTestProjectsIntoIDE() {
+  // Some of the test projects are intentionally supposed to fail. Don't import these into the main project.
+  def brokenProjects = [
+    "invalid-application-plugin", // intentionally does not import
+    "missing-library-plugin", // intentionally does not import
+    "missing-supported-plugins", // intentionally does not import
+    "multiplatform-plugin-without-android", // intentionally does not import
+    "verify-resources-java", // has nested modules so does not import correctly
+    "verify-resources-kotlin", // has nested modules so does not import correctly
```
jrodbx, line 35:
```suggestion
```

https://github.com/cashapp/paparazzi/pull/1010

jrodbx (approved):

Issue #769, opened 2023-05-05
Render mode SHRINK does not work correctly with AlertDialog composable


Th0MmyS:
Preview in AS:
<img width="283" alt="image" src="https://user-images.githubusercontent.com/11424077/236393958-52782807-ff78-4f39-8b71-f9d6d657dcb3.png">

Code:
```
@Composable
fun ComposeDialog(
	modifier: Modifier = Modifier,
	title: String,
	subtitle: String,
	openDialogState: MutableState<Boolean> = remember { mutableStateOf(false) },
	isCancelable: Boolean = true,
	onDismissRequest: () -> Unit = {
		if (isCancelable) {
			openDialogState.value = false
		}
	},
) {
	AlertDialog(
		modifier = modifier
			.fillMaxWidth()
			.clip(RoundedCornerShape(12.dp)),
		onDismissRequest = onDismissRequest,
		title = {
			Text(
				modifier = Modifier
					.fillMaxWidth(),
				textAlign = TextAlign.Center,
				text = title
			)
		},
		text = {
			Text(
				modifier = Modifier
					.fillMaxWidth(),
				textAlign = TextAlign.Center,
				text = subtitle
			)
		},
		buttons = {
			Column(
				modifier = Modifier
					.fillMaxWidth()
					.padding(
						start = 16.dp,
						end = 16.dp,
						bottom = 16.dp
					)
			) {
				Button(
					modifier = Modifier.fillMaxWidth(),
					onClick = {},
					) {
					Text("test")
				}
				Spacer(modifier = Modifier.height(16.dp))
				Button(
					modifier = Modifier.fillMaxWidth(),
					onClick = {},
				) {
					Text("test")
				}
			}
		}
	)
}

@Preview
@Composable
fun ComposeDialogPreview() {
	ComposeDialog(
		title = "Title",
		subtitle = "Subtitle"
	)
}
```

Paparazzi config:
```
@get:Rule
	val paparazzi = Paparazzi(
		maxPercentDifference = 1.0,
		deviceConfig = DeviceConfig.PIXEL_5.copy(softButtons = false),
		renderingMode = SessionParams.RenderingMode.SHRINK
	)
```

Stacktrace:

```java.lang.IllegalArgumentException: Width (0) and height (0) cannot be <= 0
	at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(DirectColorModel.java:1016)
[...]

PR #969, opened 2023-06-30
Fixes compose alert dialogs not rendering when using RenderingMode.SHRINK (#769)

adamalyyan:
This address an issue with compose layouts that include an `AlertDialog` when using `RenderingMode.SHRINK` which was caused by the layout being measured not including the `AlertDialog`'s layout and as a result, it would be ignored and not rendered in the final snapshot.

The issue is that that `AlertDialog` layout is added to the Window which is usually included as part of the layout being measured. However `WindowManagerImpl` has special logic for compose layouts where it will try to find the composable root of the layout to use as the base root view and then it adds itself to that view as a child. This logic specifically looks for a root that ends with `ComposeViewAdapter` and since our root view in Paparazzi is a `FrameLayout`, `WindowManagerImpl` would never find a composable root and always return null as the base root view. This meant that the layout never included the dialog's content so when it goes to measure the layout, it would return the size of the layout excluding the dialog content.

To address this, instead of having a FrameLayout as the root when we are measuring a compose layout, we instead use a FrameLayout subclass that simply ends with `ComposeViewAdapter` which allows the `WindowManagerImpl` to successfully find the composable root and add the layout. When there is no compose layout in the snapshot, we simply use the default FrameLayout which matches how Android Studio renders previews.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -109,7 +109,8 @@ class Paparazzi @JvmOverloads constructor(
 
   private val contentRoot = """
         |<?xml version="1.0" encoding="utf-8"?>
-        |<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+        |<${if (hasComposeRuntime) "app.cash.paparazzi.internal.ComposeViewAdapter" else "FrameLayout"}
+        |     xmlns:android="http://schemas.android.com/apk/res/android"
         |              android:layout_width="${if (renderingMode.horizAction == RenderingMode.SizeAction.SHRINK) "wrap_content" else "match_parent"}"
         |              android:layout_height="${if (renderingMode.vertAction == RenderingMode.SizeAction.SHRINK) "wrap_content" else "match_parent"}"/>
   """.trimMargin()
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
@@ -1,0 +1,16 @@
+package app.cash.paparazzi.internal
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.FrameLayout
+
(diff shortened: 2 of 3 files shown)


adamalyyan:
Hitting an error on "windows-latest, 17": `java.io.FileNotFoundException: Missing platform version 31. Install with sdkmanager --install "platforms;android-31"`

I was hitting this locally, but was able to download SDK 31 to solve it. Was hitting this on master too!

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziComposeViewAdapter.kt
```diff
@@ -0,0 +1,14 @@
+package app.cash.paparazzi.internal
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.FrameLayout
+
+/**
+ * A wrapper layout for compose based layouts which allows [android.view.WindowManagerImpl] to find
```
jrodbx, line 8:
```suggestion
 * A wrapper layout for compose-based layouts which allows [android.view.WindowManagerImpl] to find
```

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziComposeViewAdapter.kt
```diff
@@ -0,0 +1,14 @@
...
+ * a composable root
+ */
+class PaparazziComposeViewAdapter(
```
jrodbx, line 11:
```suggestion
class ComposeViewAdapter(
```

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziComposeViewAdapter.kt
jrodbx, line 8:
```suggestion
/**
 * Ported from: [can we add a link to the analogous file in AOSP here?]
 *
 * A wrapper layout for compose based layouts which allows [android.view.WindowManagerImpl] to find
```

jrodbx (approved):
sweet and simple.  thanks!

jrodbx (approved):

PR #977, opened 2023-07-03
Ensure UTF-8 encoding on SAXParser in MultiResourceFile

### LINKED ISSUE #975: SAXParser fails to parse merged resources file when an emoji in strings resources is treated incorrectly
**Description**
Paparazzi fails to parse merged resources file with emoji. `SAXParser` fails silently in [`MultiResourceFile.java`](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/deprecated/com/android/ide/common/resources/deprecated/MultiResourceFile.java#L166-L175). It fails to load remaining content, leading to styles and other items of `values.xml` not being parsed, leading to consequent failures like an absence of themes in resource repository.

Core failure:
```
Invalid byte 2 of 4-byte UTF-8 sequence.
```
String, causing the issue:
```xml
<string name="components_car_owner_instant_discount_title">"Share policy docs with the owner and get 10% off your policy 🎉"
</string>
```

Worth noticing, we tried to fix it by converting our emojis to respective HTML codes. Yet, it looks like when aapt merges resources in the intermediates, it converts emojis back to UTF-8, leading to the same issue.

**Steps to Reproduce**
1. Add emoji to string resources.
2. Emoji octets should be split with the buffer (2048 bytes) and appear as the buffer's first character.

**Expected behavior**
Paparazzi loads resources without failure.

**Additional information:**
- Paparazzi Version: 1.3.0
- OS: Linux, MacOS
- Compile SDK: 33
- Gradle Version: 8.1.1
- Android Gradle Plugin Version: 8.0.2

Merged file for reference: [values.xml.zip](https://github.com/cashapp/paparazzi/files/11935412/values.xml.zip)

dniHze:
Resolves #975

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/deprecated/com/android/ide/common/resources/deprecated/MultiResourceFile.java
+++ b/paparazzi/src/main/java/app/cash/paparazzi/deprecated/com/android/ide/common/resources/deprecated/MultiResourceFile.java
@@ -19,7 +19,6 @@ package app.cash.paparazzi.deprecated.com.android.ide.common.resources.deprecate
 import app.cash.paparazzi.deprecated.com.android.ide.common.resources.deprecated.ValueResourceParser.IValueResourceRepository;
 import app.cash.paparazzi.deprecated.com.android.io.IAbstractFile;
 import app.cash.paparazzi.deprecated.com.android.io.StreamException;
-import kotlin.text.Charsets;
 
 import com.android.ide.common.rendering.api.ResourceValue;
 import com.android.ide.common.rendering.api.ResourceValueImpl;
@@ -27,7 +26,7 @@ import com.android.ide.common.resources.ResourceValueMap;
 import com.android.resources.ResourceType;
 import com.android.utils.XmlUtils;
 import java.io.IOException;
-import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumMap;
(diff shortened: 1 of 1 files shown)


paparazzi/src/main/java/app/cash/paparazzi/deprecated/com/android/ide/common/resources/deprecated/MultiResourceFile.java
```diff
@@ -166,7 +171,10 @@ private void updateResourceItems(ScanningContext context) {
     private void parseFile() {
         try {
             SAXParser parser = XmlUtils.createSaxParser(sParserFactory);
-            parser.parse(getFile().getContents(), new ValueResourceParser(this, isFramework(), null));
+            InputStreamReader reader = new InputStreamReader(getFile().getContents());
+            InputSource source = new InputSource(reader);
```
TWiStErRob, line 175: [InputSource.setEncoding](https://docs.oracle.com/javase/8/docs/api/org/xml/sax/InputSource.html#setEncoding-java.lang.String-):

> This method has no effect when the application provides a character stream.

A `Reader` is a "character stream", so I think this (and previous line) can be abbreviated to:

```suggestion
            InputSource source = new InputSource(getFile().getContents());
```
and this also makes the encoding change (next line) take effect according to docs.

paparazzi/src/main/java/app/cash/paparazzi/deprecated/com/android/ide/common/resources/deprecated/MultiResourceFile.java
```diff
@@ -19,19 +19,24 @@
 import app.cash.paparazzi.deprecated.com.android.ide.common.resources.deprecated.ValueResourceParser.IValueResourceRepository;
 import app.cash.paparazzi.deprecated.com.android.io.IAbstractFile;
 import app.cash.paparazzi.deprecated.com.android.io.StreamException;
+import kotlin.text.Charsets;
```
TWiStErRob, line 22: This is a pure Java file, I would keep it that way:
```suggestion
import java.nio.charset.StandardCharsets;
```

TWiStErRob (commented):
I noted a few simplifications/gotchas, hope it helps.

dniHze:
> I noted a few simplifications/gotchas, hope it helps.

Thanks for the suggestions!

jrodbx (approved):
thanks!

PR #980, opened 2023-07-06
Migrate to new gradle plugin api in PaparazziPlugin

kevinzheng-ap:
Migrate to new gradle plugin api in `PaparazziPlugin` based on suggestion [here](https://github.com/cashapp/paparazzi/pull/960#discussion_r1243830233)

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -17,10 +17,12 @@ package app.cash.paparazzi.gradle
 
 import app.cash.paparazzi.gradle.utils.artifactViewFor
 import app.cash.paparazzi.gradle.utils.artifactsFor
+import com.android.build.api.artifact.SingleArtifact
+import com.android.build.api.artifact.impl.ArtifactsImpl
+import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.gradle.BaseExtension
-import com.android.build.gradle.LibraryExtension
 import com.android.build.gradle.internal.publishing.AndroidArtifacts.ArtifactType
-import com.android.build.gradle.tasks.MergeSourceSetFolders
+import com.android.build.gradle.internal.scope.InternalArtifactType
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
@@ -35,7 +37,6 @@ import org.gradle.api.logging.LogLevel.LIFECYCLE
 import org.gradle.api.plugins.JavaBasePlugin
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.PathSensitivity
-import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.options.Option
 import org.gradle.api.tasks.testing.Test
 import org.gradle.internal.os.OperatingSystem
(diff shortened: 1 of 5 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -79,139 +80,139 @@ class PaparazziPlugin : Plugin<Project> {
...
-      val testVariantSlug = testVariant.name.capitalize(Locale.US)
+    val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
+    androidComponents.onVariants { variant ->
+      project.afterEvaluate {
+        val variantSlug = variant.name.capitalize(Locale.US)
+        val testVariant = variant.unitTest ?: return@afterEvaluate
+
+        val internalArtifacts = variant.artifacts as ArtifactsImpl
+        val mergeResourcesOutputDir = internalArtifacts.get(InternalArtifactType.MERGED_RES)
+        val mergeAssetsOutputDir = variant.artifacts.get(SingleArtifact.ASSETS)
```
kevinzheng-ap, line 91: use `variant.artifacts.get(SingleArtifact.ASSETS)` to get merged assets directory. `InternalArtifactType.LIBRARY_ASSETS` gives the `library_asset` for the module itself, not merged one.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -105,7 +105,7 @@ abstract class PrepareResourcesTask : DefaultTask() {
         it.newLine()
         it.write(resourcePackageNames)
         it.newLine()
-        it.write(projectResourceDirs.joinFiles(projectDirectory))
+        it.write(projectResourceDirs.joinFiles(projectDirectory, reversed = true))
```
kevinzheng-ap, line 108: Need to reverse the project res directory order compared to previous. Previously the list is in [ascending](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/api/BaseVariant.java;drc=5ac687029454dec1e7bd50697dabbc24d5a9943c;l=130) order, but in new api, it is in [descending](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-api/src/main/java/com/android/build/api/variant/SourceDirectories.kt;drc=5ac687029454dec1e7bd50697dabbc24d5a9943c;l=110) order. And we cannot reverse it in `PaparazziPlugin`, it is too early to do it as following error. Any suggestion for better place is appreciated.

```
> Could not create task ':preparePaparazziDebugResources'.
      > Querying the mapped value of list(interface java.util.Collection, org.gradle.api.internal.provider.AbstractCollectionProperty$CollectingSupplier@7576c856) before task ':generateDebugResValues' has completed is not supported
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -742,7 +742,7 @@ class PaparazziPluginTest {
     assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
     assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
     assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary,app.cash.paparazzi.plugin.test.module1,app.cash.paparazzi.plugin.test.module2")
-    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
+    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res,build/generated/res/resValues/debug")
     assertThat(resourceFileContents[7]).isEqualTo("module1/build/intermediates/packaged_res/debug,module2/build/intermediates/packaged_res/debug")
```
kevinzheng-ap, line 746: new api gives generated resource directory even if it is empty

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -742,7 +742,7 @@ class PaparazziPluginTest {
     assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
     assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
     assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary,app.cash.paparazzi.plugin.test.module1,app.cash.paparazzi.plugin.test.module2")
-    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
+    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res,build/generated/res/resValues/debug")
```
kevinzheng-ap, line 745: New api gives generated resource directory even if it is empty

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -79,139 +80,139 @@ class PaparazziPlugin : Plugin<Project> {
...
-        task.moduleResourceDirs.from(moduleResourceDirs)
-        task.aarExplodedDirs.from(aarExplodedDirs)
-        task.paparazziResources.set(buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.txt"))
-      }
-
-      val testVariantSlug = testVariant.name.capitalize(Locale.US)
+    val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
+    androidComponents.onVariants { variant ->
+      project.afterEvaluate {
+        val variantSlug = variant.name.capitalize(Locale.US)
```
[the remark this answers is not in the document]
TWiStErRob: Huh, [interesting...](https://medium.com/androiddevelopers/new-apis-in-the-android-gradle-plugin-f5325742e614#:~:text=The%20Android%20Gradle%20Plugin%20registers%20tasks%20for%20the%20variants%20after%20both%20the%20before%5BComponent%5D%20and%20the%20on%5BComponent%5D%20callbacks%20have%20been%20called.)

Looks like they really want us to use the artifact APIs and not mess about with tasks :)

Looking at where the failure points (`tasks.named()`) are, would it make sense to wrap only those calls in `afterEvaluate { }`? and treat the rest of the `onVariant` body to be strict. (Ideally the AGP API would evolve in such a way that we don't need to reference those tasks any more, right?)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/utils/FileUtils.kt
```diff
@@ -19,9 +19,10 @@ import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.Directory
 import java.io.File
 
-fun ConfigurableFileCollection.joinFiles(directory: Directory) = files.joinToString(",") { file ->
-  directory.relativize(file)
-}
+fun ConfigurableFileCollection.joinFiles(directory: Directory, reversed: Boolean = false) =
+  files.run { if (reversed) reversed() else this }.joinToString(",") { file ->
+    directory.relativize(file)
+  }
```
TWiStErRob, line 25: Adding a boolean flag to functions is [usually a code smell](https://blog.jetbrains.com/idea/2017/09/code-smells-multi-responsibility-methods/) (long read, but awesome series!).

What do you think about
```suggestion
fun Iterable<File>.joinFiles(directory: Directory): String =
  joinToString(",") { file -> directory.relativize(file) }
```
which would change the call-site to
```kotlin
        it.write(projectResourceDirs.reversed().joinFiles(projectDirectory))
        it.newLine()
        it.write(moduleResourceDirs.joinFiles(projectDirectory))
        it.newLine()
        it.write(aarExplodedDirs.joinFiles(gradleUserHomeDirectory))
```
Notice how the other `joinFiles` calls stay the same!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
TWiStErRob: an alternative might be:
```kotlin
val localResourceDirs = project
    .files(variant.sources.res!!.all).elements.map { it.reversed() }
```
or a probably simpler variant, keeping the FileCollection variable type (_I think this is the best one_):
```kotlin
val localResourceDirs = project
    .files(variant.sources.res!!.all.map { it.reversed() })
```
but the above only changes the first layer of lists (which might be enough!), not sure how reversed it needs to be, if a fully reversed list is needed:
```kotlin
val localResourceDirs = project
    .files(variant.sources.res!!.all.map { it.flatten().reversed() })
```

sample/build.gradle
```diff
@@ -20,6 +20,15 @@ android {
...
     kotlinCompilerExtensionVersion libs.versions.composeCompiler.get()
   }
+
+  buildTypes {
+    debug {
+      resValue 'string', 'string_from_build_system', 'Generated string from build system'
+    }
+    release {
+      resValue 'string', 'string_from_build_system', 'Generated string from build system'
+    }
```
TWiStErRob, line 30: Have you considered adding a discriminator to see which buildType it came from?
```suggestion
    debug {
      resValue 'string', 'string_from_build_system', 'Generated string from build system: debug'
    }
    release {
      resValue 'string', 'string_from_build_system', 'Generated string from build system: release'
    }
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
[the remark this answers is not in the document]
TWiStErRob: Example how we might replace the first 3 `tasks.named()` in a future PR and a lot of complexity (again, untested):
```kotlin
        testVariant.sources.resources!!.addGeneratedSourceDirectory(writeResourcesTask) { task ->
          project.objects.directoryProperty().apply {
            fileProvider(task.paparazziResources.map { it.asFile.parentFile })
          }
        }
```
and the last `.named(test...)` might be replaced with something like
```kotlin
android.testOptions.unitTests.all { test: Test ->
    if (test.name != "test${variantSlug}") return@all
    ...
}
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
kevinzheng-ap: never mind, I found a way to reverse the list in PaparazziPlugin, use `map` from `Provider`

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -79,139 +80,139 @@ class PaparazziPlugin : Plugin<Project> {
...
+        val mergeResourcesOutputDir = internalArtifacts.get(InternalArtifactType.MERGED_RES)
+        val mergeAssetsOutputDir = variant.artifacts.get(SingleArtifact.ASSETS)
+        val projectDirectory = project.layout.projectDirectory
+        val buildDirectory = project.layout.buildDirectory
+        val gradleUserHomeDir = project.gradle.gradleUserHomeDir
+        val reportOutputDir = buildDirectory.dir("reports/paparazzi")
+        val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
+
+        val localResourceDirs = project
+          .files(variant.sources.res!!.all.map { it.reversed() })
```
kevinzheng-ap, line 99: Need to reverse the project res directory order compared to previous. Previously the list is in [ascending](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/api/BaseVariant.java;drc=5ac687029454dec1e7bd50697dabbc24d5a9943c;l=130) order, but in new api, it is in [descending](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-api/src/main/java/com/android/build/api/variant/SourceDirectories.kt;drc=5ac687029454dec1e7bd50697dabbc24d5a9943c;l=110) order.

jrodbx:
After speaking with the Android Tools team and seeing how we'd need to resort to more hacks to make this work, I think we should direct our efforts to making the new resource loading system stable and move away from merged res entirely, obviating the need for these types of hacks.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: for clarity, what introduces this exactly?   i'm wondering if this is worth supporting even if this PR itself doesn't land.

kevinzheng-ap:
Close this PR for now and will revisit the migration later

PR #997, opened 2023-07-15
Explicitly declare inputs to the Test task

TWiStErRob:
to make sure they're re-executed whenever the `:paparazzi` code changes.

I kept "rerunning tests" after changing Environment.kt, but my changes were not reflected in the test results. I was baffled, until I realized how the paparazzi.jar gets into the test projects.

### DIFF
--- a/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi-gradle-plugin/build.gradle
@@ -48,7 +48,10 @@ tasks.withType(Test).configureEach {
   // These are referenced with relative paths in src/test/projects/test.settings.gradle.
   // They need to be declared on the Test task to make sure it runs with the new files.
   inputs.file(rootProject.layout.projectDirectory.file("gradle/libs.versions.toml"))
-  inputs.dir(rootProject.layout.buildDirectory.dir("localMaven"))
+  inputs.dir(fileTree(rootProject.layout.buildDirectory.dir("localMaven")) {
+    // Exclude Maven repository files, to prevent cache misses because of changing `<lastUpdated>20230715095719</lastUpdated>`.
+    exclude("**/maven-metadata.xml")
+  })
   dependsOn(':paparazzi:publishMavenPublicationToProjectLocalMavenRepository')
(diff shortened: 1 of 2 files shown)


geoff-powell (approved):
cc @colinmarsch fixes the issue you tried to solve here: https://github.com/cashapp/paparazzi/pull/2038

geoff-powell:
Thanks @TWiStErRob!

paparazzi-gradle-plugin/build.gradle
```diff
@@ -45,6 +45,12 @@ buildConfig {
 }
 
 tasks.withType(Test).configureEach {
+  // These are referenced with relative paths in src/test/projects/test.settings.gradle.
+  // They need to be declared on the Test task to make sure it runs with the new files.
+  inputs.file(rootProject.layout.projectDirectory.file("gradle/libs.versions.toml"))
+    .withPropertyName("libs.versions.toml")
+  inputs.dir(rootProject.layout.buildDirectory.dir("localMaven"))
```
liutikas, line 52: Drive-by comments, this will never be up to date because publication tasks include timestamps that get generated every time maven-metadata.xml

liutikas: You'll likely want something like

```
val repo = rootProject.layout.buildDirectory.dir("localMaven")
inputs.files(
        repo.map {
            // Exclude maven-metadata.xml as they contain timestamps but have no effect on the test outcomes
            it.asFileTree.matching { exclude("**/maven-metadata.xml*") }
        }
    ).withPathSensitivity(PathSensitivity.RELATIVE).withPropertyName("repo")
```

TWiStErRob: @liutikas See the other changed build.gradle file, the publishing task will be up to date, because the input to it didn't change. Publishing task doesn't declare outputs, nor "current time" as input. So it won't rewrite the maven-metadata.xml with the timestamp on every execution when we just simply declare the folder as the output.

TWiStErRob: Hmm, just realized that the other script is declaring overlapping outputs into one folder, wonder why it didn't fail.

PR #917, opened 2023-06-02
Remove null from name on gif screenshots

BraisGabin:
Right now when you run paparazzi with `gif` without setting a name the files that paparazzi create ends with `_null_0.png`. That `"null"` is the `name`  that was not set. This PR aligns the naming of `gif` and `screenshot`. Now, if you don't set a name on `gif` the screenshot name will end with `_0.png`.

A diff is worth a thousand words:
https://github.com/BraisGabin/paparazzi/commit/21bc089b7f594dbd2c0feb47f97ef77794e49e35#diff-7522c35360e75c4ed0d453217b86bfe469ea3cc1b4cabc2597f1cbf9fc6d4173L55-R55

I splitted this PR in two commits. The first one extracts this logic to a different file and add tests to the current behaviour. On the second one I fix the naming issue.

Note: this PR shares the first commit with #902. I'll rebase this (or the other PR) when one of them is merged.

PR #1000, opened 2023-07-16
Handle null snapshot names when recording gifs

Ezard:
Currently, when recording a single snapshot with the name parameter not specified, Paparazzi will handle this properly and will ignore this null name when writing file names

However, when recording gifs, if a name is not specified then Paparazzi will include "null" in the file name of the generated image

e.g. `MyTest_null_0.png`, `MyTest_null_1.png`, etc

This change means that if the name is not specified then these file names will instead end up like `MyTest_0.png`


Ezard:
Didn't spot https://github.com/cashapp/paparazzi/pull/917 - closing this PR in favour of that one

PR #1003, opened 2023-07-18
Add pseudo locale support

kevinzheng-ap:
Add pseudo locale support to allow snapshot with pseudo texts

cc @jrodbx

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
@@ -1,0 +1,152 @@
+package app.cash.paparazzi.internal.resources
+
+import app.cash.paparazzi.internal.resources.base.BasicArrayResourceItem
+import app.cash.paparazzi.internal.resources.base.BasicPluralsResourceItem
+import app.cash.paparazzi.internal.resources.base.BasicTextValueResourceItem
+import app.cash.paparazzi.internal.resources.base.BasicValueResourceItem
+import app.cash.paparazzi.internal.resources.base.BasicValueResourceItemBase
+import com.android.ide.common.rendering.api.ResourceNamespace
+import com.android.ide.common.resources.ResourceValueMap
+import com.android.ide.common.resources.configuration.FolderConfiguration
+import com.android.ide.common.resources.configuration.LocaleQualifier
+import com.android.resources.ResourceType
+import com.google.common.collect.Table
+
+private const val DO_NOT_TRANSLATE = "donottranslate"
+
+private val configCache = hashMapOf<Pair<ResourceSourceFile, LocaleQualifier>, RepositoryConfiguration>()
+
+internal fun Table<ResourceNamespace, ResourceType, ResourceValueMap>.pseudolocalizeIfNeeded(
+  localeQualifier: LocaleQualifier
+): Table<ResourceNamespace, ResourceType, ResourceValueMap> {
+  for (namespace in rowKeySet()) {
+    for (type in columnKeySet()) {
+      val resourceValues = this[namespace, type]!!
+      resourceValues.forEach { (key, value) ->
+        if (value !is BasicValueResourceItemBase || !value.isPseudolocalizable()) {
+          return@forEach
+        }
+
+        val method = when (localeQualifier.value) {
+          "en-rXA" -> Pseudolocalizer.Method.ACCENT
+          "ar-rXB" -> Pseudolocalizer.Method.BIDI
+          else -> return@forEach
+        }
+
+        val pseudoLocaleSourceFile = value.sourceFile.forLocale(localeQualifier)
+        resourceValues[key] = when (value.resourceType) {
+          ResourceType.STRING -> pseudolocalizeString(value as BasicValueResourceItem, pseudoLocaleSourceFile, method)
+          ResourceType.PLURALS -> pseudolocalizePlural(value as BasicPluralsResourceItem, pseudoLocaleSourceFile, method)
+          ResourceType.ARRAY -> pseudolocalizeArray(value as BasicArrayResourceItem, pseudoLocaleSourceFile, method)
+          else -> return@forEach
+        }
+      }
+    }
+  }
+
+  return this
+}
+
+/**
+ * Currently, we cannot respect the "translatable" attribute from string resources as it is not stored in BasicValueResourceItemBase.
+ * As a result, for now, untranslatable strings will be considered pseudolocalizable.
+ * TODO: fix this
+ */
+private fun BasicValueResourceItemBase.isPseudolocalizable(): Boolean =
+  sourceFile.configuration.folderConfiguration.localeQualifier == null && sourceFile.relativePath?.contains(DO_NOT_TRANSLATE) != true
+
+private fun ResourceSourceFile.forLocale(localeQualifier: LocaleQualifier): ResourceSourceFile {
+  val repositoryConfiguration = configCache.getOrPut(this to localeQualifier) {
+    RepositoryConfiguration(
+      repository = configuration.repository,
+      folderConfiguration = FolderConfiguration.copyOf(configuration.folderConfiguration).apply {
+        this.localeQualifier = localeQualifier
+      }
+    )
+  }
+  return when (this) {
+    is ResourceSourceFileImpl -> ResourceSourceFileImpl(relativePath, repositoryConfiguration)
+    is ResourceFile -> ResourceFile(file, repositoryConfiguration)
+    // exhaustive to avoid potential breaking changes in future releases
+    else -> throw IllegalStateException("${javaClass.name} is not handled for new RepositoryConfiguration")
+  }
+}
+
+private fun pseudolocalizeString(
+  original: BasicValueResourceItem,
+  sourceFile: ResourceSourceFile,
+  method: Pseudolocalizer.Method
+): BasicValueResourceItem {
+  val pseudoText = original.value?.pseudoLocalize(method)
+  val pseudoItem = when (original) {
+    is BasicTextValueResourceItem -> {
+      val pseudoRawXml = (original as? BasicTextValueResourceItem)?.rawXmlValue?.pseudoLocalize(method)
+      BasicTextValueResourceItem(
+        type = original.type,
+        name = original.name,
+        sourceFile = sourceFile,
+        visibility = original.visibility,
+        textValue = pseudoText,
+        rawXmlValue = pseudoRawXml
+      )
+    }
+    else -> {
+      BasicValueResourceItem(
+        type = original.type,
+        name = original.name,
+        sourceFile = sourceFile,
+        visibility = original.visibility,
+        value = pseudoText
+      )
+    }
+  }
+  pseudoItem.namespaceResolver = original.namespaceResolver
+  return pseudoItem
+}
+
+private fun pseudolocalizePlural(
+  original: BasicPluralsResourceItem,
+  sourceFile: ResourceSourceFile,
+  method: Pseudolocalizer.Method
+): BasicPluralsResourceItem {
+  val pseudoValues = original.values.map { it.pseudoLocalize(method) }.toTypedArray()
+  val pseudoItem = BasicPluralsResourceItem(
+    name = original.name,
+    sourceFile = sourceFile,
+    visibility = original.visibility,
+    quantityValues = original.arities.mapIndexed { index, arity -> arity to pseudoValues[index] }.toMap(),
+    defaultArity = original.arities[original.defaultIndex]
+  )
+  pseudoItem.namespaceResolver = original.namespaceResolver
+  return pseudoItem
+}
+
+private fun pseudolocalizeArray(
+  original: BasicArrayResourceItem,
+  sourceFile: ResourceSourceFile,
+  method: Pseudolocalizer.Method
+): BasicArrayResourceItem {
+  val pseudoValues = original.toList().map { it.pseudoLocalize(method) }
+  val pseudoItem = BasicArrayResourceItem(
+    name = original.name,
+    sourceFile = sourceFile,
+    visibility = original.visibility,
+    elements = pseudoValues,
+    defaultIndex = original.defaultIndex
+  )
+  pseudoItem.namespaceResolver = original.namespaceResolver
+  return pseudoItem
+}
+
+private fun String.pseudoLocalize(method: Pseudolocalizer.Method) =
+  if (isBlank()) {
+    // For empty string resources like <string name="config_inputEventCompatProcessorOverrideClassName" translatable="false"></string>,
+    // we have to skip pseudolocalization as we are not unable to process the "translatable" attribute for now.
+    // Otherwise the empty string will be incorrectly pseudolocalized to "[]", which is not a valid className in above example
+    // The special handling for empty string is not necessary any more after we fix the "translatable" attribute issue
+    this
+  } else {
+    with(Pseudolocalizer(method)) {
+      start() + text(this@pseudoLocalize) + end()
+    }
+  }
--- a/paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
@@ -1,0 +1,192 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.paparazzi.internal.resources
+
+import com.android.ide.common.rendering.api.ArrayResourceValue
+import com.android.ide.common.rendering.api.PluralsResourceValue
+import com.android.ide.common.rendering.api.ResourceNamespace
+import com.android.ide.common.rendering.api.ResourceValue
+import com.android.ide.common.resources.KnownNamespacesMap
+import com.android.ide.common.resources.ResourceValueMap
+import com.android.ide.common.resources.configuration.LocaleQualifier
+import com.android.resources.ResourceType
+import com.google.common.collect.Table
+import com.google.common.collect.Tables
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import java.util.EnumMap
+
+class PseudolocaleGeneratorTest {
+  private val resFolderRoot = resolveProjectPath("src/test/resources/folders/res")
+
+  private val repository = ResourceFolderRepository(
+    resourceDir = resFolderRoot.toFile(),
+    namespace = ResourceNamespace.TODO()
+  )
+
+  @Test
+  fun accent() {
+    val table = createResourceTable()
+    table.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("en-rXA"))
+
+    // STRING
+    with(table[ResourceNamespace.TODO(), ResourceType.STRING]!!) {
+      assertThat(size).isEqualTo(4)
+      assertThat(getValue("string_name").value).isEqualTo("[Ţéšţ Šţŕîñĝ one two]")
+      assertThat(getValue("string_name_xliff").value).isEqualTo("[Ţéšţ Šţŕîñĝ »{0}« ŵîţĥ šûƒƒîх one two three]")
+      assertThat(getValue("string_name_html").value).isEqualTo("[Ţĥîš Ţéšţ Šţŕîñĝ one two three]")
+      assertThat(getValue("string_name_html").rawXmlValue).isEqualTo("[Ţĥîš <b>Ţéšţ</b> Šţŕîñĝ one two three]")
+      assertThat(getValue("string_not_translate").value).isEqualTo("Not Translatable String")
+    }
+
+    // ARRAY
+    with(table[ResourceNamespace.TODO(), ResourceType.ARRAY]!!) {
+      assertThat(size).isEqualTo(1)
+      with(getValue("string_array_name") as ArrayResourceValue) {
+        assertThat(elementCount).isEqualTo(2)
+        assertThat(getElement(0)).isEqualTo("[Fîŕšţ Ţéšţ Šţŕîñĝ one two three]")
+        assertThat(getElement(1)).isEqualTo("[Šéçöñð Ţéšţ Šţŕîñĝ one two three]")
+      }
+    }
+
+    // PLURALS
+    with(table[ResourceNamespace.TODO(), ResourceType.PLURALS]!!) {
+      assertThat(size).isEqualTo(1)
+      with(getValue("plural_name") as PluralsResourceValue) {
+        assertThat(pluralsCount).isEqualTo(2)
+        assertThat(getValue(0)).isEqualTo("[Ñöţĥîñĝ one two]")
+        assertThat(getValue(1)).isEqualTo("[Öñé Šţŕîñĝ one two]")
+      }
+    }
+  }
+
+  @Test
+  fun bidi() {
+    val table = createResourceTable()
+    table.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("ar-rXB"))
+
+    // STRING
+    with(table[ResourceNamespace.TODO(), ResourceType.STRING]!!) {
+      assertThat(size).isEqualTo(4)
+      assertThat(getValue("string_name").value).isEqualTo("${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      assertThat(getValue("string_name_xliff").value).isEqualTo("${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd $bidiWordStart{0}$bidiWordEnd ${bidiWordStart}with$bidiWordEnd ${bidiWordStart}suffix$bidiWordEnd")
+      assertThat(getValue("string_name_html").value).isEqualTo("${bidiWordStart}This$bidiWordEnd ${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      assertThat(getValue("string_name_html").rawXmlValue).isEqualTo("${bidiWordStart}This$bidiWordEnd $bidiWordStart<b>Test</b>$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      assertThat(getValue("string_not_translate").value).isEqualTo("Not Translatable String")
+    }
+
+    // ARRAY
+    with(table[ResourceNamespace.TODO(), ResourceType.ARRAY]!!) {
+      assertThat(size).isEqualTo(1)
+      with(getValue("string_array_name") as ArrayResourceValue) {
+        assertThat(elementCount).isEqualTo(2)
+        assertThat(getElement(0)).isEqualTo("${bidiWordStart}First$bidiWordEnd ${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+        assertThat(getElement(1)).isEqualTo("${bidiWordStart}Second$bidiWordEnd ${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      }
+    }
+
+    // PLURALS
+    with(table[ResourceNamespace.TODO(), ResourceType.PLURALS]!!) {
+      assertThat(size).isEqualTo(1)
+      with(getValue("plural_name") as PluralsResourceValue) {
+        assertThat(pluralsCount).isEqualTo(2)
+        assertThat(getValue(0)).isEqualTo("${bidiWordStart}Nothing$bidiWordEnd")
+        assertThat(getValue(1)).isEqualTo("${bidiWordStart}One$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      }
+    }
+  }
+
+  @Test
+  fun doesNotPseudolocalizeForOtherLocales() {
+    val result = createResourceTable()
+    result.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("en"))
+
+    // STRING
+    with(result[ResourceNamespace.TODO(), ResourceType.STRING]!!) {
+      assertThat(size).isEqualTo(4)
+      assertThat(getValue("string_name").value).isEqualTo("Test String")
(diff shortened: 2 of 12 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    else -> return this
+  }
+
+  return pseudoItem
+}
+
+private fun isPseudolocalizable(resourceItemBase: BasicValueResourceItemBase): Boolean =
+  resourceItemBase.sourceFile.relativePath?.contains(DO_NOT_TRANSLATE) != true
+
```
kevinzheng-ap, line 91: It is necessary to check if `donottranslate` is in path . One use case is `font name` is saved in [donottranslate_material.xml](https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/values/donottranslate_material.xml)

jrodbx:
Still reviewing, thanks for your patience!

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+
+  @Test
+  fun testAccentPseudolocalization() {
+    val result = createResourceTable()
+    result.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("en-rXA"))
+
+    // STRING
+    with(result[ResourceNamespace.TODO(), ResourceType.STRING]!!) {
+      Truth.assertThat(size).isEqualTo(4)
+      Truth.assertThat(this["string_name"]!!.value).isEqualTo("[Ţéšţ Šţŕîñĝ one two]")
```
jrodbx, line 64: to avoid the repeated !!, can we do something like `this.getValue["string_name"]`?

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+      start() + text(this@pseudoLocalize) + end()
+    }
+  }
+
+fun Table<ResourceNamespace, ResourceType, ResourceValueMap>.pseudolocalizeIfNeeded(
+  localeQualifier: LocaleQualifier
+): Table<ResourceNamespace, ResourceType, ResourceValueMap> {
+  for (namespace in rowKeySet()) {
+    for (type in columnKeySet()) {
+      val map = this[namespace, type]
```
jrodbx, line 33:
```suggestion
      val resourceValues = this[namespace, type]!!
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  return this
+}
+
+private fun BasicValueResourceItemBase.pseudolocalizeIfNeeded(
+  localeQualifier: LocaleQualifier
+): BasicValueResourceItemBase {
+  if (!isPseudolocalizable(this)) {
```
jrodbx, line 47: suggestion: pull this up into the caller and make a private extension for better readability

```
        map.forEach { (key, value) ->
          if(value !is BasicValueResourceItemBase || !value.isPseudolocalizable()) {
            return@forEach
          }
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+
+fun Table<ResourceNamespace, ResourceType, ResourceValueMap>.pseudolocalizeIfNeeded(
+  localeQualifier: LocaleQualifier
+): Table<ResourceNamespace, ResourceType, ResourceValueMap> {
+  for (namespace in rowKeySet()) {
+    for (type in columnKeySet()) {
+      val map = this[namespace, type]
+      map?.forEach { (name, value) ->
+        map[name] =
+          (value as? BasicValueResourceItemBase)?.pseudolocalizeIfNeeded(localeQualifier) ?: value
```
jrodbx, line 36:
```suggestion
          value.pseudolocalizeIfNeeded(localeQualifier)
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+      method
+    )
+
+    else -> return this
+  }
+
+  return pseudoItem
+}
+
+private fun isPseudolocalizable(resourceItemBase: BasicValueResourceItemBase): Boolean =
```
jrodbx, line 89: should this also check for:
1) whether this is the default locale (i.e., only pseudolocalize strings that are in the default locale)
2) whether the string is translatable, i.e., `translatable="false"`?

This would align it more to AAPT's pseudolocalizable check: https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/aaptcompiler/src/main/java/com/android/aaptcompiler/PseudolocaleGenerator.kt;l=217?q=isPseudolo&sq=

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+        map[name] =
+          (value as? BasicValueResourceItemBase)?.pseudolocalizeIfNeeded(localeQualifier) ?: value
+      }
+    }
+  }
+
+  return this
+}
+
+private fun BasicValueResourceItemBase.pseudolocalizeIfNeeded(
```
jrodbx, line 44: nit: inline this method

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.android.ide.common.rendering.api.ResourceNamespace
+import com.android.ide.common.resources.ResourceValueMap
+import com.android.ide.common.resources.configuration.FolderConfiguration
+import com.android.ide.common.resources.configuration.LocaleQualifier
+import com.android.resources.ResourceType
+import com.google.common.collect.Table
+
+private const val DO_NOT_TRANSLATE = "donottranslate"
+
+private val configCache = hashMapOf<Pair<ResourceSourceFile, LocaleQualifier>, RepositoryConfiguration>()
```
jrodbx, line 17: how beneficial is caching here?  i.e., how "heavy" or slow is it to calculate these extra resources at runtime?

also, will this translate framework resources as well?  if so, that alone might justify the caching.

would be curious if we also had some anecdotal measurements on how much the heap space grows as a result of this caching.  IIRC we saved about 1GB by migrating from legacy resource rewriting to current, is that correct?  If this new cache is in the order of MBs, then sgtm!

kevinzheng-ap: Resource rewriting saved `50MB` actually. 

And yes the cache is for framework resources as well. From my test, the framework resources memory use is `~2.8MB` without cache and reduced to `~1.7MB` with cache

kevinzheng-ap: The cache is about saving memory use. Without cache, we have to create a new `RepositoryConfiguration` every time when creating a new pseudo item. With cache, pseudo items can refer to cached `RepositoryConfiguration` for same source file. That’s how it saves memory

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+          else -> return@forEach
+        }
+
+        var repositoryConfiguration = configCache[value.sourceFile to localeQualifier]
+        if (repositoryConfiguration == null) {
+          repositoryConfiguration = value.sourceFile.configuration.onNewLocaleQualifier(localeQualifier)
+          configCache[value.sourceFile to localeQualifier] = repositoryConfiguration
+        }
+
+        val pseudoLocaleSourceFile = value.sourceFile.onNewRepositoryConfiguration(repositoryConfiguration)
```
jrodbx, line 51: Let's consolidate the two extension functions with ResourceSourceFile receiver + caching logic, since it adds a bit of noise here.  How about:

```suggestion
        val pseudoLocaleSourceFile = value.sourceFile.forLocale(localeQualifier)
```

then.... (see below)

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+}
+
+private fun ResourceSourceFile.onNewRepositoryConfiguration(configuration: RepositoryConfiguration): ResourceSourceFile {
+  return when (this) {
+    is ResourceSourceFileImpl -> ResourceSourceFileImpl(relativePath, configuration)
+    is ResourceFile -> ResourceFile(file, configuration)
+    // exhaustive to avoid potential breaking changes in future releases
+    else -> throw IllegalStateException("${javaClass.name} is not handled for new RepositoryConfiguration")
+  }
+}
```
jrodbx, line 105:
```suggestion
private fun ResourceSourceFile.forLocale(localeQualifier: LocaleQualifier): ResourceSourceFile {
  val repositoryConfiguration = configCache.getOrPut(this to localeQualifier) {
    RepositoryConfiguration(
      repository = configuration.repository,
      folderConfiguration = FolderConfiguration.copyOf(configuration.folderConfiguration).apply {
        this.localeQualifier = localeQualifier
      }
    )
  }
  return when (this) {
    is ResourceSourceFileImpl -> ResourceSourceFileImpl(relativePath, repositoryConfiguration)
    is ResourceFile -> ResourceFile(file, repositoryConfiguration)
    // exhaustive to avoid potential breaking changes in future releases
    else -> throw IllegalStateException("${javaClass.name} is not handled for new RepositoryConfiguration")
  }
}
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+            value as BasicPluralsResourceItem,
+            pseudoLocaleSourceFile,
+            method
+          )
+
+          ResourceType.ARRAY -> pseudolocalizeArray(
+            value as BasicArrayResourceItem,
+            pseudoLocaleSourceFile,
+            method
+          )
```
jrodbx, line 69: nit: braces + single line method calls

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+  }
+
+  return this
+}
+
+/**
+ * Check if the resource item is pseudolocalizable by verifying the source file name.
+ *
```
jrodbx, line 84: remove, since private comment

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+ * Currently we cannot verify the translatable attribute from raw xml as it is not stored in BasicValueResourceItemBase.
```
jrodbx, line 85:
```suggestion
 * Currently, we cannot respect the "translatable" attribute from string resources as it is not stored in BasicValueResourceItemBase.
```

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,144 @@
...
+ * This will cause an issue that untranslatable strings are translated. The solution will be a new
+ * translatable Boolean field in BasicValueResourceItemBase, but this requires changes in Android Open Source Project.
```
jrodbx, line 87:
```suggestion
 * As a result, for now, untranslatable strings will be considered pseudolocalizable.
 * TODO: fix this
```

The AOSP approach is one way; we could also update the resource classes + repository loaders themselves to provide this feature.  There may be other ways, so better not to bake that into the source comments for now.

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+
+  @Test
+  fun testAccentPseudolocalization() {
+    val result = createResourceTable()
+    result.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("en-rXA"))
+
+    // STRING
+    with(result[ResourceNamespace.TODO(), ResourceType.STRING]!!) {
+      assertThat(size).isEqualTo(4)
+      assertThat(get("string_name")?.value).isEqualTo("[Ţéšţ Šţŕîñĝ one two]")
```
jrodbx, line 64: let's either migrate from `get` to `getValue`, or use the `!!`.  null-safe'ing defers the crash to the assertion instead of throwing NPE right away.

same with the other instances below.

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    with(result[ResourceNamespace.TODO(), ResourceType.PLURALS]!!) {
+      assertThat(size).isEqualTo(1)
+      with(get("plural_name") as PluralsResourceValue) {
+        assertThat(pluralsCount).isEqualTo(2)
+        assertThat(getValue(0)).isEqualTo("[Ñöţĥîñĝ one two]")
+        assertThat(getValue(1)).isEqualTo("[Öñé Šţŕîñĝ one two]")
+      }
+    }
+
+    // Other resources should not be pseudo localized
```
jrodbx, line 91:
```suggestion
    // Other resources should not be pseudolocalized
```

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    // BOOL
+    with(result[ResourceNamespace.TODO(), ResourceType.BOOL]!!) {
+      assertThat(size).isEqualTo(2)
+      assertThat(get("screen_small")?.value).isEqualTo(true.toString())
+      assertThat(get("adjust_view_bounds")?.value).isEqualTo(false.toString())
```
jrodbx, line 96: use the literals "true" and "false" instead?

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    with(result[ResourceNamespace.TODO(), ResourceType.PLURALS]!!) {
+      assertThat(size).isEqualTo(1)
+      with(get("plural_name") as PluralsResourceValue) {
+        assertThat(pluralsCount).isEqualTo(2)
+        assertThat(getValue(0)).isEqualTo("${bidiWordStart}Nothing$bidiWordEnd")
+        assertThat(getValue(1)).isEqualTo("${bidiWordStart}One$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      }
+    }
+
+    // Other resources should not be pseudo localized
```
jrodbx, line 150:
```suggestion
    // Other resources should not be pseudolocalized
```

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+      assertThat(get("string_not_translate")?.value).isEqualTo("Not Translatable String")
+    }
+
+    // ARRAY
+    with(result[ResourceNamespace.TODO(), ResourceType.ARRAY]!!) {
+      assertThat(size).isEqualTo(1)
+      with(get("string_array_name") as ArrayResourceValue) {
+        assertThat(elementCount).isEqualTo(2)
+        assertThat(getElement(0)).isEqualTo("${bidiWordStart}First$bidiWordEnd ${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+        assertThat(getElement(1)).isEqualTo("${bidiWordStart}Second$bidiWordEnd ${bidiWordStart}Test$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
```
jrodbx, line 136: uber-nit: let's consistently use the curly braces for all the "bidi" val usages; improves readability

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+        assertThat(getValue(1)).isEqualTo("${bidiWordStart}One$bidiWordEnd ${bidiWordStart}String$bidiWordEnd")
+      }
+    }
+
+    // Other resources should not be pseudo localized
+    // BOOL
+    with(result[ResourceNamespace.TODO(), ResourceType.BOOL]!!) {
+      assertThat(size).isEqualTo(2)
+      assertThat(get("screen_small")?.value).isEqualTo(true.toString())
+      assertThat(get("adjust_view_bounds")?.value).isEqualTo(false.toString())
```
jrodbx, line 155: "true" and "false" literals

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+          putAll(items.map { it.name to it as ResourceValue })
+        }
+      )
+    }
+
+    return table
+  }
+
+  @Test
+  fun testAccentPseudolocalization() {
```
jrodbx, line 57:
```suggestion
  fun accent() {
```

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    // DIMEN
+    with(result[ResourceNamespace.TODO(), ResourceType.DIMEN]!!) {
+      assertThat(size).isEqualTo(2)
+      assertThat(get("textview_height")?.value).isEqualTo("25dp")
+      assertThat(get("textview_width")?.value).isEqualTo("150dp")
+    }
+  }
+
+  @Test
+  fun testBidiPseudolocalization() {
```
jrodbx, line 116:
```suggestion
  fun bidi() {
```

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -0,0 +1,209 @@
...
+    // DIMEN
+    with(result[ResourceNamespace.TODO(), ResourceType.DIMEN]!!) {
+      assertThat(size).isEqualTo(2)
+      assertThat(get("textview_height")?.value).isEqualTo("25dp")
+      assertThat(get("textview_width")?.value).isEqualTo("150dp")
+    }
+  }
+
+  @Test
+  fun testNoPseudolocalizationForNonPseudoLocale() {
```
jrodbx, line 175:
```suggestion
  fun doesNotPseudolocalizeForOtherLocales() {
```

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
jrodbx, line 91: nit: should the below resource types be moved out into their own test method?  i could go either way!

something like:
```
@Test
fun doesNotPseudolocalizeOtherResourceTypes() {
    val table = createResourceTable()

    table.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("en-rXA"))
    // BOOL
    // COLOR
    // DIMEN

    table.pseudolocalizeIfNeeded(LocaleQualifier.getQualifier("ar-rXB"))
    // BOOL
    // COLOR
    // DIMEN
}

**Review by jrodbx — APPROVED** [review]
Ok, this is really awesome!  Especially ❤️ how self-documenting the tests are.

So if I understand correctly, this implementation is post-processing the loaded resources via the root ResourceRepositories via decorator pattern.  That explains why we'd have to pass folderConfiguration twice (once to `getConfiguredResources`, then again to `pseudolocalizeIfNeeded`) and also explains why we have to cache RepositoryConfigurations in PseudolocaleGenerator.  Not ideal, but also not too bad, since the vast majority of the logic is extracted into its own testable components!

Also, comparing this to how AAPT generates pseudolocales, it looks like:
* donottranslate files are supported (i.e., don't pseudolocalize) ✅ 
* xliff untranslatable sections/placeholders are supported (i.e., don't pseudolocalize) ✅ 
* but...string resources with translatable="false" attribute are not supported (i.e., they will be pseudolocalized, even though they "shouldn't" be) ❌ 

Is that an accurate understanding?  If so, I think that's fine for now.   I'd then recommend submitting a PR to the Android Studio team to support this in AOSP via BasicValueResourceItem, BasicTextValueResourceItem, and the corresponding RepositoryLoader parsing code.

Is there anything else I'm missing ^

Either way, totally fine to tackle in a follow-up PR.

File: paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt [hunk 24]
```diff
@@ -0,0 +1,123 @@
...
+import com.android.ide.common.resources.configuration.FolderConfiguration
+import com.android.ide.common.resources.configuration.LocaleQualifier
+import com.android.resources.ResourceType
+import com.google.common.collect.Table
+
+private const val DO_NOT_TRANSLATE = "donottranslate"
+
+private val configCache = hashMapOf<Pair<ResourceSourceFile, LocaleQualifier>, RepositoryConfiguration>()
+
+fun String.pseudoLocalize(method: Pseudolocalizer.Method) =
```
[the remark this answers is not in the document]
kevinzheng-ap: it needs to be public as it will be used in real tests to pseudolocalize hardcoded strings

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+  if (isBlank()) {
+    this
+  } else {
```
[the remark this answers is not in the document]
kevinzheng-ap: we will not pseudolocalize empty strings, I did see some framework strings are empty

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+
+private fun pseudolocalizeString(
+  original: BasicValueResourceItem,
+  sourceFile: ResourceSourceFile,
+  method: Pseudolocalizer.Method
+): BasicValueResourceItem {
+  val pseudoText = original.value?.let { it.pseudoLocalize(method) }
+  val pseudoRawXml = (original as? BasicTextValueResourceItem)?.rawXmlValue?.let { it.pseudoLocalize(method) }
+
+  val pseudoItem = if (pseudoRawXml == null) {
```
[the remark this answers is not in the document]
kevinzheng-ap: This is [original logic](https://github.com/cashapp/paparazzi/blob/35ff04364eb9526ce31645026d45c09bbaedc3ad/paparazzi/src/main/java/app/cash/paparazzi/internal/resources/RepositoryLoader.kt#L705-L709) to handle rawXml in RepositoryLoader, so it is not a problem

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/ResourceFolderRepositoryTest.kt
```diff
@@ -66,6 +66,7 @@ class ResourceFolderRepositoryTest {
     assertThat(map[8].resourceValue.value).isEqualTo("#00000000")
     assertThat(map[9].name).isEqualTo("color_selector")
     assertThat(map[9].type).isEqualTo(ResourceType.COLOR)
+    assertThat(map[9].resourceValue.value).isEqualTo("$resFolderRoot/color/color_selector.xml")
```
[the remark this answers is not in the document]
kevinzheng-ap: It is a missing test before, are you ok to keep it here or I will have a seperate PR?

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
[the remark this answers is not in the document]
jrodbx: can you provide an example of how that might look?

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/ResourceFolderRepositoryTest.kt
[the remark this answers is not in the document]
jrodbx: neither.  I don't think the assertion is that useful, given that it has an interpolated variable and the other assertions highlight what the resourceValue.value is

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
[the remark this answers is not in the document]
kevinzheng-ap: if it is empty, it will be just like `[]` after pseudolocalization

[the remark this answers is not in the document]
kevinzheng-ap: `"Some thing".pseudoLocalize(ACCENT)`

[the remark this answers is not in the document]
jrodbx: Ah I understand now, but unfortunately, I don't think that's something we want to impose on test writers.  Anything hardcoded is not a resource and therefore is outside of Paparazzi's responsibility and in the test developer's hands.

e.g., if a <TextView> or Text() had text="Hardcoded", I'd imagine AAPT's pseudolocalizing wouldn't cover that case either.

So yea, returning to the original comment, I think we should still use private modifier

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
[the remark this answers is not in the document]
jrodbx: Sorry, I think my original question wasn't clear.

Put another way, what does "".pseudolocalize(ACCENT) or "".pseudolocalize(BIDI) look like?  I assume it'd simply output "".  Is that right?

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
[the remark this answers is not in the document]
kevinzheng-ap: No problem. I will make the change

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/PseudolocaleGenerator.kt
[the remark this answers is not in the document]
kevinzheng-ap: Yes, the output will be just "". However, if we always fall to the else clause, then the output will be "[]"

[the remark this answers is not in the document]
jrodbx: And is that undesirable?  Like, why not just simply do what AAPT does?  i imagine the [] acts like "" in this case, i.e., to highlight the emptiness of the string.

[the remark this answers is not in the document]
kevinzheng-ap: Got this error if we pseudo localize empty string. Will find out which string it is and how AAPT handles that.

```
app.cash.paparazzi.sample.ResourcesTest > compose[Accent] STANDARD_ERROR
    Aug 10, 2023 1:33:59 PM app.cash.paparazzi.internal.PaparazziLogger logAndroidFramework
    INFO: ViewRootImpl [4]: Unable to create the InputEventCompatProcessor. 
    java.lang.ClassNotFoundException: []
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:375)
        at android.view.ViewRootImpl.<init>(ViewRootImpl.java:941)
        at android.view.ViewRootImpl.<init>(ViewRootImpl.java:885)
        at android.view.AttachInfo_Accessor.setAttachInfo(AttachInfo_Accessor.java:37)
        at com.android.layoutlib.bridge.impl.RenderSessionImpl.inflate(RenderSessionImpl.java:360)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:183)
        at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:124)
        at com.google.testing.junit.testparameterinjector.PluggableTestRunner$ContextMethodRule$1.evaluate(PluggableTestRunner.java:429)
```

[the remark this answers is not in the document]
kevinzheng-ap: OK, I found the cause. [ViewRootImpl](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ViewRootImpl.java;drc=36a2b8f0b6545bc330415dcf667aa1a6e14482f6;l=976) tries to get `config_inputEventCompatProcessorOverrideClassName` and uses it if it is not empty. However, if the empty string is pseudolocalized, we will try to find a class with name `[]`, which does not exist

And in original [resource xml](https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/values/config.xml), the string resource is 
`<string name="config_inputEventCompatProcessorOverrideClassName" translatable="false"></string>`. So in AAPT will not translate it. But given we cannot handle `translatable="false"` for now, we have to skip such resource by the empty check. I will put more comments there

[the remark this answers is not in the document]
jrodbx: aha, makes sense!  thanks for the explanation.

PR #1014, opened 2023-07-20
Silence some warnings on IDE import

TWiStErRob:
See expanded commit messages for warnings.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-android/gradle.properties
+++ b/paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-android/gradle.properties
@@ -1,0 +1,1 @@
+kotlin.mpp.androidSourceSetLayoutVersion1.nowarn=true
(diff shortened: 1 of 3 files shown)


jrodbx (approved):

paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-android/gradle.properties
```diff
@@ -0,0 +1 @@
+kotlin.mpp.androidSourceSetLayoutVersion1.nowarn=true
```
jrodbx, line 1: is this one we should fix instead?

TWiStErRob: If you know how :), there's no code in that project just some plugin applied.

jrodbx: i don't.  alright then, figured i'd ask :)

Issue #709, opened 2023-01-30
Exception thrown when snapshotting GoogleMap composable


BrianGardnerAtl:
**Description**
The GoogleMap composable is able to render in Android Studio Flamingo | 2022.2.1 Canary 10 (haven't tested on other versions yet) but attempting to snapshot test it using paparazzi throws an exception. 

**Steps to Reproduce**
Repro project here: https://github.com/BrianGardnerAtl/PaparazziMapRepro
View the `PaparazziMap` composable and see how it renders (a blank box) in the Preview.
Attempt to record paparazzi screenshot from `PaparazziMapText` and see the exception

**Expected behavior**
Suuccessful snapshot taken of the composable. Since the GoogleMap composable short circuits and displays a plain box I don't expect to see the actual map, but not crashing would be nice.

**Additional information:**
- Paparazzi Version: 1.2.0
- OS: mac Ventura 13.2
- Compile SDK: 33
- Gradle Version: 
- Android Gradle Plugin Version:

**Exception log**
```
java.lang.NullPointerException: Cannot read field "metaData" because the return value of "com.google.android.gms.common.wrappers.PackageManagerWrapper.getApplicationInfo(String, int)" is null
	at com.google.android.gms.common.internal.zzag.zzc(com.google.android.gms:play-services-basement@@18.0.0:5)
	at com.google.android.gms.common.internal.zzag.zza(com.google.android.gms:play-services-basement@@18.0.0:1)
	at com.google.android.gms.common.GooglePlayServicesUtilLight.isGooglePlayServicesAvailable(com.google.android.gms:play-services-basement@@18.0.0:7)
	at com.google.android.gms.common.GooglePlayServicesUtil.isGooglePlayServicesAvailable(com.google.android.gms:play-services-base@@18.0.1:2)
	at com.google.android.gms.maps.internal.zzcb.zza(com.google.android.gms:play-services-maps@@18.1.0:3)
	at com.google.android.gms.maps.MapsInitializer.initialize(com.google.android.gms:play-services-maps@@18.1.0:4)
	at com.google.android.gms.maps.MapsInitializer.initialize(com.google.android.gms:play-services-maps@@18.1.0:1)
	at com.google.android.gms.maps.zzah.zzb(com.google.android.gms:play-services-maps@@18.1.0:2)
	at com.google.android.gms.maps.zzah.createDelegate(com.google.android.gms:play-services-maps@@18.1.0:1)
[...]

PR #1017, opened 2023-07-24
Wrap composables with a CompositionLocalProvider to set LocalInspectionMode to true

adamalyyan:
This addresses issue #709 where the `GoogleMap`s composable would crash when taking a snapshot. This was happening because the `GoogleMap`s composable was trying to initialize as it would when placed in an app. This doesn't happen in Android Studio's preview mode since the composable is wrapped with a `CompositionLocalProvider` which sets `LocalInspectionMode` to `true`. This is done in the `ComposeViewAdpater`. To mimic this behavior, before we set the composable to be snapshot on the ComposeView, we wrap it the same way the ComposeViewAdapter does. 

The ComposeViewAdapter used by `Preview` does internally, but since we don't have a reference to the paparazzi `ComposeViewAdapter` when snapshot is called, we do the wrapping outside before the `ComposeView` is added to the `ComposeViewAdapter`. The end result and layout tree is identical, it just happens at slightly different times in Paparazzi vs Android Studio (both happen before any preview/snapshot is generated)

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -32,7 +32,9 @@ import android.view.ViewGroup
 import androidx.activity.setViewTreeOnBackPressedDispatcherOwner
 import androidx.annotation.LayoutRes
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.LocalInspectionMode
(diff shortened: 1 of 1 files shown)


jrodbx:
> This doesn't happen in Android Studio's preview mode since the composable is wrapped with a CompositionLocalProvider which sets LocalInspectionMode to true

IIUC, this is similar to View.isInEditMode where Preview sets this to true, but Paparazzi deliberately sets to true, to run production code paths.  The original issue shows a NPE with respect to some missing "metadata" field; would it be possible to figure out how to resolve the NPE and see how much further along the execution flow would go?

Issue #508, opened 2022-07-23
Support Compose LocalInspectionMode


yschimke:
**Description**

LocalInspectionMode allows composables to present nicely in previews, even when layout or images are not immediately available.

See https://github.com/coil-kt/coil/blob/70300e5909a561d7e1026c660b8f28be3f0051a6/coil-compose-base/src/main/java/coil/compose/AsyncImagePainter.kt#L148

Also https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/ScalingLazyColumn.kt;l=404?q=ScalingLazyColumn

It would be nice to default this, and also allow specifying to paparazzi.gif or paparazzi.snapshot.

**Additional information:**
- Paparazzi Version: 1.0

hlandim:
I managed to overcome it by mocking the LocalInspectionMode with mockk
```
        paparazzi.snapshot {
            mockkStatic("androidx.compose.ui.platform.InspectionModeKt")
            every { LocalInspectionMode } returns staticCompositionLocalOf { true }
            MyComposableScreen()
        }
```

yschimke:
I think this works without mocking, but the original request was whether this should be done automatically, so components that might use the network such as image loaders might use more useful defaults.

        paparazzi.snapshot {
            CompositionLocalProvider(LocalInspectionMode provides true) {
              MyComposableScreen()
            }
        }

yschimke:
Another example https://github.com/cashapp/paparazzi/issues/735

adamalyyan:
Our stance is that Paparazzi shouldn't set LocalInspectionMode globally as the goal is to provide a snapshot of the true production output. Some Composables, such as `GoogleMap()`, check for `LocalInspectionMode` to short-circuit to a `@Preview`-safe Composable and others may use this flag to show debug layouts/information which means the snapshot is no longer showing what a user would see in production.

Our recommendation is to do what @yschimke mentioned and wrap the composable with a CompositionLocalProvider and set LocalInspectionMode for your individual composable that require it. The linked PR will add a section to the README highlighting this.

adamalyyan:
Related to #709

PR #1061, opened 2023-08-25
Add LocalInspectionMode section to README

adamalyyan:
Adding documentation to the README around LocalInspectionMode to provide context on the resolution of #709. This section outlines why Paparazzi doesn't set `LocalInspectionMode` globally and provides a solution for users who run into this issue.

Addresses #508 as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,25 @@ fun setup() {
 }
 ```
 
+LocalInspectionMode
+--------
+Some Composables -- such as `GoogleMap()` -- check for `LocalInspectionMode` to short-circuit to a `@Preview`-safe Composable.
+
+However, Paparazzi does not set `LocalInspectionMode` globally to ensure that the snapshot represents the true production output, similar to how it overrides `View.isInEditMode` for legacy views.
+
+As a workaround, we recommend wrapping such a Composable in a custom Composable with a `CompositionLocalProvider` and setting `LocalInspectionMode` there.
+
+```kotlin
+ @Test
+  fun inspectionModeView() {
+    paparazzi.snapshot(
+      CompositionLocalProvider(LocalInspectionMode provides true) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,6 +116,23 @@ fun setup() {
...
+
+```kotlin
+ @Test
+  fun inspectionModeView() {
+    paparazzi.snapshot(
+      CompositionLocalProvider(
+        LocalInspectionMode provides true
+        ) {
+          YourComposable()
+        }
```
jrodbx, line 131:
```suggestion
      CompositionLocalProvider(LocalInspectionMode provides true) {
        YourComposable()
      }
```

README.md
```diff
@@ -116,6 +116,23 @@ fun setup() {
 }
 ```
 
+LocalInspectionMode
+--------
+Paparazzi does not set `LocalInspectionMode` globally to ensure that the snapshot represents the true production output. Some composables, such as GoogleMap's, check this flag to replace their composable with a preview/snapshot safe composable and as a result Paparazzi will throw an exception when it isn't set. To solve this, we recommend wrapping your composition with a `CompositionLocalProvider` and setting `LocalInspectionMode` if you have composables that require it.
```
jrodbx, line 121:
```suggestion
Some Composables -- such as `GoogleMap()` -- check for `LocalInspectionMode` to short-circuit to a `@Preview`-safe Composable.

However, Paparazzi does not set `LocalInspectionMode` globally to ensure that the snapshot represents the true production output, similar to how it overrides `View.isInEditMode` for legacy views.

As a workaround, we recommend wrapping such a Composable in a custom Composable with a `CompositionLocalProvider` and setting `LocalInspectionMode` there.
```

jrodbx (approved):

Issue #1070, opened 2023-08-31
Plugin hardcodes reports directory instead of resolving via Gradle


JakeWharton:
Using 1.3.1, the source hardcodes the `reports/` directory underneath the build directory. Instead, it should resolve the directory by looking up the reporting extension: https://docs.gradle.org/current/dsl/org.gradle.api.reporting.ReportingExtension.html.

Example: https://github.com/cashapp/licensee/blob/55f0a294fce392ee82cae58e958866c9a915c41f/src/main/kotlin/app/cash/licensee/plugin.kt#L52

PR #1081, opened 2023-09-08
Resolve report dir from ReportingExtension instead of hardcoding

jrodbx:
Closes https://github.com/cashapp/paparazzi/issues/1070


JakeWharton (approved):

paparazzi-gradle-plugin/src/test/projects/custom-report-dir/build.gradle
```diff
@@ -0,0 +1,24 @@
+plugins {
+  id 'com.android.library'
+  id 'kotlin-android'
+  id 'app.cash.paparazzi'
+}
+
+reporting {
+  baseDir 'custom/our-reports'
+}
```
geoff-powell, line 9: nit: any interest in making this block `paparazzi { reporting { baseDir '...' } }`. I could see this possibly being leveraged to enable validate accessibility errors and configure allow lists in the future

geoff-powell (approved):

paparazzi-gradle-plugin/src/test/projects/custom-report-dir/build.gradle
JakeWharton: You could make a `paparazzi { reportDir .. }`, but that property within your own extension would use a default value sourced from the `reporting { }` `baseDir` value as it's a standard Gradle mechanism.

TWiStErRob: I think to take a step back: the question is if it's necessary to allow users of Paparazzi to specifically move the reports folder, or it's enough to put Paparazzi reports in the same places as "all the other reports". I think it would be nice to be able to do it, but potentially it wouldn't be used as much.

jrodbx: @geoff-powell As the rest have mentioned, this is more of a general reporting configuration vs a Paparazzi-specific one.

PR #1084, opened 2023-09-12
Add View.isInEditMode section in README.md

linean:
While reading the README, I found that the expected behavior of `View.isInEditMode` in Paparazzi was unclear to me. I had to delve into the [code](https://github.com/cashapp/paparazzi/pull/145) to grasp it. I suggest that we enhance the README by adding a dedicated section explaining the behavior of `View.isInEditMode`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,11 +116,23 @@ fun setup() {
 }
 ```
 
+View.isInEditMode
+--------
+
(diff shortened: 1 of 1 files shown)


jrodbx:
Thanks for the contribution, but this implementation detail is both inferred from the premise of the project and the reference in LocalInspectionMode.

PR #1091, opened 2023-09-14
Introducing paparazzi-annotations module

nak5ive:
To be used in a follow-up PR, where we introduce a new KSP plugin.

### DIFF
--- a/paparazzi-annotations/build.gradle
+++ b/paparazzi-annotations/build.gradle
@@ -1,0 +1,4 @@
+apply plugin: 'org.jetbrains.kotlin.jvm'
+apply plugin: 'com.vanniktech.maven.publish'
+
+dependencies {}
--- a/paparazzi-annotations/build.gradle
+++ b/paparazzi-annotations/build.gradle
@@ -1,4 +1,2 @@
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'com.vanniktech.maven.publish'
-
-dependencies {}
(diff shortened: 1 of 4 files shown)


paparazzi-annotations/build.gradle
```diff
@@ -0,0 +1,4 @@
+apply plugin: 'org.jetbrains.kotlin.jvm'
+apply plugin: 'com.vanniktech.maven.publish'
+
+dependencies {}
```
jrodbx, line 4: empty block can probably be removed?

paparazzi-annotations/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=paparazzi-annotations
+POM_NAME=Paparazzi Annotations API
```
jrodbx, line 2:
```suggestion
POM_NAME=Paparazzi Annotations
```

jrodbx (approved):

paparazzi-annotations/src/main/java/app/cash/paparazzi/annotation/api/Paparazzi.kt
```diff
@@ -0,0 +1,5 @@
+package app.cash.paparazzi.annotation.api
+
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.BINARY)
+annotation class Paparazzi
```
TWiStErRob, line 5: Are you sure about the name? It seems a bit too generic to me, Potentially lending itself to being a god-annotation with many disparate parameters in the future.

jrodbx: yep, we don't expect to use this outside of `@Preview` nor to add any parameters.  also, fine to break in the future if we're wrong.

PR #1092, opened 2023-09-14
Introduce a simple APNG Reader / Writer

gamepro65:
Implements a basic APNG Reader / Writer which includes a subset of the APNG spec (no filters / interlacing / blend). We compare frames to identify and encode only the smallest dirty rect that has found using a [fcTL](https://www.w3.org/TR/png/#fcTL-chunk) chunk which helps cut down on final file size. 

In a follow up will be used as the foundation for lossless gif diffing


paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+import java.io.File
+import java.lang.IllegalStateException
+import java.util.zip.CRC32
+import java.util.zip.Inflater
+
+/**
+ * Provides a simple sub-spec implementation for reading APNG files. Only handles
+ * chunks which are implemented by the writer.
+ */
+internal class APNGReader(file: File) : Closeable {
```
jrodbx, line 49: nit: i think Java convention is to use `Apng` vs `APNG` as a prefix?

```suggestion
internal class ApngReader(file: File) : Closeable {
```

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+import java.lang.Integer.max
+import java.util.zip.CRC32
+import java.util.zip.Deflater
+
+/**
+ * Provides a basic implementation for encoding APNG's.
+ *
+ * Refs: https://www.w3.org/TR/png/#4Concepts.Encoding, https://wiki.mozilla.org/APNG_Specification
+ */
+internal class APNGWriter(
```
jrodbx, line 50: similarly?

```suggestion
internal class ApngWriter(
```

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGReaderTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    assertThat(reader.frameCount).isEqualTo(reader.frameNumber)
+  }
+
+  @Test(expected = IllegalStateException::class)
+  fun enforcesAnimationChunkSequence() {
+    val file = javaClass.classLoader.getResource("invalid_sequence.png")
+    val reader = APNGReader(File(file!!.toURI()))
+
+    while (!reader.finished()) { reader.getNextFrame() }
+  }
```
jrodbx, line 49: nit: private and move to end of this file

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGReaderTest.kt
```diff
@@ -0,0 +1,65 @@
...
+      val expectedImage = ImageIO.read(expectedFile)
+      val actualImage = reader.getNextFrame()!!
+      ImageUtils.assertImageSimilar(expectedFile!!.path, expectedImage, actualImage, 0.0)
+    }
+
+    assertThat(reader.finished()).isTrue()
+    assertThat(reader.frameCount).isEqualTo(reader.frameNumber)
+  }
+
+  @Test(expected = IllegalStateException::class)
```
jrodbx, line 43: IIRC, technically `@Test(expected)` is considered bad form, since the exception doesn't specify precisely the call site from which it's thrown.

Accordingly, can we move this into the form:

```
try {
  body()
  fail()
} catch (e: IllegalStateException) {
  assertThat(e).hasMessage("whevs")
} 
```
?

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGReaderTest.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun enforcesAnimationChunkSequence() {
+    val file = javaClass.classLoader.getResource("invalid_sequence.png")
+    val reader = APNGReader(File(file!!.toURI()))
+
+    while (!reader.finished()) { reader.getNextFrame() }
+  }
+
+  @Test(expected = IllegalStateException::class)
```
jrodbx, line 51: similar

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGReaderTest.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun enforcesCRC() {
+    val file = javaClass.classLoader.getResource("invalid_crc.png")
+    val reader = APNGReader(File(file!!.toURI()))
+
+    while (!reader.finished()) { reader.getNextFrame() }
+  }
+
+  @Test(expected = IllegalStateException::class)
```
jrodbx, line 59: similar

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,305 @@
...
+import java.awt.image.BufferedImage
+import java.io.File
+import java.util.zip.CRC32
+import java.util.zip.Inflater
+
+class APNGWriterTest {
+
+  @Test
+  fun writesAnimationMetadata() {
+    val testFile = File.createTempFile("writesAnimationMetadata.png", null)
```
jrodbx, line 48: will this temp file be cleaned up?  Should a TemporaryFolder rule be used instead?

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,305 @@
...
+    val testFile = File.createTempFile("writesAllFramesWithSameFrameRate.png", null)
+    APNGWriter(testFile, 3, 3).use { writer ->
+      writer.writeImage(createImage(squareOffset = Point(5, 5)))
+      writer.writeImage(createImage(squareOffset = Point(15, 15)))
+      writer.writeImage(createImage(squareOffset = Point(25, 25)))
+    }
+
+    FileSystem.SYSTEM.openReadOnly(testFile.path.toPath()).use {
+      it.source(0L).buffer().use {
+        it.skip(PNG_SIG.size.toLong()) // Header
```
jrodbx, line 133: nit: can we use named lambda args for clarity/disambiguation of nested `it`s?

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,305 @@
...
+
+        val (idat, idatData) = it.assertNextChunk()
+        assertThat(idat).isEqualTo(IDAT)
+        val decompress = idatData.decompress()
+        assertThat(decompress.size).isEqualTo((MAX_SIZE * MAX_SIZE * 4L) + MAX_SIZE) // 4 Bytes Per Pixel + 1 Byte Per Row
+      }
+    }
+  }
+
+  @Test(expected = IllegalStateException::class)
```
jrodbx, line 239: similar with expected

paparazzi/src/test/resources/simple_animation.png
jrodbx, line 1: nice!

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,305 @@
...
+      g.dispose()
+    }
+
+  private fun BufferedSource.decompress(): Buffer {
+    val inflater = Inflater().apply {
+      val readByteArray = readByteArray()
+      setInput(readByteArray)
+    }
+
+    val buffer = ByteArray(1_024_000)
```
jrodbx, line 286: nit: magic number?  extract to const?

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,305 @@
...
+import app.cash.paparazzi.internal.apng.PngConsts.PNG_BITS_PER_PIXEL
+import app.cash.paparazzi.internal.apng.PngConsts.PNG_COLOR_TYPE_RGBA
+import app.cash.paparazzi.internal.apng.PngConsts.PNG_SIG
+import com.google.common.truth.Truth.assertThat
+import okio.Buffer
+import okio.BufferedSource
+import okio.FileSystem
+import okio.Path.Companion.toPath
+import okio.buffer
+import okio.internal.commonToUtf8String
```
jrodbx, line 35: is it ok to rely on okio internals?  is there an alternate public Okio method to use?

jrodbx (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
JakeWharton: Lots of violations of our best practice naming rules here and blow in a few function names.

Acronyms are treated as words, so this should be `Apng` reader. Saves us from situations like Java's `HttpURLConnection` or iOS's `NSURLConnection`.

Kotlin's general rules are here (which have weird carve-out for two-letter acronyms that I don't like): https://kotlinlang.org/docs/coding-conventions.html#choose-good-names

Google's Kotlin and Java ones are better: https://developer.android.com/kotlin/style-guide#camel_case. Results in much more consistency.

JakeWharton: This constructor should accept a `FileHandle` directly not a `File` to remove the implicit dependency on the system filesystem. Dependency inversion FTW. It will make testing below easier since you can operate entirely in-memory rather than on a real filesystem.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+  private var sequenceNumber = -1
+
+  var frameCount = 1
+  var frameNumber = 0
+  var width = 0
+  var height = 0
+
+  init {
+    val header = buffer.readByteArray(PNG_SIG.size.toLong())
+    if (!header.contentEquals(PNG_SIG)) {
```
JakeWharton: Consider `rangeEquals` + `skip` to avoid allocating in the happy path.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+import java.util.zip.Inflater
+
+/**
+ * Provides a simple sub-spec implementation for reading APNG files. Only handles
+ * chunks which are implemented by the writer.
+ */
+internal class APNGReader(file: File) : Closeable {
+  private val crcEngine = CRC32()
+  private val fileHandle = FileSystem.SYSTEM.openReadOnly(file.path.toPath())
+  private val buffer = fileHandle.source().buffer()
```
JakeWharton: nit: this is probably better named `source` since it's `BufferedSource` where `Source` is the noun and `Buffered` is the adjective. There's also a `Buffer` type which we don't want to make it seem like we're using.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+
+  var frameCount = 1
+  var frameNumber = 0
+  var width = 0
+  var height = 0
+
+  init {
+    val header = buffer.readByteArray(PNG_SIG.size.toLong())
+    if (!header.contentEquals(PNG_SIG)) {
+      throw IllegalStateException("Input is not a valid PNG")
```
JakeWharton: Consider including expected vs. actual hexidecimal bytes here

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+  private fun BufferedSource.parseIHDR() {
+    width = readInt()
+    height = readInt()
+    scratchSpace = IntArray(width * height)
+
+    skip(1L) // Bit depth
+    hasAlpha = when (readByte()) {
+      PNG_COLOR_TYPE_RGB -> false
+      PNG_COLOR_TYPE_RGBA -> true
+      else -> throw IllegalStateException("Unsupported color model")
```
JakeWharton: consider including hex of offending byte in message

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+    if (sequenceNumber + 1 != nextSequence) {
+      throw IllegalStateException("Out of order sequence, expected: ${sequenceNumber + 1} actual: $nextSequence")
+    }
+    sequenceNumber = nextSequence
+    decompress().decodeToScratchSpace()
+  }
+
+  private fun BufferedSource.decompress(): Buffer {
+    val inflater = Inflater().apply {
+      setInput(readByteArray())
```
JakeWharton: Future optimization: Use Okio's built-in `InflaterSource` to avoid having to allocate this giant array. The built-in one will share the backing byte arrays that you don't get to see directly with the `Inflater` to avoid copies / allocations.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    val buffer = ByteArray(BUFFER_SIZE)
+    return Buffer().apply {
+      do {
+        val byteCount = inflater.inflate(buffer)
+        write(buffer, 0, byteCount)
```
JakeWharton: ...it will also save a copy here since currently Okio cannot trust this byte array and needs to copy its contents into its own instances.

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGReaderTest.kt
TWiStErRob: The `Test` annotation Javadoc explains it, and there's a org.junit.Assert.assertThrows for replacement rather than `try-catch`. There's also a Truth's `ExpectFailure`, but I can't decipher how to use it in a simple way.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,265 @@
...
+      for (x in 0 until frameWidth) {
+        val pixelBytes = if (hasAlpha) readByteArray(4) else readByteArray(3)
+        val rgba = Buffer().run {
+          write(pixelBytes)
+          if (!hasAlpha) {
+            writeByte(0)
+          }
+
+          readInt()
+        }
```
JakeWharton, line 209: Kill these allocations!
```suggestion
        val rgba = if (hasAlpha) {
          readInt()
        } else {
          (readByte() and 0xff shl 24)
            or (readByte() and 0xff shl 16)
            or (readByte() and 0xff shl 8)
        }
```

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/APNGWriterTest.kt
```diff
@@ -0,0 +1,308 @@
...
+import org.junit.Test
+import java.awt.Color
+import java.awt.Point
+import java.awt.image.BufferedImage
+import java.io.File
+import java.util.zip.CRC32
+import java.util.zip.Inflater
+
+private const val i = 1_024_000
```
TWiStErRob, line 44: leftover from extracting `ONE_MB`?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,265 @@
...
+
+    crcEngine.reset()
+    crcEngine.update(chunkId, 0, 4)
+    if (dataLength > 0) crcEngine.update(data, 0, dataLength)
+
+    if (crcEngine.value.toInt() != crc) {
+      throw IllegalStateException("CRC Mismatch decoding ${chunkId.commonToUtf8String()}, invalid data")
+    }
+
+    return Header.valueOf(chunkId.commonToUtf8String().uppercase()) to Buffer().apply { write(data) }
```
JakeWharton: Instead of reading chunk to byte array and then converting to string read it directly to a UTF-8 string to avoid allocation. Should just be something like `readUt8String(4)`

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    val data = readByteArray(dataLength.toLong())
+    val crc = readInt()
+
+    crcEngine.reset()
+    crcEngine.update(chunkId, 0, 4)
+    if (dataLength > 0) crcEngine.update(data, 0, dataLength)
+
+    if (crcEngine.value.toInt() != crc) {
+      throw IllegalStateException("CRC Mismatch decoding ${chunkId.decodeToString()}, invalid data")
+    }
```
JakeWharton, line 232: Okio has a `HashingSink` and `HashingSource`, but I don't think it does CRC. We should add one to avoid this giant allocation.

JakeWharton (commented):
Gotta run. Didn't review the writer yet. Good stuff in general!

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/APNGReader.kt
gamepro65: 🤯

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -312,6 +313,40 @@ internal object ImageUtils {
...
   }
 
+  fun smallestDiffRect(firstImage: BufferedImage, secondImage: BufferedImage): Rectangle? {
+    var rect: Rectangle? = null
+    val width = max(firstImage.width, secondImage.width)
+    val height = max(firstImage.height, secondImage.height)
+
+    for (y in 0 until height) {
+      for (x in 0 until width) {
+        val firstRgb = if (x < firstImage.width && y < firstImage.height) firstImage.getRGB(x, y) else 0
```
swankjesse, line 323: `0` is a real value here; black. Is that okay?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -312,6 +313,40 @@ internal object ImageUtils {
...
+    return rect
+  }
+
+  fun BufferedImage.resize(targetWidth: Int, targetHeight: Int): BufferedImage {
+    if (width == targetWidth && height == targetHeight) {
+      return this
+    }
+
+    return BufferedImage(targetWidth, targetHeight, type).apply {
+      val g = graphics
```
swankjesse, line 344: I think you want this for better image quality?
```
 g.setRenderingHint(RenderingHints.KEY_INTERPOLATION,
                RenderingHints.VALUE_INTERPOLATION_BICUBIC);
```

swankjesse:
```suggestion
      val g = createGraphics()
```

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -312,6 +313,40 @@ internal object ImageUtils {
...
+          }
+        }
+      }
+    }
+
+    return rect
+  }
+
+  fun BufferedImage.resize(targetWidth: Int, targetHeight: Int): BufferedImage {
+    if (width == targetWidth && height == targetHeight) {
```
swankjesse, line 339: I don’t like this optimization because the subject is mutable. In the actually-resizing case I’m safe to further manipulate the result as much as I like; in the not-resizing case I can’t.

Maybe just don’t ?

swankjesse: (And if the 1:1 case is common, and worthy of optimization, make that the caller’s problem?)

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+import java.io.File
+import java.lang.IllegalStateException
+import java.util.zip.CRC32
+import java.util.zip.Inflater
+
+/**
+ * Provides a simple sub-spec implementation for reading APNG files. Only handles
+ * chunks which are implemented by the writer.
+ */
+internal class ApngReader(file: File) : Closeable {
```
swankjesse, line 48: Go full Okio? Change this API to accept a `Path` and a `FileSystem` !

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+  private val crcEngine = CRC32()
+  private val fileHandle = FileSystem.SYSTEM.openReadOnly(file.path.toPath())
+  private val buffer = fileHandle.source().buffer()
```
swankjesse, line 51:
```suggestion
  private val buffer = Buffer()
```

What this does is it creates a `Source` object and immediately leaks it.

swankjesse: Actually no it doesn’t, don’t mind me.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
gamepro65: This ByteArray is unfortunately needed for updating the CRC which takes either the ByteArray or ByteBuffer

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
swankjesse, line 51: I don’t like the name of this because it’s a stream but it’s named like a buffer. Maybe call it `source` instead?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+  private var hasAlpha = false
+  private var sequenceNumber = -1
+
+  var frameCount = 1
+  var frameNumber = 0
+  var width = 0
+  var height = 0
+
+  init {
+    val header = buffer.readByteArray(PNG_SIG.size.toLong())
```
swankjesse, line 65: Read a ByteString so you can use a plain old `!=` instead of ByteArray’s awkward `contentEquals()` function?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    if (!header.contentEquals(PNG_SIG)) {
+      throw IllegalStateException("Input is not a valid PNG")
```
swankjesse, line 67: This looks a lot like a `check()` statement

Probably best to throw `IOException` on invalid data in a file

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    processMetadata()
+  }
+
+  fun getNextFrame(): BufferedImage? {
```
swankjesse, line 73: Is this `readNextFrame()` ? I don’t like it when `get` prefixed functions have side-effects

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+
+  private fun processMetadata() {
+    processTo(IHDR)
+
+    if (buffer.peekNextChunkId().contentEquals(ACTL.bytes)) {
+      processTo(ACTL)
+    }
+  }
+
+  fun finished() = buffer.peekNextChunkId().contentEquals(IEND.bytes)
```
swankjesse, line 106: I don’t like this function name because it looks both like a question “is this finished” and also something to apply a side-effect “this is finished”.

Better to call it `isFinished()`

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+
+  override fun close() {
+    fileHandle.close()
+    buffer.close()
```
swankjesse, line 110: Yeah I don’t like the name `buffer` for this. It’s like calling a motorbike a `motor`

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+  private fun processTo(header: Header) {
+    do {
+      val (currentHeader, data) = buffer.readChunk()
+      when (currentHeader) {
+        IHDR -> data.parseIHDR()
+        ACTL -> data.parseACTL()
+        FCTL -> data.parseFCTL()
+        IDAT -> data.parseIDAT()
+        FDAT -> data.parseFDAT()
+        IEND -> throw IllegalStateException("PNG ended while processing to $header")
```
swankjesse, line 122: IOException ?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+      }
+    } while (!header.bytes.contentEquals(currentHeader.bytes) && !finished())
```
swankjesse, line 124: Make currentHeader.bytes into a ByteString

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  private fun BufferedSource.parseIHDR() {
```
swankjesse, line 127: I believe this uses the words `parse` and `read` interchangeably? Maybe `process` also? 

My preference is to stick to one, `read` if at all possible

(If there’s functions that do different things, they should have different prefixes)

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    width = readInt()
+    height = readInt()
+    scratchSpace = IntArray(width * height)
+
+    skip(1L) // Bit depth
+    hasAlpha = when (readByte()) {
+      PNG_COLOR_TYPE_RGB -> false
+      PNG_COLOR_TYPE_RGBA -> true
+      else -> throw IllegalStateException("Unsupported color model")
```
swankjesse, line 136: IOException

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+      width = readInt(),
+      height = readInt(),
+      offsetX = readInt(),
+      offsetY = readInt(),
+      delayNumerator = readShort(),
+      delayDenominator = readShort()
+    )
+
+    if (sequenceNumber + 1 != nextSequence) {
+      throw IllegalStateException("Out of order sequence, expected: ${sequenceNumber + 1} actual: $nextSequence")
```
swankjesse, line 156: IOException

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+  private fun BufferedSource.parseFDAT() {
+    val nextSequence = readInt()
+    if (sequenceNumber + 1 != nextSequence) {
+      throw IllegalStateException("Out of order sequence, expected: ${sequenceNumber + 1} actual: $nextSequence")
+    }
+    sequenceNumber = nextSequence
+    decompress().decodeToScratchSpace()
+  }
+
+  private fun BufferedSource.decompress(): Buffer {
```
swankjesse, line 174: Neato

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    val inflater = Inflater().apply {
```
swankjesse, line 175: I think this function can be smaller

```
InflaterSource(this, Inflater()).use { inflaterSource ->
  return Buffer().writeAll(inflaterSource)
}
```

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
TWiStErRob: @swankjesse Hungarian does that ;)

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
gamepro65: Yes it's ok, if they both end up being black then nones the wiser as the pixel "hasn't" changed between the images so we don't need this specific pixel, also if the images are different sized then we likely are failing and will be generating the delta gif which would cover this case

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    val frameOffsetY = frameInfo?.offsetY ?: 0
+
+    for (y in 0 until frameHeight) {
+      val filterByte = readByte()
+      if (filterByte != 0.toByte()) {
+        throw IllegalStateException("Only filter None is supported, $filterByte")
+      }
+
+      for (x in 0 until frameWidth) {
+        val pixelBytes = if (hasAlpha) readByteArray(4) else readByteArray(3)
```
swankjesse, line 201: Also pretty allocation-heavy! This might run faster

val rgba = when {
  hasAlpha -> readInt()
  else -> (readByte() and 0xff shl 16) + (readByte() and 0xff shl 8) + (readByte() and 0xff)
}

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
TWiStErRob: Even if firstImage is less wide than secondImage, and secondImage has black/non-black pixels on the wider part?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+        scratchSpace[position] = argb
+      }
+    }
+
+    frameNumber++
+  }
+
+  private fun BufferedSource.readChunk(): Pair<Header, BufferedSource> {
+    val dataLength = readInt()
+    val chunkId = readByteArray(4L)
```
swankjesse, line 222: Keep this as an int?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
gamepro65: Should I just call `setRenderingHints`?

```
  private fun setRenderingHints(g2: Graphics2D) {
    g2.setRenderingHint(KEY_INTERPOLATION, VALUE_INTERPOLATION_BILINEAR)
    g2.setRenderingHint(KEY_RENDERING, VALUE_RENDER_QUALITY)
    g2.setRenderingHint(KEY_ANTIALIASING, VALUE_ANTIALIAS_ON)
  }
```

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
swankjesse: Yeah lets do that. CRC isn’t going anywhere, sadly

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
swankjesse: ahhh tragically you cannot

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
swankjesse: A good rule of thumb when doing I/O: if there’s a ByteArray anywhere then Jesse won’t like it

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+
+    if (crcEngine.value.toInt() != crc) {
+      throw IllegalStateException("CRC Mismatch decoding ${chunkId.decodeToString()}, invalid data")
+    }
+
+    return Header.valueOf(chunkId.decodeToString().uppercase()) to Buffer().apply { write(data) }
+  }
+
+  private fun BufferedSource.peekNextChunkId() =
+    peek().run {
```
swankjesse, line 238: Note to self: make peeks allocate less

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+import java.util.zip.CRC32
+import java.util.zip.Deflater
+
+/**
+ * Provides a basic implementation for encoding APNG's.
+ *
+ * Refs: https://www.w3.org/TR/png/#4Concepts.Encoding, https://wiki.mozilla.org/APNG_Specification
+ */
+internal class ApngWriter(
+  private val file: File,
```
swankjesse, line 51: Same: Remove `File`; add `Path` and `FileSystem`

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+    frameCount++
+  }
+
+  override fun close() {
+    if (frameCount != totalFrameCount) {
+      throw IllegalStateException("Expected $totalFrameCount total frames, actual count $frameCount")
+    }
+
+    sink.writeIEND()
+    fileHandle.resize(fileHandle.position(sink))
```
swankjesse, line 105: Is this necessary?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  /**
+   * The IHDR chunk defines the image size from which all animation frames must be
+   * encapsulated. If over the course of the animation we need to render a larger
+   * image than the first frame we need to resize the first frame such that the
+   * entire animation will fit within. Subsequent animation frames are still correct
+   * and don't need to be re-computed.
+   */
+  private fun correctFirstFrame() {
```
swankjesse, line 124: Could this operate on a FileHandle? Closing the file and opening it again seems needless?

swankjesse: Or is there something about the tempFile I’m missing?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
gamepro65: The PNG header as a string doesn't read well, it isn't just letters/numbers

TWiStErRob: @gamepro65 I think he means this: https://javadoc.io/doc/com.squareup.okio/okio/1.17.4/okio/ByteString.html

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    writeChunk(ACTL, data)
+  }
+
+  private fun BufferedSink.writeFCTL(
+    sequenceNumber: Int,
+    rectangle: Rectangle
+  ) {
+    val data = Buffer().apply {
```
swankjesse, line 182: Creating all of these temporary buffers shouldn’t be necessary? What if we changed the signature of writeChunk to this:

```
  private fun BufferedSink.writeChunk(header: Header, content: BufferedSink.() -> {}) {
```

Then you could rewind to save some copies.

For now as-is is okay!

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,276 @@
...
+          if (raster.numBands == 4) {
+            writeByte(argb ushr 24)
+          }
+        }
+      }
+    }
+  }
+
+  private fun Buffer.compress(level: Int): Buffer {
+    val deflater = Deflater(level).apply {
```
swankjesse, line 245:
```
val result = Buffer()
DeflaterSink(result, Deflater()).use { deflaterSink ->
  deflaterSink.writeAll(this)
}
return result
```

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
gamepro65: Maybe? Will check, I remember in the case that this file already exists... except looking at it now it's the temp file which we cleanup with deleteOnExit so I'm thinking it is a relic of before I added that in

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
gamepro65: I think it already is doing that... `val unstableSource = fileHandle.source(iDATChunkEndOffset)` moves our fileHandle back to it's iDATChunkEndOffset, then we open the real file path into stableHandle which is where we want the bytes to ultimately end up

gamepro65: 🤦  I see what you mean, because we close before calling  this fun.

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
gamepro65: Good call. That would leave the resulting frame's pixel transparent instead of black if first frame didn't contain the pixel and second contained black.

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
gamepro65: Actually I don't think we need to do this, we are doing a direct pixel copy into the resulting image with no scaling or anything that would benefit from hints

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
gamepro65: Still don't think this is right. PNG_SIG is never represented as a ByteString, don't want to convert it just for the != convenience

gamepro65: Pulling from https://github.com/cashapp/paparazzi/pull/1092#discussion_r1326476402 does bring a ByteString into play. 

Wonder if Okio should include a `rangeEquals(Long, ByteArray)` so we wouldn't need to allocate the ByteString

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
```diff
@@ -0,0 +1,264 @@
...
+    val width: Int,
+    val height: Int,
+    val offsetX: Int,
+    val offsetY: Int,
+    val delayNumerator: Short,
+    val delayDenominator: Short
+  )
+
+  companion object {
+    private const val BUFFER_SIZE = 1_024_000
```
JakeWharton, line 262: Is this value dictated by the spec or something? If so, does it have a better name? Not sure it makes sense to be a constant otherwise–it's only used in one place and has no semantic meaning. I would probably put `1000 * 1024` directly at the call-site with a comment.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,277 @@
...
+internal class ApngWriter(
+  private val path: Path,
+  private val totalFrameCount: Int,
+  fps: Int,
+  fileSystem: FileSystem = FileSystem.SYSTEM
+) : Closeable {
+
+  private val fpsNum = 1
+  private val fpsDen = fps
+  private val tempFile = File(FileSystem.SYSTEM_TEMPORARY_DIRECTORY.toFile(), "temp-${path.name}").apply { deleteOnExit() }
```
JakeWharton, line 60: Okio is hurting you here more than helping
```suggestion
  private val tempFile = File.createTempFile(path.name, "apng").apply { deleteOnExit() }
```

JakeWharton: Also can we delete this eagerly in `close()` rather than with `deleteOnExit()`? We shouldn't make assumptions for how long the VM will be alive.

JakeWharton: Also also, have you considered a peer `Path` with `.tmp` or something suffixed? That would allow you keep all writes going through the supplied `FileSystem` rather than needing to rely on the system one (better again for tests).

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * The IHDR chunk defines the image size from which all animation frames must be
+   * encapsulated. If over the course of the animation we need to render a larger
+   * image than the first frame we need to resize the first frame such that the
+   * entire animation will fit within. Subsequent animation frames are still correct
+   * and don't need to be re-computed.
+   */
+  private fun correctFirstFrame() {
+    val resizedFirstFrame = firstFrame.resize(maxWidth, maxHeight)
+    val stableHandle = FileSystem.SYSTEM.openReadWrite(path)
```
JakeWharton, line 128:
```suggestion
    val stableHandle = fileSystem.openReadWrite(path)
```
Also needs the ctor parameter to become a `private val`

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngWriterTest.kt
```diff
@@ -0,0 +1,306 @@
...
+import java.awt.image.BufferedImage
+import java.io.File
+import java.util.zip.CRC32
+import java.util.zip.Inflater
+
+class ApngWriterTest {
+
+  @Test
+  fun writesAnimationMetadata() {
+    val testPath = File.createTempFile("writesAnimationMetadata.png", null).path.toPath()
```
JakeWharton, line 48: Use JUnit's `TemporaryDirectory` rule to create temp files which ensures anything you create gets cleaned up at the end of the test suite.

JakeWharton (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
JakeWharton: https://github.com/square/okio/issues/1346

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngWriter.kt
TWiStErRob: Re the last from @JakeWharton, can you assume that the input path's parent is writable? Isn't the temp a safer bet to create new files?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -312,6 +313,36 @@ internal object ImageUtils {
...
   }
 
+  fun smallestDiffRect(firstImage: BufferedImage, secondImage: BufferedImage): Rectangle? {
+    var rect: Rectangle? = null
+    val width = max(firstImage.width, secondImage.width)
+    val height = max(firstImage.height, secondImage.height)
+
+    for (y in 0 until height) {
+      for (x in 0 until width) {
+        val firstRgb = if (x < firstImage.width && y < firstImage.height) firstImage.getRGB(x, y) else null
```
TWiStErRob, line 323: This looks like per-pixel boxing (allocation) on the happy path.

Could it be done with control flow somehow e.g. `continue`?

TWiStErRob, line 323: Would it worth caching the 2*2 w/h in local variables to prevent function calls?

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
gamepro65: Not dictated by the spec, just an arbitrary size for the Deflater. Will move to call-site

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
TWiStErRob: Sorry, I meant to put `val firstWidth = firstImage.width` outside the loop.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngReader.kt
JakeWharton: Seems like `PNG_SIG` can be a `ByteString` though. Okio wants you to never have to touch a `ByteArray` because it's a mutable type that can never be trusted. Seems like the only uses are `rangeEquals`, `skip` (length), `reposition` (length), and `write` which all either accept `ByteString` or its length.

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
gamepro65: Aha! Makes sense

gamepro65: I think it could be improved upon in that if the running dirty width is equal to max width we only need to check the next row up to the first different pixel to modify the 'bottom' and then can skip the rest of the row since it's already going to be included. But in the happy path you will still be required to check all the pixels to be sure the last pixel didn't suddenly change. Can investigate in a follow up

PR #1093, opened 2023-09-14
Introducing paparazzi annotation processor

nak5ive:


paparazzi-annotation-processor/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=paparazzi-annotation-processor
+POM_NAME=Paparazzi Annotation Processor
+POM_DESCRIPTION=An annotation processor used to generate Paparazzi test data from Composable @Previews
```
jrodbx, line 3:
```suggestion
POM_DESCRIPTION=An annotation processor for integrating Paparazzi with Composable @Previews
```

paparazzi-annotation-processor/build.gradle
```diff
@@ -0,0 +1,6 @@
+apply plugin: 'org.jetbrains.kotlin.jvm'
+apply plugin: 'com.vanniktech.maven.publish'
+
+dependencies {
+  implementation libs.ksp.api
```
jrodbx, line 5:
```suggestion
  implementation libs.ksp
```

gradle/libs.versions.toml
```diff
@@ -38,6 +39,8 @@ jcodec-javase = { module = "org.jcodec:jcodec-javase", version.ref = "jcodec" }
 kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version = "1.7.3" }
 
+ksp-api = { module = "com.google.devtools.ksp:symbol-processing-api", version.ref = "ksp" }
```
jrodbx, line 42:
```suggestion
ksp = { module = "com.google.devtools.ksp:symbol-processing-api", version.ref = "ksp" }
```

jrodbx (approved):

JakeWharton:
I would consider naming this just "processor" or "symbol processor". Annotation processing is specifically a Java thing and will have people thinking you're using Java's annotation processing (supported by kapt).

jrodbx:
> I would consider naming this just "processor" or "symbol processor". Annotation processing is specifically a Java thing and will have people thinking you're using Java's annotation processing (supported by kapt).

Makes sense re: avoiding confusion.  But how about `compiler-plugin` (similar to `poko`) or `preview-processor` (more specific to its use case), instead of `symbol-processor` which on second thought sounds too generic?

JakeWharton:
Poko is an actual compiler plugin so I wouldn't choose that because this is specific to KSP. Just using processor on whatever prefix you want seems fine, too.

PR #1100, opened 2023-09-19
Refactor okio bits of ApngWriter

gamepro65:
Paired with @swankjesse to simplify and take advantage of okio in the writer.

### DIFF
--- a/paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngWriterTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngWriterTest.kt
@@ -50,7 +50,7 @@ class ApngWriterTest {
   @Test
   fun writesAnimationMetadata() {
     val testPath = tempFolderRule.newFile("writesAnimationMetadata.png").path.toPath()
-    ApngWriter(testPath, 3, 1).use { writer ->
+    ApngWriter(testPath, 1).use { writer ->
       writer.writeImage(createImage(squareOffset = Point(5, 5)))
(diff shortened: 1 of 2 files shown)


paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngWriterTest.kt
```diff
@@ -240,14 +240,6 @@ class ApngWriterTest {
     }
   }
 
-  @Test(expected = IllegalStateException::class)
-  fun throwsExceptionWhenWrittenFramesNotEqualTotalFrames() {
-    val testPath = tempFolderRule.newFile("throwsExceptionWhenWrittenFramesNotEqualTotalFrames.png").path.toPath()
-    ApngWriter(testPath, 2, 1).use { writer ->
-      writer.writeImage(createImage(squareOffset = Point(5, 5)))
-    }
-  }
```
jrodbx, line 249: Does deletion of this test imply that this is no longer possible?

jrodbx (approved):

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngWriterTest.kt
gamepro65: Yes, we discussed the up front totalFrameCount being an opinionated API choice specific to Paparazzi, not a strict requirement for the writer itself.

jrodbx: Ok, cool!

PR #1105, opened 2023-09-22
Make AGP and KGP deps compileOnly

ZacSweers:
Dependencies on other gradle plugins should be `compileOnly` dependencies, as it can accidentally impose the version paparazzi depends on rather than just use the one that the consuming project uses. For example – in our project, this transitive dep prevents us from testing newer AGP versions due to multiple versions appearing on the buildscript classpath.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,7 @@ docs/1.x
 docs/changelog.md
 docs/index.md
 site/
+
+# Generated in tests
+/paparazzi-gradle-plugin/src/test/projects/rerun-resource-change/src/main/res/values/colors.xml
+/paparazzi-gradle-plugin/src/test/projects/rerun-asset-change/src/main/assets/secret.txt
--- a/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi-gradle-plugin/build.gradle
@@ -29,6 +29,11 @@ dependencies {
     because "SymbolUtils.getPackageNameFromManifest removed in AGP 7.0.  Replace?"
   }
 
+  testImplementation libs.plugin.kotlin
+  testImplementation libs.plugin.android
+  testImplementation(libs.tools.sdkCommon) {
+    because "SymbolUtils.getPackageNameFromManifest removed in AGP 7.0.  Replace?"
+  }
   testImplementation libs.junit
   testImplementation libs.truth
 }
--- a/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi-gradle-plugin/build.gradle
@@ -20,6 +20,23 @@ gradlePlugin {
   }
 }
 
+// Fix missing implicit task dependency in Gradle's test kit
+tasks.named("processTestResources") { dependsOn("pluginUnderTestMetadata") }
+
+// Fix missing implicit task dependency in Gradle's test kit
+tasks.named("processTestResources") { it.dependsOn("pluginUnderTestMetadata") }
+
+Configuration addTestPlugin = configurations.create("addTestPlugin")
+
+configurations {
+  testImplementation.extendsFrom(addTestPlugin)
+}
+
+tasks.pluginUnderTestMetadata {
+  // make sure the test can access plugins for coordination.
+  pluginClasspath.from(addTestPlugin)
+}
+
 dependencies {
   compileOnly gradleApi()
   implementation platform(libs.kotlin.bom)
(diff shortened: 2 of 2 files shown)


paparazzi-gradle-plugin/build.gradle
```diff
@@ -23,12 +23,17 @@ gradlePlugin {
...
-  implementation(libs.tools.sdkCommon) {
+  compileOnly libs.plugin.kotlin
+  compileOnly libs.plugin.android
+  compileOnly(libs.tools.sdkCommon) {
     because "SymbolUtils.getPackageNameFromManifest removed in AGP 7.0.  Replace?"
   }
 
+  testImplementation libs.plugin.kotlin
+  testImplementation libs.plugin.android
+  testImplementation(libs.tools.sdkCommon) {
```
TWiStErRob, line 34: Were you trying to replace the `runtimeOnly` part of the `implementation` with these to help Gradle TestKit?

<details><summary>If so, I hacked around this area of Gradle before, and got it working.</summary>

A likely cause for the [CI failure](https://github.com/cashapp/paparazzi/actions/runs/6278471146/job/17052299239?pr=1105#step:5:17129) is `PluginUnderTestMetadata.getPluginClasspath`, because the [default only uses production runtime classpath](https://github.com/gradle/gradle/blob/v8.3.0/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L203).

Here's my solution: [definition](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/e132379a7b51ad0ac2bef5e7f7df3767fca7fb92/gradle/plugins/src/main/kotlin/net/twisterrob/gradle/build/testing/net.twisterrob.gradle.build.testing.plugin-under-test-metadata-extras.gradle.kts#L64-L70) and [usage](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/e132379a7b51ad0ac2bef5e7f7df3767fca7fb92/plugin/building/build.gradle.kts#L34-L39).  
(If you use this, you could probably ignore my version constraint usage, I needed it because I have AGP/KGP in the test matrix.)

</details>

ZacSweers: The above should work more or less without any hacks, not sure why CI is failing. WIll look more later

jrodbx:
> Dependencies on other gradle plugins should be `compileOnly` dependencies, as it can accidentally impose the version paparazzi depends on rather than just use the one that the consuming project uses. For example – in our project, this transitive dep prevents us from testing newer AGP versions due to multiple versions appearing on the buildscript classpath.

Is the newer plugin version check a new thing?  We've definitely been able to pull in a newer AGP version on CashApp, with the risk of binary incompatibilities resulting in Paparazzi crashes, but that's different from your error message.

Re: the premise of `compileOnly`, I've previously attempted such (even consulting @autonomousapps given his success in having a test matrix in [dependency-analysis-gradle-plugin](https://github.com/autonomousapps/dependency-analysis-gradle-plugin)), but @JakeWharton and I were of the opinion that it probably wasn't worth the maintenance burden, given AGP's dev cycle.   Not opposed to it, if successful though.

jrodbx:
accidentally hit "Close with comment", just meant to "Comment".  My bad.

ZacSweers:
> Is the newer plugin version check a new thing?

It's not a check, it's just how gradle and classloading work. It's a first-one-wins policy in many cases due to the hierarchical nature of gradle and generally just good plugin hygiene to not impose other plugin versions transitively. Things get tricky when you start assuming things about versions when a dependency appears sooner in that hierarchy (i.e. such as settings.gradle classpath) and later projects assume their dependency will win or be resolved with others.

.gitignore
```diff
@@ -18,3 +18,7 @@ docs/1.x
 docs/changelog.md
 docs/index.md
 site/
+
+# Generated in tests
+/paparazzi-gradle-plugin/src/test/projects/rerun-resource-change/src/main/res/values/colors.xml
+/paparazzi-gradle-plugin/src/test/projects/rerun-asset-change/src/main/assets/secret.txt
```
ZacSweers, line 24: These were getting generated as untracked files when I ran tests locally, so opportunistically adding them here

paparazzi-gradle-plugin/build.gradle
```diff
@@ -20,15 +20,37 @@ gradlePlugin {
...
+Configuration addTestPlugin = configurations.create("addTestPlugin")
+
+configurations {
+  testImplementation.extendsFrom(addTestPlugin)
+}
+
+tasks.pluginUnderTestMetadata {
+  // make sure the test can access plugins for coordination.
+  pluginClasspath.from(addTestPlugin)
+}
```
ZacSweers, line 38: This pattern is what we use in Keeper to test against non-bundled KGP/AGP deps

PR #1106, opened 2023-09-23
Update compile sdk version to 34

FogelAO:


jrodbx:
Thanks!

Unfortunately, it looks like the earliest version of LayoutLib to support this will be 2023.2.1 (Iguana): 

- https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+log/ae1f91d3aad81e796fa894ff8c48a2f7d7d9bab2
- https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/33e6573aee8b717d2970138ca5dfd3d342c31d40

FogelAO:
@jrodbx Thank you for your response! How could I update it in my project? Because I have an error when update target sdk to 34

Issue #595, opened 2022-10-05
Snapshots should be placed into `src/{sourceSetName}/snapshots/` for MPP plugin


JakeWharton:
Snapshots seem to be unconditionally placed in `src/test/snapshots/`. For Kotlin MPP, the name of the Android-specific source set is arbitrary and that name should be used as the source set folder so that the `snapshots/` directly is placed relative to the tests which produced the snapshots.

You can see the undesired default behavior in action here: https://github.com/cashapp/redwood/tree/0f708ea3cdd950a427dde46547c4ce67ba01a60c/redwood-layout/src

colinmarsch:
Did https://github.com/cashapp/paparazzi/pull/2115 resolve this?

oldergod:
Fixed with #2115, yes

PR #1109, opened 2023-09-28
Clean up Kotlin plugin checks

jrodbx:
Prep for https://github.com/cashapp/paparazzi/issues/595

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -37,6 +37,7 @@ import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.FileCollection
 import org.gradle.api.internal.artifacts.transform.UnzipTransform
+import org.gradle.api.logging.LogLevel.INFO
 import org.gradle.api.logging.LogLevel.LIFECYCLE
 import org.gradle.api.plugins.JavaBasePlugin
 import org.gradle.api.provider.Provider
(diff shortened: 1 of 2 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -180,26 +182,24 @@ class PaparazziPlugin : Plugin<Project> {
...
-        project.tasks.named("compile${testVariantSlug}JavaWithJavac")
-          .configure { it.dependsOn(writeResourcesTask) }
+        testVariant.javaCompileProvider.configure { it.dependsOn(writeResourcesTask) }
       }
 
-      project.plugins.withType(KotlinMultiplatformPluginWrapper::class.java) {
-        val multiplatformExtension =
-          project.extensions.getByType(KotlinMultiplatformExtension::class.java)
-        check(multiplatformExtension.targets.any { target -> target is KotlinAndroidTarget }) {
+      if (project.plugins.hasPlugin(KotlinMultiplatformPluginWrapper::class.java)) {
```
TWiStErRob, line 189: Will this create a plugin ordering requirement between Paparazzi and KGP? Or is it handled somehow?

JakeWharton: Yes. Should still use a with function to configure as before. If you need to detect to case of no Kotlin plugin you can set a var to true when a plugin is detected and check it was set to true in afterEvaluate

JakeWharton (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Similar to this: https://github.com/cashapp/paparazzi/pull/1111#issuecomment-1739779868

I'm not quite satisfied with the execution flow here, given how the snapshotSourceSet is either determined via async callbacks (making it challenging to have all the logic in the same place) or sync hasPlugin calls (introducing a plugin ordering problem).  Will probably need to refactor the plugin code more to make this easier.

PR #1111, opened 2023-09-28
Vary snapshot output dir based on applied kotlin plugin

jrodbx:
Closes https://github.com/cashapp/paparazzi/issues/595


jrodbx:
Closing, to think about this a little more.  

Moving away from `withType` callbacks results in spaghetti code around determining `snapshotSourceSet` depending on which plugin is applied.  Using `hasPlugin` synchronous code causes plugin ordering issues.

I think the plugin code might need some more refactoring to better enable this.

jrodbx:
question for @JakeWharton @TWiStErRob: should we require a kotlin plugin to be added?  I think right now, we haven't required it, but it might be antiquated to support this?

JakeWharton:
Not a huge fan of forcing it when it's not strictly required. There's absolutely people still writing Java that will use Paparazzi.

TWiStErRob:
Agreed. Support it, but don't require it.
What would be the reason to require it?

I think requiring near-latest AGP is already a higher than necessary barrier to entry, but it's a good tradeoff for maintenance and moving fast.

autonomousapps:
I'm still strongly in favor of using `compileOnly` for dependencies on other plugins. I note that this plugin has an `implementation` dependency in KGP; is that necessary? That'll put KGP on the build classpath even if the project using Paparazzi doesn't use Kotlin.

The other issue with having strong dependencies on other plugins is the way Gradle's classloader hierarchy works. Depending on where this plugin gets added to a build, it can complicate things for users who want to use different versions of the other plugins.

Issue #964, opened 2023-06-29
Delta images break on dimension change with RenderingMode.SHRINK


ko-bra:
**Description**
When using RenderingMode.SHRINK: If the dimensions of a recorded image during verify are different from the dimensions of the golden image, the delta image may lack: 

- the diff image and/or
- parts of the actual and expected image.

**Steps to Reproduce**
1. Set the rendering mode to `RenderingMode.SHRINK`
2. Set up a test and `record` the golden image for a component
3. Change the test so that either
3.1  the width of the component is reduced, or
3.2 the height of a component is changed.
5. Run a `verify`

**Expected behavior**
The produced delta image file should contain
- the expected image
- the diff image
- the actual image

All of the above should be rendered in full size and be visible.

**Observed behavior**
- On width reduction: The _expected image_ overrides the _delta image_.
- On height change: Either the _actual_ or the _expected_ image is cut off in height.

**Potential cause**
See this excerpt of the function [assertImageSimilar(...)](https://github.com/cashapp/paparazzi/blob/8f69445ff1b157cab6bbc6202598dd99a2782b04/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt#L98) of `ImageUtils`:

```
val imageWidth = Math.min(goldenImage.width, image.width)
val imageHeight = Math.min(goldenImage.height, image.height)
val width = 3 * imageWidth
val deltaImage = BufferedImage(width, imageHeight, TYPE_INT_ARGB)
```
After drawing the delta image in the middle, the expected and actual images are drawn:
```
g.drawImage(goldenImage, 0, 0, null)
g.drawImage(image, 2 * imageWidth, 0, null)
```
Due to tripling the width of the smaller _actual image_, the larger _expected image_ does not fit in its "slot" on the left and will overwrite the delta "slot".
On a height change, one of the images will be cropped because the height is set to the minimum height between the two images.

**Additional information:**
- Paparazzi Version: 1.2
- OS: macOS 13.1 (22C65) aarch64
- Compile SDK: 31
- Gradle Version: 7.5.1
- Android Gradle Plugin Version: 7.4.0

jrodbx:
Fixed in #1122

PR #1122, opened 2023-10-02
Update delta images to support showing diff when width and height differ

adamalyyan:
This address an issue raised in https://github.com/cashapp/paparazzi/issues/964 when using `RenderingMode.SHRINK` where the delta image doesn't show the correct delta and/or is missing part of the expected/actual image. 

What's happening is the calculation for the output of the delta image takes the min size between the expected and actual and uses this as the width/height for all 3 components (the expected, actual and delta images). This results in the delta not showing the correct delta or the expected/actual image not being complete when you run verify and the view has changed in one of more dimensions.

This PR addresses this issue by ensuring the delta image is a max of the expected/actual image so the full delta can be displayed and uses this calculation to ensure the final composite image doesn't have anything cut off as showed in the examples below.

**Golden Image:**
golden-image

--------

**Comparison (changing width of expected image):**
_Before_: 
before-verify-against-more-width
_After_:
verify-against-more-width

----------

**Comparison (changing height of expected image):**
_Before_:
before-verify-against-more-height
_After_:
verify-against-more-height

-----------

**Comparison (changing width & height of expected image):**
_Before_:
before-verify-against-more-width-and-height
_After_:
verify-against-more-width-and-height

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
@@ -114,33 +114,48 @@ internal object ImageUtils {
     }
     assertEquals(TYPE_INT_ARGB.toLong(), goldenImage.type.toLong())
 
-    val imageWidth = Math.min(goldenImage.width, image.width)
-    val imageHeight = Math.min(goldenImage.height, image.height)
+    val goldenImageWidth = goldenImage.width
+    val goldenImageHeight = goldenImage.height
+
+    val imageWidth = image.width
+    val imageHeight = image.height
+
+    val deltaWidth = Math.max(goldenImageWidth, imageWidth)
+    val deltaHeight = Math.max(goldenImageHeight, imageHeight)
 
     // Blur the images to account for the scenarios where there are pixel
     // differences
     // in where a sharp edge occurs
     // goldenImage = blur(goldenImage, 6);
     // image = blur(image, 6);
-
-    val width = 3 * imageWidth
-    val deltaImage = BufferedImage(width, imageHeight, TYPE_INT_ARGB)
+    val width = goldenImageWidth + deltaWidth + imageWidth
+    val deltaImage = BufferedImage(width, deltaHeight, TYPE_INT_ARGB)
     val g = deltaImage.graphics
 
     // Compute delta map
     var delta: Long = 0
-    for (y in 0 until imageHeight) {
-      for (x in 0 until imageWidth) {
-        val goldenRgb = goldenImage.getRGB(x, y)
-        val rgb = image.getRGB(x, y)
+    for (y in 0 until deltaHeight) {
+      for (x in 0 until deltaWidth) {
+        val goldenRgb = if (x >= goldenImageWidth || y >= goldenImageHeight) {
+          0x00808080
+        } else {
+          goldenImage.getRGB(x, y)
+        }
+
+        val rgb = if (x >= imageWidth || y >= imageHeight) {
+          0x00808080
+        } else {
+          image.getRGB(x, y)
+        }
+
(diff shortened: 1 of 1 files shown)


geoff-powell (approved):
LGTM, appreciate the snapshots in the description 🙏🏽

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -114,33 +114,45 @@ internal object ImageUtils {
...
       for (x in 0 until imageWidth) {
-        val goldenRgb = goldenImage.getRGB(x, y)
-        val rgb = image.getRGB(x, y)
+        val goldenRgb = if (x >= goldenImageWidth || y >= goldenImageHeight) {
+          0x00808080
+        } else {
+          goldenImage.getRGB(x, y)
+        }
+
+        val rgb = if (x >= image.width || y >= image.height) {
```
TWiStErRob, line 142: Is there a reason for treating `image.width` and `goldenImageWidth` differently? I mean the fact that one is cached in a `val` and the other isn't.

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -114,33 +114,45 @@ internal object ImageUtils {
     }
     assertEquals(TYPE_INT_ARGB.toLong(), goldenImage.type.toLong())
 
-    val imageWidth = Math.min(goldenImage.width, image.width)
-    val imageHeight = Math.min(goldenImage.height, image.height)
+    val goldenImageWidth = goldenImage.width
+    val goldenImageHeight = goldenImage.height
+
+    val imageWidth = Math.max(goldenImageWidth, image.width)
+    val imageHeight = Math.max(goldenImageHeight, image.height)
```
TWiStErRob, line 121: Might it be clearer to name these differently? e.g.:
```suggestion
    val deltaWidth = Math.max(goldenImageWidth, image.width)
    val deltaHeight = Math.max(goldenImageHeight, image.height)
```

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -114,33 +114,45 @@ internal object ImageUtils {
...
 
     // Blur the images to account for the scenarios where there are pixel
     // differences
     // in where a sharp edge occurs
     // goldenImage = blur(goldenImage, 6);
     // image = blur(image, 6);
-
-    val width = 3 * imageWidth
+    val width = goldenImageWidth + imageWidth + image.width
```
TWiStErRob, line 128: Did you also try what happens if the image gets smaller on either/both dimensions?

adamalyyan: Yeah, in that scenario it basically looks like the images in the description except the goldenImage is larger and the image on the right is smaller

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
TWiStErRob: Yep, I mean that you could
* either inline `goldenImageWidth`
* or extract an `actualImageWidth` variable

It looks mixed up with using both approaches for very similar things.

adamalyyan: Ah gotcha, yeah that's a good point, the only reason goldenImageWidth and goldenImageHeight are in a val is because they are also used in other calculations outside the loop, but definitely makes sense to be consistent, will update this

jrodbx (approved):
lgtm!

PR #1136, opened 2023-10-10
Migrate to AGP 8 APIs and off of legacy variants API

ZacSweers:
This migrates the plugin to newer AGP 8.0 APIs and using `AndroidComponents` APIs. This offers a few benefits:
- Lazier evaluation across the board
- Depending on artifact outputs rather than tasks

Note that `MergeResources` doesn't yet have a public artifact, but does have an internal one. Since Paparazzi uses multiple other internal artifact APIs, I figured this was fair game to use.

I did leave one extra note as well - there's a `compile*WithJavac` task gated on a BaseJavaPlugin block, but this task is actually created by KGP. Didn't adjust it in this PR to reduce scope, but FYI.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -113,9 +113,9 @@ class PaparazziPlugin : Plugin<Project> {
       it.description = "Record golden images for all variants"
     }
 
-    variants.all { variant ->
+    variants.configureEach { variant ->
       val variantSlug = variant.name.capitalize(Locale.US)
-      val testVariant = variant.unitTestVariant ?: return@all
+      val testVariant = variant.unitTestVariant ?: return@configureEach
 
       val mergeResourcesOutputDir = variant.mergeResourcesProvider.flatMap { it.outputDir }
       val mergeAssetsProvider =
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -30,6 +30,7 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.DomainObjectSet
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE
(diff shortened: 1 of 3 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -113,154 +112,193 @@ class PaparazziPlugin : Plugin<Project> {
...
-          // They need special handling, so they're added as inputs.property above, and systemProperty here.
-          test.systemProperties["paparazzi.platform.data.root"] =
-            nativePlatformFileCollection.singleFile.absolutePath
-          test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
-          test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
-        }
+    val testTaskName = "test$testVariantSlug"
+    project.tasks.matching {
+      it is Test && it.name == testTaskName
+    }.configureEach { rawTest ->
```
TWiStErRob, line 258: Similar consistency as above:
```suggestion
    project.tasks
      .matching { it is Test && it.name == testTaskName }
      .configureEach { rawTest ->
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -113,154 +112,193 @@ class PaparazziPlugin : Plugin<Project> {
...
+      val test = rawTest as Test
```
TWiStErRob, line 259: Did you know this also works? (yay Smart Casts!)
```suggestion
    }.configureEach { test ->
      test as Test
```
It might be simpler, but maybe too magical?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -829,7 +831,7 @@ class PaparazziPluginTest {
     assertThat(resourceFileContents[1]).isEqualTo("intermediates/merged_res/debug")
     assertThat(resourceFileContents[4]).isEqualTo("intermediates/assets/debug")
     assertThat(resourceFileContents[5]).isEqualTo("app.cash.paparazzi.plugin.test,com.example.mylibrary,app.cash.paparazzi.plugin.test.module1,app.cash.paparazzi.plugin.test.module2")
-    assertThat(resourceFileContents[6]).isEqualTo("src/main/res,src/debug/res")
+    assertThat(resourceFileContents[6]).isEqualTo("build/generated/res/resValues/debug,src/debug/res,src/main/res")
```
TWiStErRob, line 834: mind the order of `debug` vs `main`: https://github.com/cashapp/paparazzi/pull/980/files#r1254387829

It's important because of [variant folder precedence](https://developer.android.com/build/build-variants#sourceset-build), I think.

Related (depending on where the reversal happens): https://github.com/cashapp/paparazzi/pull/980#discussion_r1254159048

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -86,20 +88,17 @@ class PaparazziPlugin : Plugin<Project> {
...
       supportedPlugins.forEach { plugin ->
         project.plugins.withId(plugin) {
-          val variants = when (val extension = project.extensions.getByType(TestedExtension::class.java)) {
-            is LibraryExtension -> extension.libraryVariants
-            is BaseAppModuleExtension -> extension.applicationVariants
-            is DynamicFeatureExtension -> extension.applicationVariants
-            // exhaustive to avoid potential breaking changes in future AGP releases
-            else -> throw IllegalStateException("${extension.javaClass.name} from $plugin is not supported in Paparazzi")
+          val extension = project.extensions.getByType(AndroidComponentsExtension::class.java)
+          check(extension is LibraryAndroidComponentsExtension || extension is ApplicationAndroidComponentsExtension || extension is DynamicFeatureAndroidComponentsExtension) {
```
TWiStErRob, line 92: Is this still necessary?

Is there maybe a one-to-one mapping between plugin ID and `androidComponents { }` type we can rely on?

TWiStErRob (commented):
Really nice, I'm glad you made it work cleaner than our previous try. It's always nice to learn something new from PRs.

Tip for future reviewers: use IDEA diff with "Ignore whitespaces", GitHub mostly red/green because of indent changes.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -113,154 +112,193 @@ class PaparazziPlugin : Plugin<Project> {
...
-          .configure { it.dependsOn(writeResourcesTask) }
+    project.plugins.withType(KotlinMultiplatformPluginWrapper::class.java) {
+      val multiplatformExtension =
+        project.extensions.getByType(KotlinMultiplatformExtension::class.java)
+      check(multiplatformExtension.targets.any { target -> target is KotlinAndroidTarget }) {
+        "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin"
       }
+      project.tasks
+        .matching { it.name == "compile${testVariantSlug}KotlinAndroid" }
+        .configureEach { it.dependsOn(writeResourcesTask) }
```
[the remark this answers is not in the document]
ZacSweers: yep it's a lazier approach and a little more correct, as afterEvaluate doesn't necessarily guarantee it since there can be multiple afterEvaluate levels

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -113,154 +112,193 @@ class PaparazziPlugin : Plugin<Project> {
...
-        val multiplatformExtension =
-          project.extensions.getByType(KotlinMultiplatformExtension::class.java)
-        check(multiplatformExtension.targets.any { target -> target is KotlinAndroidTarget }) {
-          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin"
+    // TODO this task isn't actually produced by the JavaBasePlugin, this is a Kotlin task
+    project.plugins.withType(JavaBasePlugin::class.java) {
+      project.tasks
+        .matching {
+          it.name == "compile${testVariantSlug}JavaWithJavac"
         }
```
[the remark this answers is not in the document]
ZacSweers: I just ran spotless and checked in what it output. Not going to deviate from that

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
ZacSweers: Same answer as above

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
ZacSweers: I prefer the practical approach

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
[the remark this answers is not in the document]
TWiStErRob: Huh, interesting...

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: Sorry, I didn't include enough context in the highlight:
```kotlin
val supportedPlugins = listOf("com.android.application", "com.android.library", "com.android.dynamic-feature")
...
supportedPlugins.forEach { plugin ->
    project.plugins.withId(plugin) {
        val extension = project.extensions.getByType(AndroidComponentsExtension::class.java)
        // At this point we can be quite sure extension is not `TestAndroidComponentsExtension`,
        // because that would imply someone applied `com.android.test` plugin,
        // which is mutually exclusive to all the others in `supportedPlugins`.
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
TWiStErRob: @kevinzheng-ap, can you please explain in why the order matters (if it does)?

(From the earlier PR I understood it's because Paparazzi/layoutlib is emulating a full-on Android build/runtime, so the folder order determines which resources are loaded at runtime. Consider having a same named resource in `src/main/res/values/x.xml` and `src/debug/res/values/x.xml` and what'll be rendered on the screenshot vs on device.)

> not something we can change anyway

It's up to Paparazzi how it interprets and uses the AGP APIs.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -113,154 +112,193 @@ class PaparazziPlugin : Plugin<Project> {
...
+    verifyVariants: TaskProvider<Task>,
+    nativePlatformFileCollection: FileCollection
+  ) {
+    val variantSlug = variant.name.capitalize(Locale.US)
+    val testVariant = if (variant is HasUnitTest) {
+      @Suppress("DEPRECATION") // This API isn't actually un-deprecated in any subtypes
+      variant.unitTest ?: return
+    } else {
+      return
+    }
```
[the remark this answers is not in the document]
ZacSweers: ah you're right. Not taking the suggestion as I had a change locally but updating with it 👍

jrodbx:
@ZacSweers mind: 1) rebasing and 2) inlining configureVariant for a better diff before an initial review?

PR #1137, opened 2023-10-11
Only print report path if it was generated

### LINKED ISSUE #1135: Allow not logging the paparazzi report
When applying the paparazzi plugin, it logs the following at the end of every test run:
```
See the Paparazzi report at: ...
```

This is confusing for our devs because we conditionally run our screenshot tests so the referenced file often doesn't exist. It would be great if that message was either removed or made optional in some way.

ansman:
Paparazzi currently assumes a report will always be written which isn't the case if you're verifying or if you use a custom snapshot handler. With this change the logging only happens if a report was written.

This fixes #1135

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -251,11 +251,14 @@ class PaparazziPlugin : Plugin<Project> {
             nativePlatformFileCollection.singleFile.absolutePath
           test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
           test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+          reportOutputDir.get().asFile.deleteRecursively()
         }
 
         test.doLast {
-          val uri = reportOutputDir.get().asFile.toPath().resolve("index.html").toUri()
-          test.logger.log(LIFECYCLE, "See the Paparazzi report at: $uri")
+          val report = reportOutputDir.get().asFile.resolve("index.html")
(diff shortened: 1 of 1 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -251,11 +251,14 @@ class PaparazziPlugin : Plugin<Project> {
             nativePlatformFileCollection.singleFile.absolutePath
           test.systemProperties["paparazzi.test.record"] = isRecordRun.get()
           test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()
+          reportOutputDir.get().asFile.deleteRecursively()
         }
 
         test.doLast {
-          val uri = reportOutputDir.get().asFile.toPath().resolve("index.html").toUri()
-          test.logger.log(LIFECYCLE, "See the Paparazzi report at: $uri")
+          val report = reportOutputDir.get().asFile.resolve("index.html")
```
TWiStErRob, line 258: It should keep using the path API. Otherwise the link will become unclickable in IDEA.

```
println(file("build.gradle").toURI())
// output: file:/T:/xx/build.gradle
println(file("build.gradle").toPath().toUri())
// output: file:///T:/xx/build.gradle
```

image

ansman: Ah, great callout. I did not realize it mattered how we got a an `URI` instance. Will fix.

geoff-powell (approved):

jrodbx:
We're going to rework this mechanism and have a solution soon, hopefully in the next release.

PR #1162, opened 2023-11-21
fix macos architecture check

### LINKED ISSUE #1031: recordPaparazzi on Java 17 - UninitializedPropertyAccessException lateinit property sessionParamsBuilder
**Description**
Error when running paparazzi record since updating to gradle 8 / Java 17
```
 kotlin.UninitializedPropertyAccessException: lateinit property sessionParamsBuilder has not been initialized
        at app.cash.paparazzi.Paparazzi$Companion.getSessionParamsBuilder$paparazzi(Paparazzi.kt:626)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:162)
        at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:124)
        at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
```

have added toolchains workaround
```
tasks.withType(Test).configureEach {
    javaLauncher = javaToolchains.launcherFor {
        languageVersion = JavaLanguageVersion.of(11)
    }
}
```

**Steps to Reproduce**
Unable to reproduce in sample project, but will keep trying over the next few nights.

**Expected behavior**
Expect to be able to run record tasks as before.

**Additional information:**
- Paparazzi Version: 1.3.1 / 1.3.2-snapshot
- OS: macOS 13.5 (aarch64)
- Compile SDK:
- Gradle Version: 33
- Android Gradle Plugin Version: 8.0.2

Note: Problem doesn't appear on our CI builds, on Linux

evant:
use /usr/bin/arch to detect what architecture we are actually running on instead of what arch the jre was built for. This fixes the case of using an intel version of the jvm on arm.

Fixes #1031


evant:
note: hasn't been fully tested but wanted to make sure the approach made sense first

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
```diff
@@ -188,12 +189,28 @@ internal class Renderer(
...
-    val osName = System.getProperty("os.name").toLowerCase(Locale.US)
+    val osName = System.getProperty("os.name").lowercase(Locale.US)
     val osLabel = when {
       osName.startsWith("windows") -> "win"
       osName.startsWith("mac") -> {
-        val osArch = System.getProperty("os.arch").lowercase(Locale.US)
-        if (osArch.startsWith("x86")) "mac" else "mac-arm"
+        // System.getProperty("os.arch") returns the os of the jre, not necessarily of the platform we are running on,
+        // try /usr/bin/arch to get the actual architecture
+        val osArch = ProcessBuilder("/usr/bin/arch")
```
jrodbx, line 198: this method call runs on every Paparazzi test method run, so I suspect this change would introduce a slight perf regression due to frequent shelling out.

is it fair to assume from the tagged issue that the current check would be OK if and only if the proper JDK had been installed?

if so, then I'd propose moving this `/usr/bin/arch` logic to `checkInstalledJvm`, changing it to a "valid JDK for OS" check, and guarding that method with a static sentinel (since we shouldn't need to check the installed jvm on every method run either).  thoughts?

jrodbx (commented):
Thanks for the PR.  Please see suggested feedback above.

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
evant: > is it fair to assume from the tagged issue that the current check would be OK if and only if the proper JDK had been installed?

If you are not running in Rosetta then the original check did work, this pr came about because our work policy locks down what versions of the jdk we can install and they currently don't have the arm version available.

> since we shouldn't need to check the installed jvm on every method run either

That is correct, it should only need to be checked once.

evant: Updated to only check once, and it prints a warning if the jre and os architectures don't match

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
```diff
@@ -188,16 +188,7 @@ internal class Renderer(
...
-    val osLabel = when {
-      osName.startsWith("windows") -> "win"
-      osName.startsWith("mac") -> {
-        val osArch = System.getProperty("os.arch").lowercase(Locale.US)
-        if (osArch.startsWith("x86")) "mac" else "mac-arm"
-      }
-      else -> "linux"
-    }
-    return "$osLabel/lib64"
+    return "$OsLabel/lib64"
```
jrodbx, line 191:
```suggestion
    return "$osLabel/lib64"
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -110,3 +111,33 @@ private fun checkInstalledJvm() {
     )
   }
 }
+
+internal val OsLabel: String by lazy {
```
jrodbx, line 115:
```suggestion
internal val osLabel: String by lazy {
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -110,3 +111,33 @@ private fun checkInstalledJvm() {
...
+    osName.startsWith("mac") -> {
+      // System.getProperty("os.arch") returns the os of the jre, not necessarily of the platform we are running on,
+      // try /usr/bin/arch to get the actual architecture
+      val osArch = ProcessBuilder("/usr/bin/arch")
+        .start()
+        .inputStream
+        .bufferedReader()
+        .lines()
+        .findFirst()
+        .getOrNull()
```
jrodbx, line 128:
```suggestion
        .readLine()
```

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -110,3 +111,33 @@ private fun checkInstalledJvm() {
...
+        ?.lowercase(Locale.US)
```
jrodbx, line 129: while this input stream _could_ return null, is that likely in practice?

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -110,3 +111,33 @@ private fun checkInstalledJvm() {
...
+        ?.let {
+          if (it.startsWith("i386")) "mac" else "mac-arm"
+        }
+      val jreArch = System.getProperty("os.arch").lowercase(Locale.US).let {
+        if (it.startsWith("x86")) "mac" else "mac-arm"
+      }
+      if (osArch != null && osArch != jreArch) {
+        System.err.println("Mismatched architectures, os is:$osArch but the JRE is running with:$jreArch. Installing the correct JDK for your architecture will improve performance.")
+      }
+      osArch ?: jreArch
```
jrodbx, line 139:
```suggestion
            when (osArch) {
                null -> jreArch
                jreArch -> osArch
                else -> throw RuntimeException("Mismatched architectures, os is:$osArch but the JRE is running with:$jreArch. Please installing the correct JDK for your architecture.")
            } 
```

I propose making this fail fast/crash, given the error in #1031 seemed to make this a blocker for others who accidentally installed the wrong JDK.

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
```diff
@@ -188,16 +188,7 @@ internal class Renderer(
   }
 
   private fun getNativeLibDir(): String {
-    val osName = System.getProperty("os.name").toLowerCase(Locale.US)
```
jrodbx, line 191: to my earlier feedback, can we revert this method/Renderer, and instead move the osArch check to `Environment` under `checkInstalledJvm`?  My rationale is that it seems more appropriate to have this logic live in Environment, and leaving Renderer to assume things are set up correctly.

evant: my hope with this was to not just give a better error message but to actually make it work

jrodbx: I get that, but I think trying to make it work can lead to other weird behaviors if the native lib artifact downloaded via dependency resolution isn't the expected one.

jrodbx: thoughts?

evant: > can lead to other weird behaviors
Yeah fair, was trying to spend some time verifying if it would actually work or not, but hadn't reproduced the original issue, at least by running the tests in the repo. Will try to spend more time on if if I can, otherwise, not against your suggestion as a backup, at least makes the issue more clear.

PR #1163, opened 2023-11-26
Prevent running merge resources unless legacy mode is enabled

ansman:
There seems to have been several issues with legacy mode:
- There was no gradle property for enabling asset legacy mode
- Enabling legacy resource mode did not set the test system property meaning you'd have to set that yourself
- Disabling legacy mode still made the test task depend on the merge resources task

Since the whole point of the new resource loader was to not depend on the merge task this seems quite a serious issue. In our project this increased our build cache size by about 6GB across 82 modules.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -42,7 +42,6 @@ import org.gradle.api.plugins.JavaBasePlugin
 import org.gradle.api.provider.Provider
 import org.gradle.api.reporting.ReportingExtension
 import org.gradle.api.tasks.PathSensitivity
-import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.options.Option
 import org.gradle.api.tasks.testing.Test
 import org.gradle.internal.os.OperatingSystem
@@ -58,8 +57,9 @@ import kotlin.io.path.invariantSeparatorsPathString
 class PaparazziPlugin : Plugin<Project> {
   override fun apply(project: Project) {
     val legacyResourceLoadingEnabled = (project.findProperty("app.cash.paparazzi.legacy.resource.loading") as? String)?.toBoolean() ?: false
+    val legacyAssetLoadingEnabled = (project.findProperty("app.cash.paparazzi.legacy.asset.loading") as? String)?.toBoolean() ?: false
 
-    if (legacyResourceLoadingEnabled) {
+    if (legacyResourceLoadingEnabled || legacyAssetLoadingEnabled) {
       project.afterEvaluate {
         check(!project.plugins.hasPlugin("com.android.application")) {
           error(
@@ -74,7 +74,7 @@ class PaparazziPlugin : Plugin<Project> {
       }
 
       project.plugins.withId("com.android.library") {
-        setupPaparazzi(project, project.extensions.getByType(LibraryExtension::class.java).libraryVariants)
+        setupPaparazzi(project, project.extensions.getByType(LibraryExtension::class.java).libraryVariants, legacyResourceLoadingEnabled, legacyAssetLoadingEnabled)
       }
     } else {
       val supportedPlugins = listOf("com.android.application", "com.android.library", "com.android.dynamic-feature")
@@ -93,13 +93,18 @@ class PaparazziPlugin : Plugin<Project> {
             // exhaustive to avoid potential breaking changes in future AGP releases
             else -> throw IllegalStateException("${extension.javaClass.name} from $plugin is not supported in Paparazzi")
           }
-          setupPaparazzi(project, variants)
+          setupPaparazzi(project, variants, legacyResourceLoadingEnabled, legacyAssetLoadingEnabled)
         }
       }
     }
   }
 
-  private fun <T> setupPaparazzi(project: Project, variants: DomainObjectSet<T>) where T : BaseVariant, T : TestedVariant {
+  private fun <T> setupPaparazzi(
+    project: Project,
+    variants: DomainObjectSet<T>,
+    legacyResourceLoadingEnabled: Boolean,
+    legacyAssetLoadingEnabled: Boolean
+  ) where T : BaseVariant, T : TestedVariant {
     project.addTestDependency()
     val nativePlatformFileCollection = project.setupNativePlatformDependency()
 
@@ -119,7 +124,7 @@ class PaparazziPlugin : Plugin<Project> {
 
       val mergeResourcesOutputDir = variant.mergeResourcesProvider.flatMap { it.outputDir }
       val mergeAssetsProvider =
-        project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
+        project.tasks.named("merge${variantSlug}Assets", MergeSourceSetFolders::class.java)
       val mergeAssetsOutputDir = mergeAssetsProvider.flatMap { it.outputDir }
       val projectDirectory = project.layout.projectDirectory
       val buildDirectory = project.layout.buildDirectory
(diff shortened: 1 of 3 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -58,8 +57,9 @@ import kotlin.io.path.invariantSeparatorsPathString
 class PaparazziPlugin : Plugin<Project> {
   override fun apply(project: Project) {
     val legacyResourceLoadingEnabled = (project.findProperty("app.cash.paparazzi.legacy.resource.loading") as? String)?.toBoolean() ?: false
+    val legacyAssetLoadingEnabled = (project.findProperty("app.cash.paparazzi.legacy.asset.loading") as? String)?.toBoolean() ?: false
 
-    if (legacyResourceLoadingEnabled) {
+    if (legacyResourceLoadingEnabled || legacyAssetLoadingEnabled) {
```
ansman, line 62: I'm assuming we want to do an `||` rather than `&&` here. But honestly I'm not sure why we need this if-statement at all.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -230,13 +239,19 @@ class PaparazziPlugin : Plugin<Project> {
         test.systemProperties["paparazzi.snapshot.dir"] = snapshotOutputDir.toString()
         test.systemProperties["paparazzi.artifacts.cache.dir"] = gradleUserHomeDir.path
         test.systemProperties["kotlinx.coroutines.main.delay"] = true
+        test.systemProperties["app.cash.paparazzi.legacy.resource.loading"] = legacyResourceLoadingEnabled.toString()
+        test.systemProperties["app.cash.paparazzi.legacy.asset.loading"] = legacyAssetLoadingEnabled.toString()
```
ansman, line 243: I'm assuming that not setting these was an omission

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -44,34 +46,35 @@ abstract class PrepareResourcesTask : DefaultTask() {
   abstract val compileSdkVersion: Property<String>
 
   @get:InputFiles
-  @get:PathSensitive(PathSensitivity.NONE)
+  @get:PathSensitive(PathSensitivity.RELATIVE)
```
ansman, line 49: Without this, the debug and release prepare tasks have the same cache key so `mergeDebugResources` can accidentally get the cache from `mergeReleaseResources`

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1396,6 +1403,11 @@ class PaparazziPluginTest {
     var generatedSettings = false
     var generatedGradleProperties = false
 
+    File(projectRoot, "build-cache").delete()
+    projectRoot.walkTopDown()
+      .filter { it.name == "build" && it.isDirectory }
+      .forEach { it.deleteRecursively() }
+
```
ansman, line 1410: Without this I had super flaky tests.

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/src/debug/res/raw/debug.txt
ansman, line 1: We needed some resources to make sure we got different cache keys for debug and release.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: interesting, originally this if/else block was designed with only resources in mind, so I missed updating this with assets.

that being said, if one were to run legacy resources more, but modern assets mode, I'm not sure how that would behave...

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: can you elaborate some more, as I haven't noticed this in my local dev

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/src/debug/res/raw/debug.txt
jrodbx: this feels like a solution to a different problem, and therefore better extracted into a separate PR?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Can you elaborate more?  Renderer appears to be able to see these system properties without this.

ansman: That's very very strange, I'm not sure how it can. If you search the project for `app.cash.paparazzi.legacy.asset.loading` there is only a single result (the constant) and that constant is only used in a single place too.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
ansman: It would get issues like tasks being up-to-date when running the test the second time since the build directory already existed.

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/src/debug/res/raw/debug.txt
ansman: It's not really. Without these, the resource merging task will fail because either the debug or release task will be up-to-date since they have the same inputs and thus the cached result of one is used for the other. I don't see another way to easily make that test pass.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -119,7 +124,7 @@ class PaparazziPlugin : Plugin<Project> {
 
       val mergeResourcesOutputDir = variant.mergeResourcesProvider.flatMap { it.outputDir }
       val mergeAssetsProvider =
-        project.tasks.named("merge${variantSlug}Assets") as TaskProvider<MergeSourceSetFolders>
+        project.tasks.named("merge${variantSlug}Assets", MergeSourceSetFolders::class.java)
```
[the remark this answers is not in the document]
ansman: Mostly that I believe the error message will be better (you'll get some cryptic error with the unchecked cast where as the other one actually gives a proper error)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: There is a reason given how we validate which modules are supported.  Legacy mode only support library modules, whereas modern supports application/dynamic features.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
TWiStErRob: ... https://github.com/cashapp/paparazzi/pull/695

jrodbx:
Superseded by https://github.com/cashapp/paparazzi/pull/1235.

Issue #1101, opened 2023-09-20
Paparazzi is causing coroutine delays to stop working


saket:
This sounds bizarre but `Paparazzi` is affecting how coroutine delays are run. This test never finishes running after printing `"delaying…"`:

```kotlin
class PaparazziTest {
  @get:Rule val paparazzi = Paparazzi()

  @Test fun `delay issue`() = runBlocking {
    withContext(Dispatchers.Default) {
      println("Delaying on ${Thread.currentThread().name}…")
      delay(1.seconds)
      println("Delay done")
    }
  }
}
```

Commenting out the 2nd line where `Paparazzi` is instantiated fixes the problem. Any ideas what might be happening?

saket:
> Paparazzi internally controls time as well as calling Handler_Delegate.executeCallbacks() which gives the Handler a chance to process anything that's been queued

Sorry I forgot to mention one important detail in my original post: this issue with delays are also happening in unrelated tests that do not use paparazzi. Is it possible that paparazzi isn't cleaning up its hooks inside handlers?

saket:
@kevincianfarini has pointed out that `runBlocking` can also be used instead which supplies its own delay capable event loop.

saket:
I filed a feature request to unset `DefaultDelay` to potentially address this: https://youtrack.jetbrains.com/issue/KT-62098/Provide-a-way-to-unset-kotlinx.coroutines.DefaultDelay

PR #1164, opened 2023-11-30
Remove reliance on kotlinx.coroutines.main.delay

gamepro65:
Forcing delay on main was convenient however undesirable in that it effects all tests not just Paparazzi tests. By overriding the WindowRecomposerFactory that ComposeUi uses we can inject our own Recomposer and give it an appropriate CoroutineContext which includes delay on main.

Resolves:  https://github.com/cashapp/paparazzi/issues/1101

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -32,7 +32,11 @@ import android.view.ViewGroup
 import androidx.activity.setViewTreeOnBackPressedDispatcherOwner
 import androidx.annotation.LayoutRes
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.compose.ui.InternalComposeUiApi
 import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.WindowRecomposerPolicy
+import androidx.compose.ui.platform.createLifecycleAwareWindowRecomposer
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.setViewTreeLifecycleOwner
 import androidx.savedstate.setViewTreeSavedStateRegistryOwner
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -289,6 +296,11 @@ class Paparazzi @JvmOverloads constructor(
           // CompositionContext, which requires first finding the "content view", then using that
           // to find a root view with a ViewTreeLifecycleOwner
           viewGroup.id = android.R.id.content
+
+          // ComposeUi uses it's own `AndroidUiDispatcher` by default. Unfortunately this dispatcher
+          // does not implement delay which causes any delay call to run async with the test. By
+          // initializing the Recomposer with Dispatchers.Main delay will be backed by our test Handler.
+          WindowRecomposerPolicy.setFactory { it.createLifecycleAwareWindowRecomposer(Dispatchers.Main) }
```
saket, line 303: Does it matter if we're using `Main` vs `Main.immediate`?

saket (approved):
This is fantastic!

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -305,6 +303,18 @@ class Paparazzi @JvmOverloads constructor(
...
+        if (hasComposeRuntime) {
+          // During onAttachedToWindow, AbstractComposeView will attempt to resolve its parent's
+          // CompositionContext, which requires first finding the "content view", then using that
+          // to find a root view with a ViewTreeLifecycleOwner
+          viewGroup.id = android.R.id.content
+
+          // ComposeUi uses it's own `AndroidUiDispatcher` by default. Unfortunately this dispatcher
+          // does not implement delay which causes ComposeUi to use DefaultDelay which will be run
+          // async to the test. By initializing the Recomposer with Dispatchers.Main delay will be backed by our test Handler.
+          modifiedView.compositionContext = modifiedView.createLifecycleAwareWindowRecomposer(Dispatchers.Main)
```
gamepro65, line 315: `createLifecycleAwareWindowRecomposer` requires LifecycleOwner to be already installed so moving this hasComposeRuntime block below that setup.

paparazzi-gradle-plugin/src/test/projects/lifecycle-usages/build.gradle
```diff
@@ -20,6 +20,6 @@ android {
 }
 
 dependencies {
-  testImplementation 'androidx.activity:activity-compose:1.6.1'
+  testImplementation libs.androidx.appcompat
```
gamepro65, line 23: Brings in some compose but not all, `createLifecycleAwareWindowRecomposer` does not make it. These tests are not compose related, we have a separate compose lifecycle test, switching to appcompat since it covers these lifecycle tests.

jrodbx (approved):
nice work!  commented offline about comment cleanup, but otherwise, LFG!

PR #1030, opened 2023-07-31
Add docs about jacoco compatibility

tasomaniac:
I was looking into the below issue as I also hit that in the project that I was integrating Paparazzi. A potential fix for this could be more complex than needed. As I could find a much simpler workaround, my suggestion would be to close this issue by adding this as documentation in the README.

Let me know what you think.

Closes #955

Issue #979, opened 2023-07-04
Cannot record Paparazzi test due to UnsupportedOperationException


tassilo-posegga:
**Description**
Since a couple of months my colleagues complained that they are not able to record any Paparazzi tests due to an exception which I couldn't reproduce at that time and we also couldn't spot any differences in setup. Now is the time where the same issue popped up on my machine:

```
StayLoggedInScreenTest > [isAccessible=false] > testLoadingStayLoggedInContent[isAccessible=false] FAILED
    java.lang.UnsupportedOperationException at null:-2
```
Here is the full stack trace:
https://pastebin.com/dv7uH5jg

**Steps to Reproduce**
Record any Paparazzi test using :recordPaparazzi gradle command

**Expected behavior**
No Exception :)

**Additional information:**
- Paparazzi Version: 1.2.0
- OS: MacOS (M1) & Windows 11
- Compile SDK: 34
- Gradle Version: 8.1.1 (but was happening with 7.X as well)
- Android Gradle Plugin Version: 7.4.2

TWiStErRob:
Please try to upgrade to Paparazzi 1.3.0, as this has been considered fixed in #384.

Just to confirm: are you using Java 17 and jacoco?
If the answers are yes and you get it after upgrading to 1.3 too, then this is likely a duplicate of #955.

tassilo-posegga:
We couldn't upgrade to AGP 8.X yet - so we currently still run Java 11. We do have jacoco though
AFAIK we need to update AGP before upadting Paparazzi to 1.3.0

TWiStErRob:
I think it might be possible to 1.3.0 with AGP 7.4 on Gradle 7.6/8, because the AGP 8 bump happened late during Paparazzi 1.3, see the commits that happened after the bump: https://github.com/cashapp/paparazzi/compare/3d401cff34d0515507c857799c4996db1e16cd59...1.3.0, at a quick glance I don't see any breaking changes re AGP 7/8.

java target is still 11, so I think you can try this way:
 * update paparazzi 1.2.0 to 1.3.0
 * exclude transitive AGP dependencies on `buildscript > classpath` or `buildSrc`/included build that come from the `paparazzi-gradle-plugin` to prevent upgrading AGP too
 * define `android.nonTransitiveRClass` in `gradle.properties` explicitly to help paparazzi know the value (the default assumption changed in AGP 8)

jrodbx:
[...]

Issue #955, opened 2023-06-23
Java 17 + test coverage + JaCoCo incompatibility - UnsupportedOperationException: class redefinition failed: attempted to delete a method


loutry:
I'm currently using Paparazzi successfully but run into an issue while upgrading my project as follow:
* AGP from 7.4.2 to 8.0.2 
* Java from 11 to 17
* Kotlin from 1.8.10 to 1.8.21
* Compose Compiler from 1.4.3 to 1.4.7
* Compose from 1.3.1 to 1.4.3
* Paparazzi Gradle Plugin from 1.2.0 to 1.3.0

All tests using Paparazzi are failing with the following error:
```
class redefinition failed: attempted to delete a method
java.lang.UnsupportedOperationException: class redefinition failed: attempted to delete a method
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses0(Native Method)
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at net.bytebuddy.utility.Invoker$Dispatcher.invoke(Unknown Source)
	at net.bytebuddy.utility.dispatcher.JavaDispatcher$Dispatcher$ForNonStaticMethod.invoke(JavaDispatcher.java:1032)
	at net.bytebuddy.utility.dispatcher.JavaDispatcher$ProxiedInvocationHandler.invoke(JavaDispatcher.java:1162)
	at net.bytebuddy.dynamic.loading.$Proxy42.retransformClasses(Unknown Source)
	at net.bytebuddy.dynamic.loading.ClassReloadingStrategy$Strategy$2.apply(ClassReloadingStrategy.java:399)
	at net.bytebuddy.dynamic.loading.ClassReloadingStrategy.load(ClassReloadingStrategy.java:227)
	at net.bytebuddy.dynamic.TypeResolutionStrategy$Passive.initialize(TypeResolutionStrategy.java:101)
	at net.bytebuddy.dynamic.DynamicType$Default$Unloaded.load(DynamicType.java:6317)
[...]

PR #1167, opened 2023-12-08
Use a class file locator that queries the system class loader

jrodbx:
Paparazzi's current usage of ByteBuddy conflicts with Jacoco and other bytecode-transformation libraries.

Relevant issue: https://github.com/raphw/byte-buddy/issues/1248

This PR slightly changes this from:

```
      var builder = byteBuddy
        .redefine(foo.Bar::class.java)
```

which calls:
```
    public <T> DynamicType.Builder<T> redefine(Class<T> type) {
        return redefine(type, ClassFileLocator.ForClassLoader.of(type.getClassLoader()));
    }

    public <T> DynamicType.Builder<T> redefine(Class<T> type, ClassFileLocator classFileLocator) {
        return redefine(TypeDescription.ForLoadedType.of(type), classFileLocator);
    }
```

to:
```
      var builder = byteBuddy
        .redefine(TypePool.Default.ofSystemLoader().describe("foo.Bar").resolve(), ClassFileLocator.ForClassLoader.ofSystemLoader())
```

Per https://github.com/raphw/byte-buddy/issues/1248#issuecomment-1121619704, it appears this allows Bytebuddy to 
find the manipulated class file in-memory vs loading the original class definition from disk.

See also "Working with unloaded classes" in https://bytebuddy.net/#/tutorial.

Closes https://github.com/cashapp/paparazzi/issues/979.
Closes https://github.com/cashapp/paparazzi/issues/955.
Closes https://github.com/cashapp/paparazzi/pull/1030.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -1387,6 +1387,18 @@ class PaparazziPluginTest {
     )
   }
 
+  @Test
+  fun jacoco() {
+    val fixtureRoot = File("src/test/projects/jacoco")
+
+    gradleRunner
+      .withArguments("testDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    val jacocoExecutionData = File(fixtureRoot, "build/jacoco/testDebugUnitTest.exec")
+    assertThat(jacocoExecutionData.exists()).isTrue()
+  }
+
   private fun GradleRunner.runFixture(
     projectRoot: File,
     action: GradleRunner.() -> BuildResult
--- a/paparazzi-gradle-plugin/src/test/projects/jacoco/build.gradle
+++ b/paparazzi-gradle-plugin/src/test/projects/jacoco/build.gradle
@@ -1,11 +1,12 @@
-apply plugin: 'com.android.library'
-apply plugin: 'org.jetbrains.kotlin.android'
-apply plugin: 'app.cash.paparazzi'
-apply plugin: 'jacoco'
+plugins {
+  id 'com.android.library'
+  id 'kotlin-android'
+  id 'app.cash.paparazzi'
+  id 'jacoco'
+}
 
 android {
-  namespace 'app.cash.paparazzi.sample'
-
+  namespace 'app.cash.paparazzi.plugin.test'
   compileSdk libs.versions.compileSdk.get() as int
   defaultConfig {
     minSdk libs.versions.minSdk.get() as int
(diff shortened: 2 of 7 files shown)


sample/build.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.library'
 apply plugin: 'org.jetbrains.kotlin.android'
 apply plugin: 'app.cash.paparazzi'
+apply plugin: 'jacoco'
```
TWiStErRob, line 4: Is there an automated regression test that can be added?

jrodbx: My thinking is applying the jacoco plugin is that regression test, since adding that one liner demonstrates the UnsupportedOperationException.  What else did you have in mind?

jrodbx: added a plugin integration test, PTAL

TWiStErRob: Nice!

Is it worth keeping this line in sample?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1387,6 +1387,18 @@ class PaparazziPluginTest {
     )
   }
 
+  @Test
+  fun jacoco() {
+    val fixtureRoot = File("src/test/projects/jacoco")
+
+    gradleRunner
+      .withArguments("testDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
```
TWiStErRob, line 1396: I guess there's no way to ensure that this is running on Java 17. It's just the way things are because of CI/AGP.

sample/build.gradle
TWiStErRob: (can't approve the PR)

jrodbx: > Is it worth keeping this line in sample?

Let's leave it.  It still shows things are working by not failing as previously and we can build on this for CI checks in the future.

TWiStErRob: @jrodbx this causes crazy spamming while running tests in :sample, repro: `gradlew :sample:test`

```
SEVERE: Unknown type (class [Z) in R class field: private static transient boolean[] app.cash.paparazzi.sample.R$array.$jacocoData
```

CI logs are 5x bigger than before. 200k to 1M, 2500 lines to 9000 lines.
https://github.com/cashapp/paparazzi/actions/runs/7151841385/job/19476541218#step:5:257

Related: https://github.com/cashapp/paparazzi/issues/732

TWiStErRob: This is so annoying, even `spotlessApply` runs these tests...

markuskallander: Is there a solution to this available? This happens in my project as well when I run unit tests...

TWiStErRob: @markuskallander https://github.com/cashapp/paparazzi/issues/732#issuecomment-1490528929 is a workaround.

ashley-figueira: @TWiStErRob the suggested workaround does not work. The fix would be to ignore synthetic fields as suggested here https://stackoverflow.com/a/52856548

PR #1186, opened 2023-12-11
Support Paparazzi#gif with APNGs

gamepro65:
Handles all the different error configurations that can happen during gif verification. By default the verifier is only used for gif, potentially could be used for all snapshots but wanted to take it one step at a time. General performance testing a base gif during recordPaparazzi averaged ~35ms per frame on an M1 Pro with verifyPaparazzi delta video (which is 3x as wide) taking upward of ~130ms per frame.

Refactored the hard coded failure dir from ImageUtils so we can seed it properly. For testing the delta image generation we opt out of rendering java text (IE. Expected / Actual) because it is different between platforms and would require a non-zero % diff image assertion.

Mismatched frames:
delta-app cash paparazzi sample_KeypadViewTest_testViews_spin

More frames than golden:
delta-app cash paparazzi sample_KeypadViewTest_testViews_spin

Less frames than golden:
delta-app cash paparazzi sample_KeypadViewTest_testViews_spin

Mismatched fps:
delta-app cash paparazzi sample_KeypadViewTest_testViews_spin


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -931,6 +930,18 @@ class PaparazziPluginTest {
...
+  @Test
+  fun verifyGif() {
+    val fixtureRoot = File("src/test/projects/verify-gif")
+
+    val result = gradleRunner
+      .withArguments("verifyPaparazzi", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    assertThat(result.task(":preparePaparazziDebugResources")).isNotNull()
+    assertThat(result.task(":testDebugUnitTest")).isNotNull()
```
jrodbx, line 942: can/should we do more in this assertion?  like verify file paths and assert a "apng" exists?

saket:
HYPE!

saket:
Will this be compatible with composables?

gamepro65:
> Will this be compatible with composables?

Yup! Works as expected out of the box 🙌

oldergod:
What's blocking this?

geoff-powell (approved):
This PR looks good to me. Anything blocking this from getting merged? cc @gamepro65

BrianGardnerAtl (approved):
Looks good. Are you planning on adding the `HtmlReportWriter` test for the `.apng` file check?

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
oldergod, line 1: Great test @gamepro65

jrodbx (commented):
Partial feedback; will give another pass, but wanted to get this out to start the conversation.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1294,14 +1305,11 @@ class PaparazziPluginTest {
...
-    val snapshotFile = File(snapshotsDir, "9d3c31a9c79a363c26dc352b59e9e77083c300a7.png")
+    val snapshotFile = File(snapshotsDir, "a2095a4d9f78335277b48a4bda336c1437d44295.png")
     assertThat(snapshotFile.exists()).isTrue()
 
     val goldenImage = File(fixtureRoot, "src/test/resources/arrow_missing.png")
-    val actualFileBytes = Files.readAllBytes(snapshotFile.toPath())
-    val expectedFileBytes = Files.readAllBytes(goldenImage.toPath())
-
-    assertThat(actualFileBytes).isEqualTo(expectedFileBytes)
+    assertThat(goldenImage).isSimilarTo(snapshotFile).withDefaultThreshold()
```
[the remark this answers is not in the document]
gamepro65: The file changing names is still going to be a thing in this PR, don't think its worth splitting out just for this one threshold change

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -89,60 +89,34 @@ public class HtmlReportWriter @JvmOverloads constructor(
...
+        file.renameTo(hashedFile)
+        file.delete()
+
+        if (isRecording) {
+          val goldenFile = File(goldenDir, snapshot.toFileName("_", "png"))
+          hashedFile.copyTo(target = goldenFile, overwrite = true)
         }
 
-        shots += shot
+        shots += snapshot.copy(file = hashedFile.toJsonPath())
```
jrodbx, line 115: I really like how much simpler this reads!

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -89,60 +89,34 @@ public class HtmlReportWriter @JvmOverloads constructor(
...
-            val goldenFile = File(goldenVideosDirectory, snapshot.toFileName("_", "mov"))
-            if (!goldenFile.exists()) {
-              original.copyTo(goldenFile)
-            }
-          }
-          snapshot.copy(file = original.toJsonPath())
+        writer.close()
+        val hashedFile = File(snapshotDir, "${hash(hashes)}.png")
+        file.renameTo(hashedFile)
+        file.delete()
```
jrodbx, line 108: since `file` is already managed by writer, should this renameTo/delete logic be moved into the `close` implementation?  would clean this up a little more, but can also see not wanting to bake that behavior into the ApngWriter; just a thought.

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -89,60 +89,34 @@ public class HtmlReportWriter @JvmOverloads constructor(
     fps: Int
   ): FrameHandler {
     return object : FrameHandler {
+      val snapshotDir = if (fps == -1) imagesDirectory else videosDirectory
+      val goldenDir = if (fps == -1) goldenImagesDirectory else goldenVideosDirectory
       val hashes = mutableListOf<String>()
+      val file = File(snapshotDir, snapshot.toFileName(extension = "temp.png"))
```
jrodbx, line 95:
```suggestion
      val snapshotTmpFile = File(snapshotDir, snapshot.toFileName(extension = "temp.png"))
```

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -89,60 +89,34 @@ public class HtmlReportWriter @JvmOverloads constructor(
...
-          val original = File(videosDirectory, "$hash.mov")
-          if (isRecording) {
-            val goldenFile = File(goldenVideosDirectory, snapshot.toFileName("_", "mov"))
-            if (!goldenFile.exists()) {
-              original.copyTo(goldenFile)
-            }
-          }
-          snapshot.copy(file = original.toJsonPath())
+        writer.close()
+        val hashedFile = File(snapshotDir, "${hash(hashes)}.png")
```
jrodbx, line 106:
```suggestion
        val snapshotFile = File(snapshotDir, "${hash(hashes)}.png")
```

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -36,7 +36,8 @@ class HtmlReportWriterTest {
   val snapshotRoot: TemporaryFolder = TemporaryFolder()
 
   private val anyImage = BufferedImage(1, 1, BufferedImage.TYPE_INT_ARGB)
-  private val anyImageHash = "9069ca78e7450a285173431b3e52c5c25299e473"
+  private val anyImageHash = "2ef6a6e795a2b8cffabefb6d9a2066a183c8e3b6"
```
jrodbx, line 39: any idea why `anyImageHash` changes?

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -36,7 +36,8 @@ class HtmlReportWriterTest {
...
+  private val anyVideoHash = "7325996fabd92c7411c0d03ed62641323ffdf0d1"
 
   @Test
   fun happyPath() {
```
jrodbx, line 43:
```suggestion
  fun happyPathImages() {
```

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -113,48 +114,156 @@ class HtmlReportWriterTest {
 
   @Test
   fun alwaysOverwriteOnRecord() {
```
jrodbx, line 116:
```suggestion
  fun imagesAlwaysOverwriteOnRecord() {
```

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -113,48 +114,156 @@ class HtmlReportWriterTest {
...
-        timestamp = now.toDate()
+      val frameHandler = htmlReportWriter.newFrameHandler(
+        Snapshot(
+          name = "loading",
+          testName = TestName("app.cash.paparazzi", "CelebrityTest", "testSettings"),
+          timestamp = Instant.parse("2019-03-20T10:27:43Z").toDate(),
+          tags = listOf("redesign")
+        ),
+        2,
+        1
```
jrodbx, line 179: let's use named arguments here to make the test more readable (same in the predecessor, `happyPathImages`)

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
```diff
@@ -0,0 +1,221 @@
...
+import kotlin.math.min
+
+internal class ApngVerifier(
+  goldenFilePath: Path,
+  private val deltaFilePath: Path,
+  private val fps: Int,
+  private val frameCount: Int,
+  private val maxPercentDifference: Double,
+  private val fileSystem: FileSystem = FileSystem.SYSTEM,
+  private val withErrorText: Boolean = true
```
jrodbx, line 36: it looks like this is only set explicitly to false in tests.

1) is there a potential other use case for disabling?
2) if not, why not remove the field and always print the text?  (guess: hard to get consistent test successes over all platforms?)

jrodbx (commented):
Some feedback on the record/report mode; will give one last pass for verification.

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
gamepro65: The latter because the text is jdk rendered we get different default font's and subtle rendering differences even when trying to force the same font on different platforms, would have to fallback to a % diff test instead of exact image match which didn't feel great

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
gamepro65: Before snapshot specified the image hash `val original = File(imagesDirectory, "${hashes[0]}.png")` but now both snapshots and gifs share a common path / hashing in close `val hashedFile = File(snapshotDir, "${hash(hashes)}.png")`

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -89,60 +89,34 @@ public class HtmlReportWriter @JvmOverloads constructor(
...
+      val snapshotDir = if (fps == -1) imagesDirectory else videosDirectory
+      val goldenDir = if (fps == -1) goldenImagesDirectory else goldenVideosDirectory
       val hashes = mutableListOf<String>()
+      val file = File(snapshotDir, snapshot.toFileName(extension = "temp.png"))
+      val writer = ApngWriter(file.path.toPath(), fps)
 
       override fun handle(image: BufferedImage) {
-        hashes += writeImage(image)
+        hashes += hash(image)
+        writer.writeImage(image)
```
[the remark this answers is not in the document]
gamepro65: This is just keeping track of all the hashes the handler has seen, we don't use the value until we are saving when the frame handler is closed.

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
gamepro65: I don't think so, this is a unique to HtmlReportWriter not knowing the file name until the last image has been hashed.

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngReaderTest.kt
```diff
@@ -37,7 +37,13 @@ class ApngReaderTest {
       val expectedFile = javaClass.classLoader.getResource("simple_animation_$i.png")
       val expectedImage = ImageIO.read(expectedFile)
       val actualImage = reader.readNextFrame()!!
-      ImageUtils.assertImageSimilar(expectedFile!!.path, expectedImage, actualImage, 0.0)
+      ImageUtils.assertImageSimilar(
+        expectedFile!!.path,
+        File(file.path).parentFile,
```
jrodbx, line 42: if the test fails, will this directory even be writable?  if so, will it be cleaned up?

let's pass a temporary folder instead?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -51,18 +51,10 @@ internal object ImageUtils {
...
   private const val THUMBNAIL_SIZE = 1000
 
-  /** Directory where to write the thumbnails and deltas. */
-  private val failureDir: File
-    get() {
-      val buildDirString = System.getProperty("paparazzi.build.dir")
-      val failureDir = File(buildDirString, "paparazzi/failures")
-      failureDir.mkdirs()
-      return failureDir
-    }
```
jrodbx, line 61: let's refactor this in a preceding separate PR?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -98,10 +91,53 @@ internal object ImageUtils {
   @Throws(IOException::class)
   fun assertImageSimilar(
     relativePath: String,
+    failureDir: File,
```
jrodbx, line 94: nit: move to last argument

also, this would move to separate PR

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -125,6 +125,7 @@ class AccessibilityRenderExtensionTest {
           val expected = File("src/test/resources/${snapshot.name}.png")
           ImageUtils.assertImageSimilar(
             relativePath = expected.path,
+            failureDir = expected.parentFile,
```
jrodbx, line 128: similar to previous comment:

if the test fails, will this directory even be writable? if so, will it be cleaned up?  let's pass a temporary folder instead?

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
jrodbx: nvm, moving this and previous refactoring snippets out into: https://github.com/cashapp/paparazzi/pull/1377

jrodbx (approved):
awesome stuff, @gamepro65!

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngVerifierTest.kt
```diff
@@ -0,0 +1,274 @@
...
+import okio.Path.Companion.toPath
+import org.junit.Assert.fail
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.awt.Point
+import java.io.File
+import java.lang.AssertionError
+
+class ApngVerifierTest {
```
jrodbx, line 18: I really like this test

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
```diff
@@ -0,0 +1,221 @@
...
+    commonFrameRate = leastCommonMultiple(fps, pngReader.getFps())
+    actualDeltasPerFrame = commonFrameRate / fps
+    expectedDeltasPerFrame = commonFrameRate / pngReader.getFps()
+    if (frameCount != pngReader.frameCount || fps != pngReader.getFps()) {
+      deltaWriter = ApngWriter(deltaFilePath, commonFrameRate, fileSystem)
+    }
+  }
+
+  fun verifyFrame(image: BufferedImage) {
+    val (expectedFrame, actualFrame) = resize(currentGoldenFrame ?: blankFrame, image)
```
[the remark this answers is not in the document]
gamepro65: Since this falls into failure / error territory I think it's better to keep the generation lazy.

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -98,10 +91,53 @@ internal object ImageUtils {
...
+      error = saveImageAndAppendMessage(image, error, relativePath, failureDir)
+      println(error)
+      fail(error)
+    }
+  }
+
+  @Throws(IOException::class)
+  fun compareImages(
+    goldenImage: BufferedImage,
+    image: BufferedImage,
```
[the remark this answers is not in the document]
gamepro65: I'm going to leave it for now because `assertImageSimilar` uses the same naming structure, can follow up to align both

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
```diff
@@ -0,0 +1,221 @@
...
+  private val deltaFilePath: Path,
+  private val fps: Int,
+  private val frameCount: Int,
+  private val maxPercentDifference: Double,
+  private val fileSystem: FileSystem = FileSystem.SYSTEM,
+  private val withErrorText: Boolean = true
+) : Closeable {
+  private val pngReader = ApngReader(fileSystem.openReadOnly(goldenFilePath))
+  private val blankFrame by lazy { createBlankFrame(pngReader.width, pngReader.height) }
+  private val invalidFrames = mutableListOf<Float>()
```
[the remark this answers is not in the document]
gamepro65: I think the original intent was to use that to try and say % video error by averaging at the end but didn't pan out. Will make it a count for now

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
```diff
@@ -0,0 +1,221 @@
...
+          goldenImage = nextFrame,
+          image = nextFrame,
+          withText = withErrorText
+        )
+
+        writeImage(deltaImage)
+      }
+      readNextFrame()
+    }
+  }
```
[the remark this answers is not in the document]
gamepro65: It seems like the hard part is discerning reader from writer. I think this would strike a happy medium

```
  private fun ApngReader.initializeWriter(): ApngWriter {
    val currentFrameNumber = frameNumber - 1
    reset()

    return ApngWriter(deltaFilePath, commonFrameRate, fileSystem).apply {
      repeat(currentFrameNumber) {
        val nextFrame = readNextFrame() ?: blankFrame
        val (deltaImage) = ImageUtils.compareImages(
          expected = nextFrame,
          actual = nextFrame,
          withText = withErrorText
        )

        writeImage(deltaImage)
      }
      readNextFrame()
    }
  }
  ```

Issue #906, opened 2023-05-30
Truth 1.1.3 to 1.1.4 upgrade broke Paparazzi


TWiStErRob:
**Description**
Truth patch update caused missing method in Paparazzi. This is mostly an FYI, as I don't think there's much you can do about this bar adding a strict clause to the published metadata.

```
java.lang.NoSuchMethodError: 'java.util.stream.Collector com.google.common.collect.Sets.toImmutableEnumSet()'
	at com.android.resources.ResourceType.<clinit>(ResourceType.java:175)
	at app.cash.paparazzi.internal.DynamicResourceIdManager$IdProvider.<init>(DynamicResourceIdManager.kt:14)
	at app.cash.paparazzi.internal.DynamicResourceIdManager.<init>(DynamicResourceIdManager.kt:28)
	at app.cash.paparazzi.internal.PaparazziCallback.<init>(PaparazziCallback.kt:52)
	at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:145)
	at app.cash.paparazzi.Paparazzi$apply$statement$1.evaluate(Paparazzi.kt:116)
	at app.cash.paparazzi.agent.AgentTestRule$apply$1.evaluate(AgentTestRule.kt:17)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at com.google.testing.junit.testparameterinjector.PluggableTestRunner$ContextMethodRule$1.evaluate(PluggableTestRunner.java:429)
	... org.junit internals
	... gradle running `:test` task in a worker
```

**Steps to Reproduce**
See https://github.com/TWiStErRob/net.twisterrob.sun/pull/256

**Expected behavior**
Just works.

**Additional information:**
- Paparazzi Version: 1.2.0
- OS: N/A
- Compile SDK: 33
- Gradle Version: 8.2
- Android Gradle Plugin Version: 8.0.2

**Screenshots**
![image](https://github.com/cashapp/paparazzi/assets/2906988/c74c8df1-e0e8-4845-8497-88f5ef0c1d87)

TWiStErRob:
More investigation:

<details><summary>Here's a full dependency dump and diff, if anyone's interested.</summary>

[before.txt](https://github.com/cashapp/paparazzi/files/11605257/before.txt)
[after.txt](https://github.com/cashapp/paparazzi/files/11605256/after.txt)
[...]

PR #1206, opened 2024-01-05
Bump Truth 1.1.5 => 1.2.0

jrodbx:
Relates to https://github.com/cashapp/paparazzi/issues/906.  Closes https://github.com/cashapp/paparazzi/pull/1195.

Guava's publishing changed slightly in v33.0.0, using Gradle Module Metadata to help guide consumers' resolution of the appropriate variant, i.e., -jre vs -android: https://github.com/google/guava/pull/3683

Truth 1.2.0 bumps to Guava 33.0.0, so the preceding PR was failing because that new publishing mechanism by default forces the -android variant, which doesn't have the methods we expect and that are present in the -jre variant.

This is a frustrating ongoing problem and this PR is only solving the most recent manifestation of this issue. See the discussion in https://github.com/google/guava/issues/6801 for the latest discussion.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/projects/build-class-next-sdk/src/test/java/app/cash/paparazzi/plugin/test/BuildClassTest.kt
+++ b/paparazzi-gradle-plugin/src/test/projects/build-class-next-sdk/src/test/java/app/cash/paparazzi/plugin/test/BuildClassTest.kt
@@ -27,8 +27,8 @@ class BuildClassTest {
 
   @Test
   fun verifyFields() {
-    assertThat(Build.ID).isEqualTo("URD9.230712.002.A1")
-    assertThat(Build.DISPLAY).isEqualTo("sdk_phone_armv7-userdebug UpsideDownCakePrivacySandbox URD9.230712.002.A1 10628788 test-keys")
+    assertThat(Build.DISPLAY).isEqualTo("sdk_phone_armv7-userdebug UpsideDownCakePrivacySandbox URD9.231106.004.A2 11164158 test-keys")
+    assertThat(Build.ID).isEqualTo("URD9.231106.004.A2")
     assertThat(Build.PRODUCT).isEqualTo("unknown")
     assertThat(Build.DEVICE).isEqualTo("generic")
     assertThat(Build.BOARD).isEqualTo("unknown")
@@ -43,7 +43,7 @@ class BuildClassTest {
     assertThat(Build.SKU).isEqualTo("unknown")
     assertThat(Build.ODM_SKU).isEqualTo("unknown")
 
-    assertThat(Build.VERSION.INCREMENTAL).isEqualTo("10628788")
+    assertThat(Build.VERSION.INCREMENTAL).isEqualTo("11164158")
     assertThat(Build.VERSION.RELEASE).isNotNull()
     assertThat(Build.VERSION.RELEASE_OR_CODENAME).isNotNull()
     assertThat(Build.VERSION.BASE_OS).isEqualTo("")
--- a/paparazzi-gradle-plugin/src/test/projects/common.gradle
+++ b/paparazzi-gradle-plugin/src/test/projects/common.gradle
@@ -1,0 +1,20 @@
+// TODO:
+//   Remove when https://github.com/google/guava/issues/6567 is fixed.
+//   See also: https://github.com/google/guava/issues/6801.
+plugins.withId("app.cash.paparazzi") {
+  // Defer until afterEvaluate so that testImplementation is created by Android plugin.
+  afterEvaluate {
+    dependencies.constraints {
+      add("testImplementation", "com.google.guava:guava") {
+        attributes {
+          attribute(
+            TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE,
+            objects.named(TargetJvmEnvironment, TargetJvmEnvironment.STANDARD_JVM)
+          )
+        }
+        because("LayoutLib and sdk-common depend on Guava's -jre published variant." +
+          "See https://github.com/cashapp/paparazzi/issues/906.")
+      }
+    }
+  }
+}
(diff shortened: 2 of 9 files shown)


paparazzi-gradle-plugin/src/test/projects/build-class-next-sdk/src/test/java/app/cash/paparazzi/plugin/test/BuildClassTest.kt
```diff
@@ -27,8 +27,8 @@ class BuildClassTest {
 
   @Test
   fun verifyFields() {
-    assertThat(Build.ID).isEqualTo("URD9.230712.002.A1")
-    assertThat(Build.DISPLAY).isEqualTo("sdk_phone_armv7-userdebug UpsideDownCakePrivacySandbox URD9.230712.002.A1 10628788 test-keys")
+    assertThat(Build.DISPLAY).isEqualTo("sdk_phone_armv7-userdebug UpsideDownCakePrivacySandbox URD9.231106.004.A2 11164158 test-keys")
+    assertThat(Build.ID).isEqualTo("URD9.231106.004.A2")
```
TWiStErRob, line 31: How did truth/guava affect these?

jrodbx: good point, i think just non-hermetic update from Google's side.  I can move into a separate PR.

paparazzi-gradle-plugin/src/test/projects/common.gradle
```diff
@@ -0,0 +1,20 @@
...
+      add("testImplementation", "com.google.guava:guava") {
+        attributes {
+          attribute(
+            TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE,
+            objects.named(TargetJvmEnvironment, TargetJvmEnvironment.STANDARD_JVM)
+          )
+        }
+        because("LayoutLib and sdk-common depend on Guava's -jre published variant." +
+          "See https://github.com/cashapp/paparazzi/issues/906.")
+      }
```
TWiStErRob, line 17: Could/should this be somehow done in a way that it affects all of Paparazzi's users? If someone uses Paparazzi on `testImplementation`, they'll likely run into issues (we have so many reported in this repo).

paparazzi-gradle-plugin/src/test/projects/build-class-next-sdk/src/test/java/app/cash/paparazzi/plugin/test/BuildClassTest.kt
TWiStErRob: I'm happy either way, was just surprised.

Do I understand it correctly that this change depends on the Android SDK via compileSdk?

paparazzi-gradle-plugin/src/test/projects/common.gradle
jrodbx: I can't think of a way, without force bundling Guava/Truth, which I'm reluctant to do.  @JakeWharton, any thoughts?

paparazzi-gradle-plugin/src/test/projects/build-class-next-sdk/src/test/java/app/cash/paparazzi/plugin/test/BuildClassTest.kt
jrodbx: Yes, but now this change is leading me to pull on a long thread haha.  When I push an empty commit, CI passes, but if i bump Truth, it fails....why?  Is our Gradle caching setup on CI busted in some way?

TWiStErRob: Very likely
 * mandatory mention https://github.com/cashapp/paparazzi/pull/997
 * but also if something randomly reads a file from the SDK during `test` that must be also declared as an input, otherwise Gradle doesn't know.

Based on "empty commit" mention, it sounds like the second, assuming that when you push empty commit no test gets re-executed (which should be the case if every task is cached by gradle-build-action).

How this relates to truth upgrade is beyond me still though. You can try modifying this file in a non-meaningful way (e.g. add a comment) and see if it fails or passes.

TWiStErRob: Did a deep search of the SDK and found that this is located in: `platforms\android-34\build.prop`, then noticed that in the Paparazzi codebase this is mentioned here:

https://github.com/cashapp/paparazzi/blob/aec3fa37941b5ac4b580820f9fee9efc12107a28/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt#L112-L124

So the fix for this is to add all of those files as inputs to the test task, otherwise Gradle thinks that nothing has changed and load from cache.

jrodbx: nice find!  

Ok, so PrepareResourcesTask will re-run when compileSdkVersion changes:
```
```

which will cause the test task to re-run again
```
        test.systemProperties["paparazzi.test.resources"] =
```

the issue here is 99% of the time `$ANDROID_HOME/platforms/android-${compileSdkVersion}` will have static contents, but 1% (when using sandbox-level sdks), this will not be the case, as we're seeing here because build.prop is changed, but not being detected.

I'd hate to change the plugin to crawl through the entire platformDir checking for file content changes for the 1% test case, especially given initial findings that we might be caching huge payloads, so I think the "right fix" here is to disable caching for that specific plugin test.

jrodbx: https://github.com/cashapp/paparazzi/pull/1211

jrodbx: Another option -- if this comes up again -- would be to force the PrepareResourcesTask to rerun when compileSdkVersion doesn't parse to an integer value.

luis-cortes (approved):
LGTM

PR #1210, opened 2024-01-06
Key test failure artifact uploads by shard type

jrodbx:
Broken in this dependency bump due to breaking API change: https://github.com/cashapp/paparazzi/commit/5cdea5bd2b73ddef40bf119512315a5d0ec6d0b3

See #2 here: https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: test-failures
+          name: test-failures-${{ matrix.os }}-${{ matrix.java-version }}
           path: |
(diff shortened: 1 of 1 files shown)


TWiStErRob (commented):
LGTM, do you have an example failure? (just curious)

saket (approved):

jrodbx:
> LGTM, do you have an example failure? (just curious)

Example failure: https://github.com/cashapp/paparazzi/actions/runs/7400603576/job/20134571127?pr=1195

Note that every shard except for Ubuntu 18 (in this and most runs, since it tends to be the fastest running shard), the Upload Test Failures step fails.

PR #1214, opened 2024-01-09
Bump Guava to 33.0.0, apply guava fix to sample

jrodbx:
Renames `common.gradle` to `guava-fix.gradle`.  Unfortunately copied to sample, since I'm unsure how to share build logic between sample modules and functional test projects.


gamepro65 (approved):
👍

sample/build.gradle
```diff
@@ -34,3 +34,5 @@ dependencies {
 tasks.withType(com.diffplug.gradle.spotless.SpotlessTask).configureEach {
   dependsOn(tasks.withType(Test))
 }
+
+apply from: 'guava-fix.gradle'
```
TWiStErRob, line 38: I think you could include the other via a long relative path.

PR #1218, opened 2024-01-10
Always cleanup artifacts for repeatable tests

jrodbx:
Resolves https://github.com/cashapp/paparazzi/pull/695, Also applies the feedback given in https://github.com/cashapp/paparazzi/pull/695#pullrequestreview-1264422283.

Re: paparazzi tests, this adds a custom rule to swap out system properties.

Re: plugin tests, pre-cleaning potentially leftover state from previous test runs is an anti-pattern.  However, Gradle doesn't provide a straightforward way to pass in TemporaryFolder-based paths without resorting to custom test-specific property passing, such as attempted here: https://github.com/cashapp/paparazzi/pull/408/files#diff-8c52db63811b414fc18b3fe88650c819563e678a6cf71aa913ef4820e2c1cba4

File.deleteOnExit works well on files, but not for directories that have been populated after registering the shutdown hook with the VM. File.deleteRecursively handles non-empty directories, but doesn't register a shutdown hook, leaving stale files on test failures.

So we add a @After method to provide a hybrid of the two behaviors, allowing files to be cleaned up regardless of test success/failure.

To troubleshoot future test failures with this automatic cleanup, simply comment out the teardown logic between test runs.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -6,6 +6,7 @@ import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.gradle.testkit.runner.TaskOutcome.FROM_CACHE
 import org.gradle.testkit.runner.TaskOutcome.SUCCESS
+import org.junit.After
 import org.junit.Before
 import org.junit.Ignore
 import org.junit.Test
@@ -13,12 +14,19 @@ import java.io.File
 import java.nio.file.Files
 
 class PaparazziPluginTest {
+  private val filesToDelete = mutableListOf<File>()
+
   private lateinit var gradleRunner: GradleRunner
 
   @Before
   fun setUp() {
     gradleRunner = GradleRunner.create()
       .withPluginClasspath()
+  }
+
+  @After
+  fun tearDown() {
+    filesToDelete.forEach(File::deleteRecursively)
   }
 
   @Test
@@ -158,6 +166,8 @@ class PaparazziPluginTest {
   @Test
   fun prepareResourcesCaching() {
     val fixtureRoot = File("src/test/projects/prepare-resources-task-caching")
+    val buildDir = fixtureRoot.resolve("build").registerForDeletionOnExit()
+    fixtureRoot.resolve("build-cache").registerForDeletionOnExit()
 
     val firstRun = gradleRunner
       .withArguments("testRelease", "testDebug", "--build-cache", "--stacktrace")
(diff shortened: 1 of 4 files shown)


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1377,6 +1364,8 @@ class PaparazziPluginTest {
   @Test
   fun accessibilityErrorsLogged() {
     val fixtureRoot = File("src/test/projects/validate-accessibility")
+    // this is only a warning message, so subsequent runs would otherwise be UP-TO-DATE
+    fixtureRoot.resolve("build").registerForDeletionOnExit()
```
gamepro65, line 1368: I see other instances of fixtureRoot.resolve being replaced with buildDir. Does that apply here?

gamepro65 (approved):
Unfortunate but looks fine

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTestRule.kt
```diff
@@ -0,0 +1,56 @@
+package app.cash.paparazzi
+
+import org.junit.rules.ExternalResource
+import org.junit.rules.TemporaryFolder
+import org.junit.runner.Description
+import org.junit.runners.model.MultipleFailureException
+import org.junit.runners.model.Statement
+
+class PaparazziTestRule : ExternalResource() {
```
gamepro65, line 9: Could this be a `RuleChain` around Paparazzi instead? Doesn't necessarily need to manage Paparazzi too...

jrodbx:
> Unfortunate but looks fine

??  Why unfortunate?

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTestRule.kt
jrodbx: No.  Instantiating Paparazzi will automatically instantiate the SnapshotHandler which has default values set to the system properties this rule stubs out.

gamepro65: Got ya, thanks for explaining didn't catch that

gamepro65:
> > Unfortunate but looks fine
> 
> ?? Why unfortunate?

Just unfortunate how nuanced deleteOnExit is and that we have to write a custom rule to accommodate it's shortcomings.

jrodbx:
> Just unfortunate how nuanced deleteOnExit is and that we have to write a custom rule to accommodate it's shortcomings.

💯

TWiStErRob:
@jrodbx There's a way to make this "better": copy the whole test fixture to a temp directory so the Gradle builds are never executed in the src/... folder. Then the clean-up becomes as simple as using the temp rule with ensureDelete. Fresh state, every time all the time, opt-in not required inside test body.

PR #1226, opened 2024-01-13
Update task inputs for resources and assets to account for file renames and moves

jrodbx:
PathSensitivity.NONE implies that no part of the path matters -- not even the filename -- so it only looks at the content which is wrong since resource IDs are usually computed from filename.  In addition, if a resource is moved between qualifier folders, this would be undetected as it's considered UP-TO-DATE.

Adds additional test projects for verifying changes to:
* local resources/assets (src/main/debug, etc.)
* module resources/assets (:producer, etc.)
* external resources/assets (group:artifact:version, etc.)


gamepro65 (approved):
Looks good, love the test coverage

PR #1234, opened 2024-01-17
Migrate resources.txt to json format

jrodbx:
Will help delete some properties later

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
@@ -15,7 +15,9 @@
  */
 package app.cash.paparazzi.gradle
 
-import app.cash.paparazzi.gradle.utils.joinFiles
+import app.cash.paparazzi.gradle.utils.relativize
+import com.squareup.moshi.JsonClass
+import com.squareup.moshi.Moshi
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.RegularFileProperty
@@ -95,36 +97,42 @@ abstract class PrepareResourcesTask : DefaultTask() {
         artifactFiles.files.forEach { file ->
           add(file.useLines { lines -> lines.first() })
         }
-      }.joinToString(",")
+      }
     } else {
-      mainPackage
+      listOf(mainPackage)
     }
 
-    out.bufferedWriter()
-      .use {
-        it.write(mainPackage)
-        it.newLine()
-        it.write(mergeResourcesOutputDir.get())
-        it.newLine()
-        it.write(targetSdkVersion.get())
-        it.newLine()
-        // Use compileSdkVersion for system framework resources.
-        it.write("platforms/android-${compileSdkVersion.get()}/")
-        it.newLine()
-        it.write(mergeAssetsOutputDir.get())
-        it.newLine()
-        it.write(resourcePackageNames)
-        it.newLine()
-        it.write(projectResourceDirs.joinFiles(projectDirectory))
-        it.newLine()
-        it.write(moduleResourceDirs.joinFiles(projectDirectory))
-        it.newLine()
-        it.write(aarExplodedDirs.joinFiles(gradleUserHomeDirectory))
-        it.newLine()
-        it.write(projectAssetDirs.joinFiles(projectDirectory))
-        it.newLine()
-        it.write(aarAssetDirs.joinFiles(gradleUserHomeDirectory))
-        it.newLine()
-      }
+    val config = Config(
+      mainPackage = mainPackage,
+      mergeResourcesOutputDir = mergeResourcesOutputDir.get(),
+      targetSdkVersion = targetSdkVersion.get(),
+      // Use compileSdkVersion for system framework resources.
+      platformDir = "platforms/android-${compileSdkVersion.get()}/",
+      mergeAssetsOutputDir = mergeAssetsOutputDir.get(),
+      resourcePackageNames = resourcePackageNames,
+      projectResourceDirs = projectResourceDirs.relativize(projectDirectory),
+      moduleResourceDirs = moduleResourceDirs.relativize(projectDirectory),
+      aarExplodedDirs = aarExplodedDirs.relativize(gradleUserHomeDirectory),
+      projectAssetDirs = projectAssetDirs.relativize(projectDirectory),
+      aarAssetDirs = aarAssetDirs.relativize(gradleUserHomeDirectory)
+    )
+    val moshi = Moshi.Builder().build()!!
+    val json = moshi.adapter(Config::class.java).indent("  ").toJson(config)
+    out.writeText(json)
   }
+
+  @JsonClass(generateAdapter = true)
+  data class Config(
+    val mainPackage: String,
+    val mergeResourcesOutputDir: String,
+    val targetSdkVersion: String,
+    val platformDir: String,
+    val mergeAssetsOutputDir: String,
+    val resourcePackageNames: List<String>,
+    val projectResourceDirs: List<String>,
+    val moduleResourceDirs: List<String>,
+    val aarExplodedDirs: List<String>,
+    val projectAssetDirs: List<String>,
+    val aarAssetDirs: List<String>
+  )
 }
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -1,7 +1,12 @@
 package app.cash.paparazzi.gradle
 
 import app.cash.paparazzi.gradle.ImageSubject.Companion.assertThat
+import app.cash.paparazzi.gradle.PrepareResourcesTask.Config
+import com.google.common.truth.Correspondence
 import com.google.common.truth.Truth.assertThat
+import com.squareup.moshi.Moshi
+import okio.buffer
+import okio.source
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.gradle.testkit.runner.TaskOutcome.FROM_CACHE
@@ -183,12 +188,12 @@ class PaparazziPluginTest {
       assertThat(this!!.outcome).isNotEqualTo(FROM_CACHE)
     }
 
-    var resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    var resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.json")
     assertThat(resourcesFile.exists()).isTrue()
     var resourceFileContents = resourcesFile.readLines()
     assertThat(resourceFileContents.any { it.contains("release") }).isFalse()
 
-    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/release/resources.txt")
+    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/release/resources.json")
     assertThat(resourcesFile.exists()).isTrue()
     resourceFileContents = resourcesFile.readLines()
     assertThat(resourceFileContents.any { it.contains("debug") }).isFalse()
@@ -205,7 +210,7 @@ class PaparazziPluginTest {
       assertThat(this!!.outcome).isEqualTo(FROM_CACHE)
     }
 
-    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.txt")
+    resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.json")
     assertThat(resourcesFile.exists()).isTrue()
     resourceFileContents = resourcesFile.readLines()
     assertThat(resourceFileContents.any { it.contains("release") }).isFalse()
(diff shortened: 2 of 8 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -100,31 +100,25 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+      this["resourcePackageNames"] = resourcePackageNames
+      this["projectResourceDirs"] = projectResourceDirs.joinFiles(projectDirectory)
+      this["moduleResourceDirs"] = moduleResourceDirs.joinFiles(projectDirectory)
+      this["aarExplodedDirs"] = aarExplodedDirs.joinFiles(gradleUserHomeDirectory)
+      this["projectAssetDirs"] = projectAssetDirs.joinFiles(projectDirectory)
+      this["aarAssetDirs"] = aarAssetDirs.joinFiles(gradleUserHomeDirectory)
+    }
+    out.bufferedWriter().use { writer ->
+      properties.forEach { (k, v) ->
+        writer.write("$k=$v")
```
TWiStErRob, line 119: Why don't you use an actual Properties object for serialisation/deserialisation. Yes, it's old, but it takes care of escaping and is standard.

One benefit would be having multiline values, meaning we don't need to join and split commas on both sides, which would make debugging the file easier.

jrodbx: Considered it, but discovered when writing Properties, it adds a timestamp comment to the beginning of the file contents.  Could've worked around that, but figured it's just as easy to do this.

Re: escaping, I don't think any of these config values require that?

Re: multiline values, that's fair, although I've never been a fan of the multiline syntax:
```
multi.prop=value1
multi.prop=value2
...
```

Perhaps migrating to a JSON format would be a better idea?

TWiStErRob: > it adds a timestamp comment to the beginning of the file

https://docs.gradle.org/current/dsl/org.gradle.api.tasks.WriteProperties.html

> I don't think any of these config values require that?

They're file names (user input), you never know.

> I've never been a fan of the multiline syntax:

TIL, but I was thinking more like (without having to worry about how it is generated/read):
```
multi.prop=value1\
value2
```

> join/split being annoying, perhaps migrating to a JSON format

Potentially, you could use the same data class on both sides. But then you need a dependency on some serialization lib.

jrodbx: going with json!

jrodbx:
Need to clean up some things, but here's an initial look.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1701,4 +1744,12 @@ class PaparazziPluginTest {
   }
 
   private fun File.registerForDeletionOnExit() = apply { filesToDelete += this }
+
+  companion object {
+    private val CONFIG_ADAPTER = Moshi.Builder().build()!!.adapter(Config::class.java).indent("  ")
```
TWiStErRob, line 1749:
```suggestion
    private val CONFIG_ADAPTER = Moshi.Builder().build()!!.adapter(Config::class.java)
```
this one is not serializing, or is it validating indentation?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1701,4 +1744,12 @@ class PaparazziPluginTest {
...
+    private val MATCHES_PATTERN = Correspondence.from<String, String>(
+      { actual, expected -> actual.matches(expected.toRegex()) },
+      "matches"
+    )
```
TWiStErRob, line 1753: TIL

paparazzi/build.gradle
```diff
@@ -102,17 +102,19 @@ def generateTestConfig = tasks.register("generateTestConfig") {
...
-      writer.writeLine("31")
-      writer.writeLine("platforms/android-31/")
-      writer.writeLine(".")
-      writer.writeLine("app.cash.paparazzi")
-      writer.writeLine("")
-      writer.writeLine("")
-      writer.writeLine("")
-      writer.writeLine("")
-      writer.writeLine("")
+      writer << """{
```
TWiStErRob, line 105: `configFile.writeText("""...`

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -55,31 +59,45 @@ fun androidHome() = System.getenv("ANDROID_SDK_ROOT")
...
-  val configLines = resourcesFile.readLines()
-
   val projectDir = Paths.get(System.getProperty("paparazzi.project.dir"))
   val appTestDir = Paths.get(System.getProperty("paparazzi.build.dir"))
   val artifactsCacheDir = Paths.get(System.getProperty("paparazzi.artifacts.cache.dir"))
   val androidHome = Paths.get(androidHome())
+
+  val resourcesFile = File(System.getProperty("paparazzi.test.resources"))
+  val moshi = Moshi.Builder().build()!!
+  val config = moshi.adapter(Config::class.java).indent("  ").fromJson(resourcesFile.source().buffer())!!
```
TWiStErRob, line 69:
```suggestion
  val config = moshi.adapter(Config::class.java).fromJson(resourcesFile.source().buffer())!!
```

TWiStErRob (commented):
Random thoughts, looks nicer :)

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: oops, good catch.  updated.

paparazzi/build.gradle
jrodbx: updated

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
jrodbx: updated

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -95,36 +97,42 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+      resourcePackageNames = resourcePackageNames,
+      projectResourceDirs = projectResourceDirs.relativize(projectDirectory),
+      moduleResourceDirs = moduleResourceDirs.relativize(projectDirectory),
+      aarExplodedDirs = aarExplodedDirs.relativize(gradleUserHomeDirectory),
+      projectAssetDirs = projectAssetDirs.relativize(projectDirectory),
+      aarAssetDirs = aarAssetDirs.relativize(gradleUserHomeDirectory)
+    )
+    val moshi = Moshi.Builder().build()!!
+    val json = moshi.adapter(Config::class.java).indent("  ").toJson(config)
+    out.writeText(json)
```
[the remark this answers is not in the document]
jrodbx: updated

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -95,36 +97,42 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+      mergeAssetsOutputDir = mergeAssetsOutputDir.get(),
+      resourcePackageNames = resourcePackageNames,
+      projectResourceDirs = projectResourceDirs.relativize(projectDirectory),
+      moduleResourceDirs = moduleResourceDirs.relativize(projectDirectory),
+      aarExplodedDirs = aarExplodedDirs.relativize(gradleUserHomeDirectory),
+      projectAssetDirs = projectAssetDirs.relativize(projectDirectory),
+      aarAssetDirs = aarAssetDirs.relativize(gradleUserHomeDirectory)
+    )
+    val moshi = Moshi.Builder().build()!!
+    val json = moshi.adapter(Config::class.java).indent("  ").toJson(config)
```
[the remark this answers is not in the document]
jrodbx: the resources.json is an input to the test task, so should be fine!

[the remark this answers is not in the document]
jrodbx: that being said, the next PR removing 2 properties will confirm/refute that claim :)

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
```diff
@@ -55,31 +59,45 @@ fun androidHome() = System.getenv("ANDROID_SDK_ROOT")
...
+  val targetSdkVersion: String,
+  val platformDir: String,
+  val mergeAssetsOutputDir: String,
+  val resourcePackageNames: List<String>,
+  val projectResourceDirs: List<String>,
+  val moduleResourceDirs: List<String>,
+  val aarExplodedDirs: List<String>,
+  val projectAssetDirs: List<String>,
+  val aarAssetDirs: List<String>
+)
```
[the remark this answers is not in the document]
jrodbx: TODO: consolidate this

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
TWiStErRob: (I can't resolve it)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -95,36 +97,42 @@ abstract class PrepareResourcesTask : DefaultTask() {
...
+    out.writeText(json)
   }
+
+  @JsonClass(generateAdapter = true)
+  data class Config(
+    val mainPackage: String,
+    val mergeResourcesOutputDir: String,
+    val targetSdkVersion: String,
+    val platformDir: String,
```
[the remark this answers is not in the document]
jrodbx: the paths are all relative, so the relativization logic would probably then have to move into some custom Moshi de/serialization logic.  i'm inclined to leave as is.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: wdym?

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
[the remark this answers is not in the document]
jrodbx: tried to export Config into a separate `paparazzi-config` module, but our particular Gradle structure (where the plugin is setup in build-logic) is making this a little more confusing than I'd like to allocate to this, so I'll come back to this later.

Maybe @TWiStErRob or @JakeWharton has ideas how to get the module structure working.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
TWiStErRob: There's no Resolve button for me, because I'm neither author, nor collaborator on this repo/PR. So I, and probably other contributors too, cannot resolve conversations.

Screenshot_20240119-203708

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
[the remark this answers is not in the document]
TWiStErRob: I think you would create a module, and the include it in both settings.gradles. Gradle should be fine with this, I don't know what havoc it wreaks in IDEA sync though.

---

But I think creating a source folder, like `src/main/shared` in one place and using srcDir in both build.gradles is simpler. There'll be two .class files, in the plugin and in the library (like now), but a single source (of truth).

Check out:
https://github.com/TWiStErRob/net.twisterrob.gradle/blob/main/plugin/settings/build.gradle.kts#L52
+
https://github.com/TWiStErRob/net.twisterrob.gradle/blob/main/gradle/plugins/build.gradle.kts#L27

Note: the extra task is optional, if you are happy with a warning on IDEA sync. (I've seen nothing break yet, was using it this way for years. I just don't like warnings so I fixed it.)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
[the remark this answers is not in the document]
TWiStErRob: Makes sense. Relativization would be tricky inside serialisation, because the "relative to" might differ.

That said, a File object can contain a relative path, and toString() should give you the right thing. So a type change could "just work".

Feel free to resolve this, just wanted to throw it out there.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: ah understood.  no worries, it's not a requirement on this repo to resolve all discussions before merge.

paparazzi/src/main/java/app/cash/paparazzi/Environment.kt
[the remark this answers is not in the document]
jrodbx: oh nice, will explore in a follow-up.

gamepro65 (approved):
Nice

PR #1235, opened 2024-01-17
Remove legacy resources/assets loading mechanism

jrodbx:


JakeWharton (approved):

luis-cortes (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -117,10 +90,6 @@ class PaparazziPlugin : Plugin<Project> {
       val variantSlug = variant.name.capitalize(Locale.US)
```
TWiStErRob, line 90: Consider moving this/these down to usages.

TWiStErRob (commented):
Woohoo, finally we can revisit https://github.com/cashapp/paparazzi/pull/1136

Why is there a dependency on the other PR?

jrodbx:
> Why is there a dependency on the other PR?

To explicitly highlight which properties were going away!

kevinzheng-ap (approved):

PR #1239, opened 2024-01-18
Set HardwareConfig width and height based on orientation

### LINKED ISSUE #1190: `DeviceConfig#orientation` is not honored
**Description**
Using `DeviceConfig.PIXEL_C.copy(orientation = ScreenOrientation.PORTRAIT)` does not generate a portrait screenshot, instead i have to use `DeviceConfig.PIXEL_C.copy(screenHeight = DeviceConfig.PIXEL_C.screenWidth, screenWidth = DeviceConfig.PIXEL_C.screenHeight),`.

**Steps to Reproduce**
Attempt to render a portrait screen by creating a portrait `PIXEL_C` `DeviceConfig`

**Expected behavior**
Rendered screenshots should match orientation property of `DeviceConfig`.

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: MacOS Sonoma 
- Compile SDK: 34
- Gradle Version: 8.2
- Android Gradle Plugin Version: 8.2.0

**Screenshots**

BrianGardnerAtl:
Closes #1190.

Android Studio does this from within the HardwareConfigHelper class. It determines the width and height of the current orientation by checking which size is larger or smaller. The width is always the smallest side in portrait and the width is the largest side in landscape.


paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
```diff
@@ -110,9 +112,22 @@ data class DeviceConfig(
         versionQualifier = VersionQualifier()
       }
 
+  private val currentWidth: Int
+    get() = if (orientation == ScreenOrientation.PORTRAIT) {
```
TWiStErRob, line 116: Should this be exhaustive considering [square](https://developer.android.com/reference/android/content/res/Configuration#ORIENTATION_SQUARE) is unsupported?

```
Portrait - min
Landscape - max
Square - error
```

BrianGardnerAtl: Makes sense to me :+1:

jrodbx: nah, Square is deprecated since API 16

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -30,7 +31,8 @@ class AccessibilityRenderExtensionTest {
     deviceConfig = DeviceConfig.NEXUS_5.copy(
       // Needed to render accessibility content next to main content.
       screenWidth = DeviceConfig.NEXUS_5.screenWidth * 2,
-      softButtons = false
+      softButtons = false,
+      orientation = ScreenOrientation.LANDSCAPE,
```
BrianGardnerAtl, line 35: I had to add this because the accessibility test doubles the screen width, which the new `DeviceConfig` would consider to be `LANDSCAPE`

paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
```diff
@@ -110,9 +112,22 @@ data class DeviceConfig(
...
+      min(screenWidth, screenHeight)
```
jrodbx: curious: are there any current usages of `screenWidth` and `screenHeight` that might benefit from now using `currentWidth` and `currentHeight`?  Maybe not, but just checking.

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
jrodbx: how come?  is screenWidth * 2 >= screenHeight?

paparazzi-gradle-plugin/src/test/projects/verify-orientation/src/test/java/app/cash/paparazzi/plugin/test/ScreenOrientationTest.kt
```diff
@@ -0,0 +1,45 @@
...
+        setBackgroundColor(0)
+      },
+      "landscape"
+    )
+    paparazzi.unsafeUpdateConfig(deviceConfig = PIXEL_C.copy(orientation = PORTRAIT))
+    paparazzi.snapshot(
+      FrameLayout(paparazzi.context).apply {
+        setBackgroundColor(-1)
+      },
+      "portrait"
```
jrodbx, line 42: nit: named properties (same for above)

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
jrodbx: looks like the answer is yes...  my concern here is that this isn't straightforward to users.

@adamalyyan @geoff-powell can we move this double-width configuration out of the test and inside Paparazzi when it detects ARE is being used?

jrodbx (approved):
1) I think the change to AccessibilityRenderExtensionTest is the only mini-blocker to this landing.
2) Can we delete DeviceConfig.NEXUS_5_LAND and replace its consuming tests with something more appropriate to this change? :)

paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
jrodbx: whoops, nvm.  Brian beat me to it.  No biggy.

jrodbx (commented):
also, needs a spotless run to fix CI errors

paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
BrianGardnerAtl: There's a couple other screen orientations so I'll probably update the error to say portrait and landscape are the only supported ones. I want to look into the others at some point to see if they're worth supporting

TWiStErRob: They might be trivial to support, right?

paparazzi-gradle-plugin/src/test/projects/verify-orientation/src/test/java/app/cash/paparazzi/plugin/test/ScreenOrientationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  @Test
+  fun test() {
+    paparazzi.snapshot(
+      FrameLayout(paparazzi.context).apply {
+        setBackgroundColor(0)
+      },
+      landscape_snapshot_name
+    )
+    paparazzi.unsafeUpdateConfig(deviceConfig = PIXEL_C.copy(orientation = PORTRAIT))
```
jrodbx, line 37: nit: use TestParameterInjector instead?

paparazzi-gradle-plugin/src/test/projects/verify-orientation/src/test/java/app/cash/paparazzi/plugin/test/ScreenOrientationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    paparazzi.snapshot(
+      FrameLayout(paparazzi.context).apply {
+        setBackgroundColor(-1)
+      },
+      portrait_snapshot_name
```
jrodbx, line 42: use named arguments?

paparazzi-gradle-plugin/src/test/projects/verify-orientation/src/test/java/app/cash/paparazzi/plugin/test/ScreenOrientationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    )
+  }
+
+  companion object {
+    private const val landscape_snapshot_name = "landscape"
+    private const val portrait_snapshot_name = "portrait"
```
jrodbx, line 48: inline, since only one use and it's a test

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -625,9 +626,12 @@ class Paparazzi @JvmOverloads constructor(
 
   private fun DeviceConfig.updateIfAccessibilityTest(): DeviceConfig =
     if (renderExtensions.any { it is AccessibilityRenderExtension }) {
+      val newWidth = screenWidth * 2
+      val newOrientation = if (newWidth > screenHeight) ScreenOrientation.LANDSCAPE else ScreenOrientation.PORTRAIT
```
jrodbx, line 630: wait, why is this still necessary?

jrodbx (changes requested):

paparazzi-gradle-plugin/src/test/projects/verify-orientation/src/test/java/app/cash/paparazzi/plugin/test/ScreenOrientationTest.kt
jrodbx: i.e., 
```
view = FrameLayout...
name = "portrait"
```

BrianGardnerAtl: whoops, not sure where my brain was going at first. Fixing

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
BrianGardnerAtl: If we don't set the orientation then the `DeviceConfig` will flip the orientation, resulting in a taller screen because it will use the new 'width' as a height value instead

paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
BrianGardnerAtl: I looked through the usages and I don't think we'd need to use these elsewhere yet

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: got it.  hmm, ok, let's leave this as is then and figure out a better long-term solution.

jrodbx (approved):

PR #1240, opened 2024-01-18
Updates the DeviceConfig screenWidth internally for accessibility tests so consumers do not have to

adamalyyan:
This removes the need to double the `DeviceConfig.screenWidth` in individual tests when using the `AccessibilityRenderExtension` and instead Paparazzi will handle this internally


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -665,3 +672,13 @@ class Paparazzi @JvmOverloads constructor(
       }
   }
 }
+
+fun DeviceConfig.maybeForAccessibility(accessibilityRenderExtensionEnabled: Boolean): DeviceConfig =
```
jrodbx, line 676:
```suggestion
fun DeviceConfig.updateIfAccessibilityTest(accessibilityRenderExtensionEnabled: Boolean): DeviceConfig =
```

jrodbx, line 676: move extension function into Paparazzi class and inline the `accessibilityRenderExtensionEnabled` logic?

jrodbx (approved):

PR #1244, opened 2024-01-19
Update accessibility test snapshots

jrodbx:
Looks like these snapshots should have updated as part of https://github.com/cashapp/paparazzi/pull/1240?


gamepro65 (approved):
👍

adamalyyan:
Ah my bad, thanks for fixing!

PR #1247, opened 2024-01-20
Fix variant caching issues in new resource/asset loading mechanisms

jrodbx:
Similar to https://github.com/cashapp/paparazzi/pull/720.  Unblocks the current build failures in https://github.com/cashapp/paparazzi/pull/1235.

Using task properties like the following:
```
  @get:InputFiles
  @get:PathSensitive(PathSensitivity.RELATIVE)
  abstract val projectResourceDirs: ConfigurableFileCollection
```
is great for working with Gradle remote caching (since it ignores absolute paths which can be user-specific) and will only examine subpath contents.  However, this doesn't jive well with Android variants, which can often be similar in contents, but differ in the base path, e.g., 

`src/debug/res` vs`src/release/res` or,
`../some_module/build/intermediates/packaged_res/debug` and `../some_module/build/intermediates/packaged_res/release`

In those cases, we've seen CI jobs fail when the `prepareResourcesTask` is considered `FROM_CACHE` because the contents of the respective folders haven't changed, but the cached `resources.json` file contains, say `debug` when Paparazzi is being ran for a `release` build, resulting in a crash due to the proposed file/path not being found.

To remedy, the properties will now be `ListProperty<String>` allowing `resources.json` to be re-created on path changes due to changes in variant builds, and move the contents checking to the test task properties.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -887,11 +887,15 @@ class PaparazziPluginTest {
     firstResourceFile.copyTo(destResourceFile, overwrite = false)
 
     val firstRun = gradleRunner
-      .withArguments(":preparePaparazziDebugResources", "--build-cache", "--stacktrace")
+      .withArguments("testDebug", "--build-cache", "--stacktrace")
       .forwardOutput()
       .runFixture(fixtureRoot) { build() }
 
     with(firstRun.task(":preparePaparazziDebugResources")) {
+      assertThat(this).isNotNull()
+      assertThat(this!!.outcome).isEqualTo(SUCCESS)
+    }
+    with(firstRun.task(":testDebugUnitTest")) {
       assertThat(this).isNotNull()
       assertThat(this!!.outcome).isEqualTo(SUCCESS)
     }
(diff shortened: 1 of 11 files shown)


jrodbx:
Seems like this PR is only failing on Windows, with a consistent error message across all 4 tests:
```
    expected: FROM_CACHE
    but was : UP_TO_DATE
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -887,33 +887,47 @@ class PaparazziPluginTest {
...
 
     val resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.json")
 
+    println("JROD (run 1): ${resourcesFile.readText()}")
     var config = resourcesFile.loadConfig()
     assertThat(config.projectResourceDirs).containsExactly("src/main/res", "src/debug/res")
 
+    assertThat(resourcesFile.exists()).isTrue()
     buildDir.deleteRecursively()
+    assertThat(resourcesFile.exists()).isFalse()
```
jrodbx, line 911: Looks like this assertion fails on Windows

jrodbx: ah, looks like we're holding onto the File handle after loading the json, which is preventing the parent buildDir folder from being cleaned up, i.e., 
```
  private fun File.loadConfig() = CONFIG_ADAPTER.fromJson(source().buffer())!!
```
vs
```
  private fun File.loadConfig() = source().buffer().use { CONFIG_ADAPTER.fromJson(it)!! }
```

Interesting that this is failing on Windows only, though.

kevinzheng-ap (approved):

PR #1264, opened 2024-01-26
Restrict Paparazzi's public API

jrodbx:
As a precursor to the SDKifying work, I've sadly realized that we haven't done a great job of restricting the API surface of the project.  This PR attempts to correct that by adding the kotlin binary validation plugin + slapping internal on a lot of things.

One thing to call out is this PR will make `HtmlReportWriter` and `SnapshotVerifier` *internal* as well.  I felt keeping `SnapshotHandler` public was enough, but due to some current design shortcomings in the snapshot pipeline (to be addressed at some point!), that level of strictness might break current consumers of the project (which I'd like to avoid!) who have implemented workarounds.

Therefore, I'd be open to reverting those 2 classes (and others!) for the time being, until we come up with a better solution/design.  Please speak up!

cc: @TWiStErRob

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,18 +10,26 @@ buildscript {
 
   dependencies {
     classpath libs.plugin.kotlin
+    classpath libs.plugin.kotlinApiDump
     classpath libs.plugin.android
     classpath libs.plugin.mavenPublish
     classpath libs.plugin.dokka
     classpath libs.plugin.versions
     classpath libs.plugin.spotless
     classpath libs.plugin.buildConfig
+    classpath libs.plugin.poko
     classpath libs.grgit
 
     // Normally you would declare a version here, but we use dependency substitution in
     // settings.gradle to use the version built from inside the repo.
     classpath 'app.cash.paparazzi:paparazzi-gradle-plugin'
   }
+}
+
+apply plugin: 'binary-compatibility-validator'
+
+apiValidation {
+  ignoredProjects += ['libs', 'sample']
 }
 
 subprojects {
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -241,7 +241,7 @@ class PaparazziPlugin : Plugin<Project> {
     }
   }
 
-  open class PaparazziTask : DefaultTask() {
+  abstract class PaparazziTask : DefaultTask() {
     @Option(option = "tests", description = "Sets test class or method name to be included, '*' is supported.")
     open fun setTestNameIncludePatterns(testNamePattern: List<String>): PaparazziTask {
       project.tasks.withType(Test::class.java).configureEach {
--- a/paparazzi/api/paparazzi.api
+++ b/paparazzi/api/paparazzi.api
@@ -1,0 +1,205 @@
+public final class app/cash/paparazzi/DeviceConfig {
+	public static final field $stable I
+	public static final field Companion Lapp/cash/paparazzi/DeviceConfig$Companion;
+	public static final field GALAXY_WATCH4_CLASSIC_LARGE Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_10 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_4 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_5 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_5_LAND Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_7 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field NEXUS_7_2012 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_2 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_2_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_3 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_3A Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_3A_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_3_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_4 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_4A Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_4_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_5 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_6 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_6_PRO Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_C Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field WEAR_OS_SMALL_ROUND Lapp/cash/paparazzi/DeviceConfig;
+	public static final field WEAR_OS_SQUARE Lapp/cash/paparazzi/DeviceConfig;
+	public fun <init> ()V
+	public fun <init> (IIIILcom/android/resources/ScreenOrientation;Lcom/android/resources/UiMode;Lcom/android/resources/NightMode;Lcom/android/resources/Density;FLcom/android/resources/LayoutDirection;Ljava/lang/String;Lcom/android/resources/ScreenRatio;Lcom/android/resources/ScreenSize;Lcom/android/resources/Keyboard;Lcom/android/resources/TouchScreen;Lcom/android/resources/KeyboardState;ZLcom/android/resources/Navigation;Lcom/android/resources/ScreenRound;Ljava/lang/String;)V
+	public synthetic fun <init> (IIIILcom/android/resources/ScreenOrientation;Lcom/android/resources/UiMode;Lcom/android/resources/NightMode;Lcom/android/resources/Density;FLcom/android/resources/LayoutDirection;Ljava/lang/String;Lcom/android/resources/ScreenRatio;Lcom/android/resources/ScreenSize;Lcom/android/resources/Keyboard;Lcom/android/resources/TouchScreen;Lcom/android/resources/KeyboardState;ZLcom/android/resources/Navigation;Lcom/android/resources/ScreenRound;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun copy (IIIILcom/android/resources/ScreenOrientation;Lcom/android/resources/UiMode;Lcom/android/resources/NightMode;Lcom/android/resources/Density;FLcom/android/resources/LayoutDirection;Ljava/lang/String;Lcom/android/resources/ScreenRatio;Lcom/android/resources/ScreenSize;Lcom/android/resources/Keyboard;Lcom/android/resources/TouchScreen;Lcom/android/resources/KeyboardState;ZLcom/android/resources/Navigation;Lcom/android/resources/ScreenRound;)Lapp/cash/paparazzi/DeviceConfig;
+	public static synthetic fun copy$default (Lapp/cash/paparazzi/DeviceConfig;IIIILcom/android/resources/ScreenOrientation;Lcom/android/resources/UiMode;Lcom/android/resources/NightMode;Lcom/android/resources/Density;FLcom/android/resources/LayoutDirection;Ljava/lang/String;Lcom/android/resources/ScreenRatio;Lcom/android/resources/ScreenSize;Lcom/android/resources/Keyboard;Lcom/android/resources/TouchScreen;Lcom/android/resources/KeyboardState;ZLcom/android/resources/Navigation;Lcom/android/resources/ScreenRound;ILjava/lang/Object;)Lapp/cash/paparazzi/DeviceConfig;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getDensity ()Lcom/android/resources/Density;
+	public final fun getFolderConfiguration ()Lcom/android/ide/common/resources/configuration/FolderConfiguration;
+	public final fun getFontScale ()F
+	public final fun getHardwareConfig ()Lcom/android/ide/common/rendering/api/HardwareConfig;
+	public final fun getKeyboard ()Lcom/android/resources/Keyboard;
+	public final fun getKeyboardState ()Lcom/android/resources/KeyboardState;
+	public final fun getLayoutDirection ()Lcom/android/resources/LayoutDirection;
+	public final fun getLocale ()Ljava/lang/String;
+	public final fun getNavigation ()Lcom/android/resources/Navigation;
+	public final fun getNightMode ()Lcom/android/resources/NightMode;
+	public final fun getOrientation ()Lcom/android/resources/ScreenOrientation;
+	public final fun getRatio ()Lcom/android/resources/ScreenRatio;
+	public final fun getReleased ()Ljava/lang/String;
+	public final fun getScreenHeight ()I
+	public final fun getScreenRound ()Lcom/android/resources/ScreenRound;
+	public final fun getScreenWidth ()I
+	public final fun getSize ()Lcom/android/resources/ScreenSize;
+	public final fun getSoftButtons ()Z
+	public final fun getTouchScreen ()Lcom/android/resources/TouchScreen;
+	public final fun getUiMode ()Lcom/android/resources/UiMode;
+	public final fun getUiModeMask ()I
+	public final fun getXdpi ()I
+	public final fun getYdpi ()I
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class app/cash/paparazzi/DeviceConfig$Companion {
+}
+
+public final class app/cash/paparazzi/Environment {
+	public static final field $stable I
+	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;)V
+	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;)Lapp/cash/paparazzi/Environment;
+	public static synthetic fun copy$default (Lapp/cash/paparazzi/Environment;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;ILjava/lang/Object;)Lapp/cash/paparazzi/Environment;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getAllModuleAssetDirs ()Ljava/util/List;
+	public final fun getAppTestDir ()Ljava/lang/String;
+	public final fun getCompileSdkVersion ()I
+	public final fun getLibraryAssetDirs ()Ljava/util/List;
+	public final fun getLibraryResourceDirs ()Ljava/util/List;
+	public final fun getLocalResourceDirs ()Ljava/util/List;
+	public final fun getModuleResourceDirs ()Ljava/util/List;
+	public final fun getPackageName ()Ljava/lang/String;
+	public final fun getPlatformDir ()Ljava/lang/String;
+	public final fun getResourcePackageNames ()Ljava/util/List;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class app/cash/paparazzi/EnvironmentKt {
+	public static final fun androidHome ()Ljava/lang/String;
+	public static final fun detectEnvironment ()Lapp/cash/paparazzi/Environment;
+}
+
+public final class app/cash/paparazzi/Flags {
+	public static final field $stable I
+	public static final field DEBUG_LINKED_OBJECTS Ljava/lang/String;
+	public static final field INSTANCE Lapp/cash/paparazzi/Flags;
+}
+
+public final class app/cash/paparazzi/HtmlReportWriter : app/cash/paparazzi/SnapshotHandler {
+	public static final field $stable I
+	public fun <init> ()V
+	public fun <init> (Ljava/lang/String;)V
+	public fun <init> (Ljava/lang/String;Ljava/io/File;)V
+	public fun <init> (Ljava/lang/String;Ljava/io/File;Ljava/io/File;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/io/File;Ljava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun close ()V
+	public fun newFrameHandler (Lapp/cash/paparazzi/Snapshot;II)Lapp/cash/paparazzi/SnapshotHandler$FrameHandler;
+}
+
+public final class app/cash/paparazzi/InstantAnimationsRule : org/junit/rules/TestRule {
+	public static final field $stable I
+	public fun <init> ()V
+	public fun apply (Lorg/junit/runners/model/Statement;Lorg/junit/runner/Description;)Lorg/junit/runners/model/Statement;
+}
+
+public final class app/cash/paparazzi/Paparazzi : org/junit/rules/TestRule {
+	public static final field $stable I
+	public fun <init> ()V
+	public fun <init> (Lapp/cash/paparazzi/Environment;)V
+	public fun <init> (Lapp/cash/paparazzi/Environment;Lapp/cash/paparazzi/DeviceConfig;)V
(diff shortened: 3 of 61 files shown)


saket (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -50,7 +50,7 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinAndroidTarget
 import java.util.Locale
 
 @Suppress("unused")
-class PaparazziPlugin : Plugin<Project> {
+internal class PaparazziPlugin : Plugin<Project> {
```
jrodbx, line 53: @JakeWharton is there any reason why Paraphrase keeps the plugin type public? https://github.com/cashapp/paraphrase/blob/main/plugin/api/plugin.api

Seems like Gradle is able to access the plugin and task types without the need to keep them plugin, which I guess makes sense, since build stuff should mostly be implementation detail.

jrodbx:
> that level of strictness might break current consumers of the project (which I'd like to avoid!) who have implemented workarounds.

@gamepro65 kindly reminds me that we have also internally implemented a workaround using `HtmlReportWriter` and `SnapshotVerifier` LOL.  ok, reverting those for now...

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
drewhamilton: I suppose this works because `internal` classes are public from the JVM's point of view, but I think Gradle plugins are semantically public in the sense that the external Gradle tool needs to instantiate them and invoke their API.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -241,7 +241,7 @@ class PaparazziPlugin : Plugin<Project> {
     }
   }
 
-  open class PaparazziTask : DefaultTask() {
+  internal open class PaparazziTask : DefaultTask() {
```
drewhamilton, line 244: Is there no case in which a consumer would need to do `tasks.withType(PaparazziTask)`? (And same q for `PrepareResourcesTask`.)

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,259 @@
...
+	public final fun snapshot (Landroid/view/View;Ljava/lang/String;)V
+	public final fun snapshot (Landroid/view/View;Ljava/lang/String;J)V
+	public final fun snapshot (Ljava/lang/String;Lkotlin/jvm/functions/Function2;)V
+	public static synthetic fun snapshot$default (Lapp/cash/paparazzi/Paparazzi;Landroid/view/View;Ljava/lang/String;JILjava/lang/Object;)V
+	public static synthetic fun snapshot$default (Lapp/cash/paparazzi/Paparazzi;Ljava/lang/String;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
+	public final fun unsafeUpdateConfig (Lapp/cash/paparazzi/DeviceConfig;Ljava/lang/String;Lcom/android/ide/common/rendering/api/SessionParams$RenderingMode;)V
+	public static synthetic fun unsafeUpdateConfig$default (Lapp/cash/paparazzi/Paparazzi;Lapp/cash/paparazzi/DeviceConfig;Ljava/lang/String;Lcom/android/ide/common/rendering/api/SessionParams$RenderingMode;ILjava/lang/Object;)V
+}
+
+public final class app/cash/paparazzi/Paparazzi$Companion {
```
drewhamilton, line 183: Paparazzi's `companion object` can be internal since all of its properties are internal or private.

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,259 @@
...
+public final class app/cash/paparazzi/SnapshotVerifier : app/cash/paparazzi/SnapshotHandler {
+	public static final field $stable I
+	public fun <init> (D)V
+	public fun <init> (DLjava/io/File;)V
+	public synthetic fun <init> (DLjava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun close ()V
+	public fun newFrameHandler (Lapp/cash/paparazzi/Snapshot;II)Lapp/cash/paparazzi/SnapshotHandler$FrameHandler;
+}
+
+public final class app/cash/paparazzi/TestName {
```
drewhamilton, line 238: I think this can be internal

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: no, these tasks are meant to be implementation detail.  consumers of plugins should use depend on task artifacts, not the task types themselves (although it's not always simple to get this right).

paparazzi/api/paparazzi.api
jrodbx: good call

paparazzi/api/paparazzi.api
jrodbx: not this one unfortunately, as it's a member of Snapshot which is a member of SnapshotHandler which is a public type.

build.gradle
TWiStErRob, line 1: What do you think about turning on library mode, so public API is explicitly `public ...`?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: Also `project.pluginManager.apply(PaparazziPlugin::class.java)` is a way of applying, I would suspect this breaks some users. Migration is easy though, so you can add to release notes to use `.apply("app.cash.paparazzi")` instead.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
TWiStErRob: I agree these are implementation details, but considering the amount of workarounds necessary to keep Paparazzi alive in a project, I would imagine there might be a need to adjust some properties on the task.

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,255 @@
...
+	public final fun component1 ()Ljava/lang/String;
+	public final fun component10 ()Ljava/util/List;
+	public final fun component2 ()Ljava/lang/String;
+	public final fun component3 ()Ljava/lang/String;
+	public final fun component4 ()I
+	public final fun component5 ()Ljava/util/List;
+	public final fun component6 ()Ljava/util/List;
+	public final fun component7 ()Ljava/util/List;
+	public final fun component8 ()Ljava/util/List;
+	public final fun component9 ()Ljava/util/List;
```
TWiStErRob, line 98: Do you really want to publish this? adding a prop will be breaking unless it's last.

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,255 @@
...
+	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;)Lapp/cash/paparazzi/Environment;
+	public static synthetic fun copy$default (Lapp/cash/paparazzi/Environment;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;ILjava/lang/Object;)Lapp/cash/paparazzi/Environment;
+	public fun equals (Ljava/lang/Object;)Z
```
TWiStErRob, line 101: does equals/hashcode make sense on this?

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,255 @@
...
+	public final fun getLocalResourceDirs ()Ljava/util/List;
+	public final fun getModuleResourceDirs ()Ljava/util/List;
+	public final fun getPackageName ()Ljava/lang/String;
+	public final fun getPlatformDir ()Ljava/lang/String;
+	public final fun getResourcePackageNames ()Ljava/util/List;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class app/cash/paparazzi/EnvironmentKt {
```
TWiStErRob, line 116: Is there a way to make these `@JvmStatic` on `Environment` for Java consumers? Probably breaking for Kotlin users due to import change.

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,255 @@
...
+}
+
+public final class app/cash/paparazzi/InstantAnimationsRule : org/junit/rules/TestRule {
+	public static final field $stable I
+	public fun <init> ()V
+	public fun apply (Lorg/junit/runners/model/Statement;Lorg/junit/runner/Description;)Lorg/junit/runners/model/Statement;
+}
+
+public final class app/cash/paparazzi/Paparazzi : org/junit/rules/TestRule {
+	public static final field $stable I
```
TWiStErRob, line 145: who dis?

paparazzi/api/paparazzi.api
TWiStErRob: (applies to other classes too)

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -637,7 +637,7 @@ class Paparazzi @JvmOverloads constructor(
       this
     }
 
-  companion object {
+  internal companion object {
```
TWiStErRob, line 640: Keep this `public`, if someone added `fun Paparazzi.Companion.createMyPaparazzi()` to get `Paparazzi.createMyPaparazzi()` function, they'll break. Internals are hidden enough.

paparazzi/api/paparazzi.api
```diff
@@ -0,0 +1,255 @@
...
+	public static final field PIXEL_4_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_5 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_6 Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_6_PRO Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_C Lapp/cash/paparazzi/DeviceConfig;
+	public static final field PIXEL_XL Lapp/cash/paparazzi/DeviceConfig;
+	public static final field WEAR_OS_SMALL_ROUND Lapp/cash/paparazzi/DeviceConfig;
+	public static final field WEAR_OS_SQUARE Lapp/cash/paparazzi/DeviceConfig;
+	public fun <init> ()V
+	public fun <init> (IIIILcom/android/resources/ScreenOrientation;Lcom/android/resources/UiMode;Lcom/android/resources/NightMode;Lcom/android/resources/Density;FLcom/android/resources/LayoutDirection;Ljava/lang/String;Lcom/android/resources/ScreenRatio;Lcom/android/resources/ScreenSize;Lcom/android/resources/Keyboard;Lcom/android/resources/TouchScreen;Lcom/android/resources/KeyboardState;ZLcom/android/resources/Navigation;Lcom/android/resources/ScreenRound;Ljava/lang/String;)V
```
TWiStErRob, line 29: Wanted to highlight that you might be able to hide the SnapshotHandler implementations IF you make `determineHandler` public (which it should be anyway, because it's tricky to replicate that logic otherwise.)

paparazzi/api/paparazzi.api
TWiStErRob: [copy is useful though](https://github.com/TWiStErRob/net.twisterrob.sun/blob/8bd175671ae513cb090e1b65284322c2384ca80e/component/paparazzi/src/main/java/net/twisterrob/sun/test/screenshots/PaparazziFactory.kt#L43-L46)... tricky one.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
TWiStErRob: cc @drewhamilton

paparazzi/api/paparazzi.api
drewhamilton: Aha, missed that one

build.gradle
jrodbx: Yup, will do in a follow-up.  Doing so here would have made the PR bigger.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: ah, fair point.  although I prefer the string-based id approach, I'm fine with leaving this public.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: fair point, reverting this

paparazzi/api/paparazzi.api
jrodbx: yea, I think that's fine?  since this would also be a constructor argument change, which unless we have default values, would also be a breaking change.  what do other libraries typically do here w.r.t. exposing data classes?

paparazzi/api/paparazzi.api
jrodbx: it's part of the data class codegen

paparazzi/api/paparazzi.api
jrodbx: they're top level functions, so Java consumers can already access them like so:

```
import static app.cash.paparazzi.EnvironmentKt.detectEnvironment;

import app.cash.paparazzi.Environment;

public class Foo {
  Environment env = detectEnvironment();
}
```

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: I don't follow, as @drewhamilton mentioned, all the members of this companion object are either private or internal.

Re: the internal members,
```
    internal val TIME_OFFSET_NANOS = TimeUnit.HOURS.toNanos(1L)
    internal lateinit var renderer: Renderer
    internal val isInitialized get() = ::renderer.isInitialized
    internal lateinit var sessionParamsBuilder: SessionParamsBuilder
```

...we certainly don't want these depended on or exposed.  What am I missing?

paparazzi/api/paparazzi.api
jrodbx: good question.  i don't know.  @JakeWharton @drewhamilton ?

paparazzi/api/paparazzi.api
jrodbx: Yea, I considered this, but it seems to be that the conditional logic varies by consumer, making `determineHandler` less useful than exposing the implementations directly (for now...).

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
TWiStErRob: Yeah, that is true, but you want to keep the class Paparazzi$Companion.class to exist as public, even though it's contents are (package) private. So people can attach extension functions on it. Without a public declaration there's no way. For example I think most java.* classes don't have companions, so it's impossible to create "static" extension functions, even if you wanted to. I'm afk, but I'm sure you can find some empty companions in other libraries.

paparazzi/api/paparazzi.api
TWiStErRob: I know it is, I'm questioning the use of the (data class) tool, not it's result :)

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
TWiStErRob: Try my extension fun out on the sample w/ and w/o companion.

paparazzi/api/paparazzi.api
TWiStErRob: Yeah, it's just leaking the implementation details of Kotlin (Kt suffix)

paparazzi/api/paparazzi.api
TWiStErRob: I think they don't, but @JakeWharton probably have a lot of experience with this.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: Discussed offline.

I hadn't realized how popular it is to use extension functions on companion objects as factory methods, i.e.,
```
fun Paparazzi.Companion.createPaparazzi() = Paparazzi(...)
```

I would have thought the unused receiver would make this a bit of an anti-pattern, but @TWiStErRob has convinced me that this pattern is quite helpful for namespacing reasons, vs using top-level functions which pollute the global namespace, i.e.,
```
fun createPaparazzi() = Paparazzi(...)
```

So I'm reverting this one!

paparazzi/api/paparazzi.api
jrodbx: Ah got it.  I can't think of any use case; it came for free with the data class codegen, which was added in the first place for the nice `copy` semantics.

Earlier, I spiked out what things would look like by removing `data` and manually writing `toString` and `copy`.  Simple enough, but it's prone to forgetting adding new members in the future.

I'm inclined to leave for now, and maybe move this to a Builder/newBuilder pattern later.  Thoughts?

paparazzi/api/paparazzi.api
jrodbx: Addressed overall in this comment (below): https://github.com/cashapp/paparazzi/pull/1264#discussion_r1469102437

Egorand (approved):

paparazzi/api/paparazzi.api
JakeWharton: The Compose compiler added it. Not sure why it isn't marked synthetic.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
JakeWharton: Yes. Public for apply and withType and such. Plus Gradle requires it (at least in bytecode).

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
JakeWharton: > consumers of plugins should use depend on task artifacts, not the task types themselves

The best practice to get task artifacts are through properties on the task type since it gives you an implicit task dependency.

I would remove `open`, however, and consider something like `abstract` instead.

paparazzi/api/paparazzi.api
JakeWharton: data classes should never be used in public library API

paparazzi/api/paparazzi.api
JakeWharton: Use Poko

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
JakeWharton: Not convinced that you need to do this.

paparazzi/api/paparazzi.api
drewhamilton: Poko helps with the `toString`, but not the `copy`/builder semantics (yet???).

JakeWharton: Ah yeah. I kinda read the thread backwards so my mind ended up on equals/hashCode.

You have to implement copy yourself since for every new property you must add a new overload.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: Can you elaborate more?  Not convinced that it should remain `internal` or revert back to `public`?

TWiStErRob: Note: a few example from the "wild":

 * [Kotlin > Boolean](https://github.com/JetBrains/kotlin/blob/36d5ffd4eb646497f447495f3737cdf174eb3d1d/core/builtins/native/kotlin/Boolean.kt#L17)
 * [kotlinx-datetime > Clock](https://github.com/Kotlin/kotlinx-datetime/blob/aabed342cd1471b1138fc1130245c1a7a1c54a40/core/common/src/Clock.kt#L28)
 * [ktor > Url](https://github.com/ktorio/ktor/blob/dd9b42e3a548cd821fa8411801fcc1f8b83f61a1/ktor-http/common/src/io/ktor/http/Url.kt#L121)
 * [kotlin > TimeSource](https://github.com/kackey0-1/kotlin/blob/b5eafb9eb76311b9507786242a3f2b3b6021f0d0/libraries/stdlib/src/kotlin/time/TimeSource.kt#L104)
 * [mirai > MiraiLogger](https://github.com/killerprojecte/mirai/blob/e60384126b7e49fb00017d54dd6f80e332d2d927/mirai-core-api/src/commonMain/kotlin/utils/MiraiLogger.kt#L69)

JakeWharton: I suppose the list of counter examples would then be every other class written ever? So I think that wins by many orders of magnitude.

Hanging an empty companion is not your problem. It's [the language's problem](https://github.com/Kotlin/KEEP/issues/348).

jrodbx: Gotcha.  Makes sense.  Reapplying `internal`.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: changing to `abstract`

paparazzi/api/paparazzi.api
jrodbx: discussed offline.  plan is to go with manual copy + poko + manual equals/hashcode to remove the componentN functions.  a possible additional improvement will be to update poko to allow selectivity of which methods to generate.

jrodbx: @drewhamilton @JakeWharton @TWiStErRob PTAAL at the Poko changes

drewhamilton: LGTM. Similar to when it was a data class, the generated `equals` and `hashCode` aren't really needed, but aren't a compatibility concern and don't hurt anything by being there.

TWiStErRob: You missed `DeviceConfig.released` from `copy`, see https://github.com/cashapp/paparazzi/compare/44a138a58f960c49c322ddd87e0e4f9fbee18dbb..11367688e8275ea34b2b60d6a3b75f6a71c41efe#diff-c2bf71ab4fcca2f6780e9537deb521cc35dba6a7f9a26ee4ed8ff860dce9514fR31 vs 51 on the left.

jrodbx: that was intentional, because I didn't think it made sense to copy that member over.  does that intention violate a best practice?

paparazzi/src/main/java/app/cash/paparazzi/Snapshot.kt
```diff
@@ -16,17 +16,27 @@
 package app.cash.paparazzi
 
 import com.squareup.moshi.JsonClass
+import dev.drewhamilton.poko.Poko
 import java.util.Date
 import java.util.Locale
 
 @JsonClass(generateAdapter = true)
```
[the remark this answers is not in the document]
jrodbx: Yup, unfortunately Snapshot can't be internal, because it's a member of SnapshotHandler which is a public interface.  Is there a way to mark a JsonAdapter to be internal despite its "adapted class" being public?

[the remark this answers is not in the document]
TWiStErRob: I just meant the adapter. If you ever change serialization framework, it'll be a breaking change, which sounds strange.

paparazzi/api/paparazzi.api
TWiStErRob: I think it violates the principle of least astonishment. `DeviceConfig.PIXEL_XL.copy(locale = X)` would suddenly take the release date of Nexus 4.

paparazzi/src/main/java/app/cash/paparazzi/Snapshot.kt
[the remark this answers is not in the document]
jrodbx: I understand you meant the Adapter, but it seems like Moshi generates the adapter with the same access modifier as the backing class.  I'm all ears if there's a way to configure Moshi to generate an internal adapter.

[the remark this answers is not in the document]
jrodbx: Hmm looks like I can copy the codegen'd adapter, mark it as internal and point to that mutated class, using [this](https://github.com/square/moshi/blob/2cc9ebb88e91ce8af690f1e9fa6fcb80167d41e1/moshi/src/main/java/com/squareup/moshi/JsonClass.kt#L66)?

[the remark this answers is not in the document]
TWiStErRob: Sounds like a possibility, question is: is it worth it?
I'll defer to someone more experienced here (and/or a contributor to Moshi).

paparazzi/api/paparazzi.api
jrodbx: oh i see what my mistake was: [b2cb074](https://github.com/cashapp/paparazzi/pull/1264/commits/b2cb074ef3d0fcd0b7b9d57f3170df939801f1ef)

PR #1268, opened 2024-01-29
Switch to moshi reflection adapter to avoid exposing a generated adapter in public API

jrodbx:
precursor to https://github.com/cashapp/paparazzi/pull/1264

cc: @TWiStErRob

### DIFF
--- a/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi-gradle-plugin/build.gradle
@@ -1,7 +1,6 @@
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'java-gradle-plugin'
 apply plugin: 'com.github.gmazzo.buildconfig'
-apply plugin: 'com.google.devtools.ksp'
 
 // This module is included in two projects:
 // - In the root project where it's released as one of our artifacts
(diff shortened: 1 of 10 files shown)


paparazzi/build.gradle
```diff
@@ -55,11 +55,10 @@ dependencies {
   api platform(libs.kotlin.bom)
   implementation libs.moshi.core
   implementation libs.moshi.adapters
+  implementation libs.moshi.kotlinReflect
   implementation libs.jcodec.core
   implementation libs.jcodec.javase
 
-  ksp libs.moshi.kotlinCodegen
```
TWiStErRob, line 61: hmm, bye-bye KSP Gradle Plugin for :paparazzi and :paparazzi-gradle-plugin?

jrodbx: i mean i guess i couldddddd use reflection only for `Snapshot` and codegen'd for the rest, but felt this was cleaner.  Any reason to reconsider?

TWiStErRob: That might be confusing to maintain, one way of serialization is probably enough.
Aren't all serialized classes public? It's the same problem for all.

Btw, my point was that this is the last `ksp` line, so the apply plugin can be removed. Which might totally remove KSP from the build.

paparazzi-gradle-plugin/build.gradle
```diff
@@ -1,7 +1,6 @@
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'java-gradle-plugin'
 apply plugin: 'com.github.gmazzo.buildconfig'
-apply plugin: 'com.google.devtools.ksp'
```
JakeWharton, line 4: Bonus!

JakeWharton (approved):

PR #1286, opened 2024-02-07
Initial API spike based on Paparazzi2.kt notes

BrianGardnerAtl:
TODO:

- Delete code from the original paparazzi module 
- Figure out what tests and resources also need to move over
- Figure out what needs to be removed from build.gradle file
- Refactor RenderExtension set on Device to interceptor pattern
- Figure out which class to use for class loader in Renderer.kt L153

`Paparazzi2.kt` has the previous API notes


paparazzi2/src/main/java/app/cash/paparazzi/ClipSpec.kt
```diff
@@ -0,0 +1,8 @@
+package app.cash.paparazzi
+
+data class ClipSpec(
+  val frameSpec: FrameSpec,
+  val start: Long,
+  val end: Long,
+  val fps: Int
```
swankjesse, line 7: Should we choose something more discrete for this? Like start + frameDelay + frameCount? I think we should offer a factory method that uses fps, but I think the model shouldn’t require us to have tricky rules around rounding. (Like what happens if we do 1 second at 30 FPS? Does that make 31 frames?)

paparazzi2/src/main/java/app/cash/paparazzi/Device.kt
```diff
@@ -0,0 +1,556 @@
...
+    view: View,
+    clipSpec: ClipSpec
+  ) = Clip(
+    spec = clipSpec,
+    images = flow {
+      val viewGroup = bridgeRenderSession.rootViews[0].viewObject as ViewGroup
+      val modifiedView = prepareSnapshot(view, viewGroup)
+      // Add one to the frame count so we get the last frame. Otherwise a 1 second, 60 FPS animation
+      // our 60th frame will be at time 983 ms, and we want our last frame to be 1,000 ms. This gets
+      // us 61 frames for a 1 second animation, 121 frames for a 2 second animation, etc.
```
swankjesse, line 146: I like this behaviour! But I’d like to move this math out to the code that produces the ClipSpec?

paparazzi2/src/main/java/app/cash/paparazzi/DeviceConfig.kt
```diff
@@ -0,0 +1,622 @@
...
+  val size: ScreenSize = ScreenSize.NORMAL,
+  val keyboard: Keyboard = Keyboard.NOKEY,
+  val touchScreen: TouchScreen = TouchScreen.FINGER,
+  val keyboardState: KeyboardState = KeyboardState.SOFT,
+  val softButtons: Boolean = true,
+  val navigation: Navigation = Navigation.NONAV,
+  val screenRound: com.android.resources.ScreenRound? = null,
+  val released: String = "November 13, 2012"
+) {
+  val folderConfiguration: FolderConfiguration
```
swankjesse, line 87: Off-topic: I’m surprised that this is called folder!

paparazzi2/src/main/java/app/cash/paparazzi/ClipSpec.kt
```diff
@@ -0,0 +1,31 @@
+package app.cash.paparazzi
+
+import java.util.concurrent.TimeUnit
+
+public data class ClipSpec(
+  val frameSpec: FrameSpec,
+  val start: Long,
+  val frameDelay: Int,
```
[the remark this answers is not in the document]
BrianGardnerAtl: Makes sense, plus this will make some of the logic more straightforward

paparazzi2/src/main/java/app/cash/paparazzi/DeviceConfig.kt
BrianGardnerAtl: Everyone loves resource qualifiers 🙃

swankjesse (approved):

swankjesse:
# Please defer my feedback

I made a bunch of small nitpicky feedback on this PR, but I think it’s gonna make everyone happier if you merge first. We can address the nitpicks later! Landing this is more important.

Issue #549, opened 2022-08-18
Feature Request: Provide a way of simplifying snapshot name


Chrisjc28:
**Description**
It would be great if the snapshot didn't have to have all of the values in it such as the package name, class name and the method name and you could provide a custom name with only the details you wanted.

JulianEggers:
I would suggest making the naming schema configurable. 
The package of our setup is always the same which makes it unnecessary clutter in the filename.

jrodbx:
@Chrisjc28 (or anyone) mind elaborating on the "why"?  Would help prioritizing if some use cases were provided.

Chrisjc28:
We had a recent issue where the memory reference was being used in the name of the snapshot, when running locally all of our tests worked fine. However when running in our CI the test would use a different memory reference in the name causing them to fail. This is a niche issue but if we were able to have more control over what the file name was this wouldn't have happened. We have put something in place as a temporary fix but it has place a code smell in our project i would rather not have.

jrodbx:
I've seen this before with usages of TestParameterInjector with custom provided values.  In those cases, providing a toString method or implementing as a data class is the suggested fix.

Chrisjc28:
In our case we have used a secondary constructor and it's being changed to a data class, however there is no benefit for us to have the data class. 

An example of the class we are using in our test is - 

sealed Class AppState {
        object: Loading,
        data class Success(val data: Any)
        data class Error(val error:Any)
} 
 

For us converting loading to a data class has no benefit and makes us have to make use of a secondary constructor just so we don't have to and parentheses to every instance of loading.


The suggestion of using toString would you mean to implement that in the loading object itself?

apkelly:
Being able to specify a custom filename would be great, instead of the hard-coded list which includes things like packagename

```
internal fun Snapshot.toFileName(
  delimiter: String = "_",
  extension: String
): String {
  val formattedLabel = if (name != null) {
    "$delimiter${name.toLowerCase(Locale.US).replace("\\s".toRegex(), delimiter)}"
  } else {
    ""
  }
  return "${testName.packageName}${delimiter}${testName.className}${delimiter}${testName.methodName}$formattedLabel.$extension"
}
```

Being able to provide a custom filename would allow me to remove the packageName part, which is the same for all my tests e.g. `com.example.tests.ui.paparazzi`, I only really need className and methodName.

KatieBarnett:
[...]

PR #1292, opened 2024-02-14
Add a way to provide a file name for Snapshots

magnusvs:
Adds a FileNameProvider that can be implemented in any way to specify file names for a recorded Snapshot. 
Closes feature request #549

We wanted this feature because we had a developer on Linux with ecryptfs which only supports file names up to 143 characters, and we we're exceeding this with the current Snapshot file names.


jdeebee:
This is really needed, can you get this merged please? Right now, we ended up setting up a new module in our project (named "paparazzi-utils") to workaround this issue (as a lot of changes need to be made to private classes which in turn depend on a host of other internal classes - which all need to be copied - leading to hassle and overhead especially when maintaining, leading to making updating Paparazzi difficult therefore less adoption of new releases). Another workaround I can think of would be forking this repo and hosting and publishing it in an internal Nexus repository, but that would also require a lot of overhead and maintenance hassle as well. Would love it if we can this PR merged or provide this functionality some other way. We re-use the screenshots taken by Paparazzi for our "storybook gallery" and the files need to be named properly as we wish (name of the screen + light/dark theme + name of the device - but this convention is subject to change so the naming should be custom).

safa007:
+1, this is very much needed. @jrodbx and paparazzi team could we get eyes on this please?

jrodbx:
It might be helpful to know that we have discussed moving the current file naming scheme into, say, a json descriptor, and making the filename more of a hash (similar to how we do in the build directory).  The point being that the file names are an implementation detail of Paparazzi and should not depended on or tweaked.

safa007:
> It might be helpful to know that we have discussed moving the current file naming scheme into, say, a json descriptor, and making the filename more of a hash (similar to how we do in the build directory). The point being that the file names are an implementation detail of Paparazzi and should not depended on or tweaked.

@jrodbx the current problem our team is facing is that `toFileName` is using `testName.methodName` within the function which is adding an index that changes every time verify snapshots runs.

For example: `test[3.Group.ScreenName,1.PIXEL_4,1.light]` - the first `3` here is an index set by `testName.methodName` and it fluctuates depending on the test ordering.

This behavior makes Paparazzi effectively unusable for us. With over 1000 snapshots, any minor change to previews, or adding new previews, causes all of the snapshot files to be regenerated with new file names. This makes it very challenging to isolate actual changes in a pull request.

I believe the most effective solution would be to allow customization of the file naming. It seems reasonable that this flexibility should be supported.

Issue #1016, opened 2023-07-21
Error inflating view that uses DataBinding


danielgomesoliveira:
**Description**
I've been trying to write a simple test for a view that uses data binding (root is layout), but I keep getting the error 
```
Caused by: android.view.InflateException: Binary XML file line #1: Error inflating class layout
Caused by: java.lang.ClassNotFoundException: onCreateView
...
Caused by: java.lang.ClassNotFoundException: layout
```

On the project sample I can see examples using View Binding, but not Data Binding.

**Steps to Reproduce**
Try to inflate a view with Data Binding

```
@Test
fun testDefaultBinding() { 
    val view = FragmentBinding.inflate(paparazzi.layoutInflater)
    paparazzi.snapshot(view.root)
}
```

**Expected behavior**
Test runs normally

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: Mac OS
- Compile SDK: 33
- Gradle Version: 8.0.2
- Android Gradle Plugin Version: 8.0.2

**Screenshots**
N\A

jrodbx:
@danielgomesoliveira I know you provided the repro steps above, but would you mind providing a link to a project repo with data binding enabled?

danielgomesoliveira:
@jrodbx thank you for the quick answer. I prepared a small repo where this can be reproduced [here].(https://github.com/danielgomesoliveira/PaparazziWithDataBinding/blob/main/app/src/test/java/com/test/paparazziwithdatabinding/PaparazziDataBindingTest.kt).

Let me know if you need any other information

anonlatte:
You may try to use [this workaround](https://github.com/cashapp/paparazzi/issues/700#issuecomment-1402222087) but for me it didn't work.

PR #196, opened 2021-02-23
Add support for inline complex XML resources

jrodbx:
Supersedes and closes https://github.com/cashapp/paparazzi/pull/153.  Closes https://github.com/cashapp/paparazzi/issues/116.

Complex XML resources (via the aapt:attr tag) are extracted/transformed by the Android Asset Packaging Tool (AAPT) at build time, such that subsequent APK resource inflation is unaware of this process.

To support real-time rendering, LayoutEditor/layoutlib-impl instead uses a delegating parser whose internal parser uses native IntelliJ PSI APIs to transform the resource file into an in-memory DOM and create dynamic references to the subtags referenced by the <aapt:attr> tag.  During the inflation walk, these dynamic references are mapped to their representing resources.

Paparazzi mimics this by reimplementing/copying much of the IntelliJ implementation.  The bulk of the work is moving LayoutPullParser's old DOM parsing behavior into ResourceParser, and replacing it with the in-memory "parsing" state machine implemented in InMemoryParser.

1:1 of IntelliJ classes to Paparazzi classes:
* [LayoutlibCallbackImpl](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-master-dev:android/src/com/android/tools/idea/rendering/LayoutlibCallbackImpl.java;l=137;drc=21ce82eed9d3315e4d59950860f9e2b82d6c3dd7;bpv=0;bpt=0) -> PaparazziCallback
* [LayoutPsiPullParser](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-master-dev:android/src/com/android/tools/idea/rendering/parsers/LayoutPsiPullParser.java;bpv=0;bpt=0) -> LayoutPullParser
* [LayoutPullParser](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-master-dev:android/src/com/android/tools/idea/rendering/parsers/LayoutPullParser.java;bpv=0;bpt=0) -> InMemoryParser
* [TagSnapshot](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-master-dev:android/src/com/android/tools/idea/rendering/parsers/TagSnapshot.java;bpv=0;bpt=0) -> TagSnapshot

PR #1305, opened 2024-02-25
DataBinding fixing

BrianGardnerAtl:
Related to #1016 

- [ ] Fix resolution of generated `BR` class when using the `<data>` tag 
- [ ] Refactor to generate the `DataBindingMapperImpl` instead of requiring consumers to add one themselves

Two issues I ran into trying to fix this. The first is that attempting to access any data binding from the test results in a class not found for `DataBindingMapperImpl`. One does get generated but it is not accessible. I found [a similar issue related to Roboectric](https://issuetracker.google.com/issues/126775542?pli=1) where one of the posts mentioned creating a fake version of the class and adding the actual version via reflection which worked.
[Here is the example version](https://github.com/annypatel/databinding-test/blob/master/binding-mapper/src/main/java/androidx/databinding/DataBinderMapperImpl.kt).
I tweaked the implementation since we are unable to query the application package name from the unit test variant.

The second issue is that paparazzi is not loading the processed layout files which is why it's erroring when parsing the `layout` tag. I found an additional resource directory under `build/intermediates/packaged_res/debug` that has the processed layouts. Adding this to the `projectResourceDirs` in the paparazzi plugin allows the correct layout to load.
*important* this has to be the last resource directory. I tried having it first but it was defaulting to the original, unprocessed layout which fails on the `layout` tag.


paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/java/androidx/databinding/DataBinderMapperImpl.kt
```diff
@@ -0,0 +1,17 @@
+package androidx.databinding
+
+import android.annotation.SuppressLint
+
+/**
+ * Running instrumented tests with Robolectric on Data Binding enabled library modules throws [NoClassDefFoundError] for
```
BrianGardnerAtl, line 6: Snagged this from an associated thread about Robolectric having issues finding the same DataBinderMapperImpl. I can redo the comment here and link to the original thread for more context

paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/res/layout/layout.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
```
gamepro65, line 2: Have you tried using the <data/> tag?

BrianGardnerAtl: Not yet but I can add a test for that next

paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/java/androidx/databinding/DataBinderMapperImpl.kt
BrianGardnerAtl: Yeah that's a big concern. ByteBuddy could work as long as we can access the module's package to find the generated class

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -759,7 +759,11 @@ class PaparazziPluginTest {
       "app.cash.paparazzi.plugin.test.module1",
       "app.cash.paparazzi.plugin.test.module2"
     )
-    assertThat(config.projectResourceDirs).containsExactly("src/main/res", "src/debug/res")
+    assertThat(config.projectResourceDirs).containsExactly(
+      "src/main/res",
+      "src/debug/res",
+      "build/intermediates/packaged_res/debug"
```
gamepro65, line 765: Interesting idea pulling this, looking at how AS handles it we could also do something similar in our `LayoutPullParser`

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/parsers/LayoutRenderPullParser.java;l=972

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:android/rendering/testSrc/com/android/tools/idea/rendering/parsers/LayoutRenderPullParserTest.java;l=492?q=AaptAttr

paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/res/layout/layout.xml
jrodbx: I think let's not support `<data/>` tags; I feel like it's a Pandora's box and not worth the squeeze.

The original issue in https://github.com/cashapp/paparazzi/issues/1016 doesn't use it, and Paparazzi takes a fairly opinionated stance about separating business logic from view.  Instead, users can grab a handle to textview and set their logic-driven properties directly in test code.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: agreed.  I noticed similar, but I think our current AaptParser logic might need some refactoring.  I'm exploring that now.

paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/java/androidx/databinding/DataBinderMapperImpl.kt
jrodbx: So this is interesting.  Note that in https://github.com/cashapp/paparazzi/issues/1016, the repro project is an Android application, not a library project.  This is because DataBinding generates `DataBinderMapperImpl` in the app module.

Now we could support that, but coupled with my comment about `<data>`, to what cause?

I'm of the opinion that we should strive to support ViewBinding style inflations, via`FooBinding.inflate(..)`, but not the full suite of possibilities that DataBinding offers.  Thoughts?

jrodbx: Nonetheless, if the feeling is a resounding "yes, we should support databinding", then I think the way to tackle this unpleasant dev experience would be to check for the Gradle option being set in the buildFeatures, pass that as a System Property ("paparazzi.androidx.databinding.enabled") and load this impl in Paparazzi.kt when that property is set.

jrodbx:
> The second issue is that paparazzi is not loading the processed layout files which is why it's erroring when parsing the layout tag. I found an additional resource directory under build/intermediates/packaged_res/debug that has the processed layouts. Adding this to the projectResourceDirs in the paparazzi plugin allows the correct layout to load.

Yea, this will be partly fixed in https://github.com/cashapp/paparazzi/pull/1136

paparazzi-gradle-plugin/src/test/projects/verify-data-binding/src/main/java/androidx/databinding/DataBinderMapperImpl.kt
BrianGardnerAtl: I think we still need the `DataBinderMapperImpl` to support ViewBinding so I can look into setting up a system property

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: following up on this: I debugged how LayoutLib renders `fragment_dashboard.xml` (i.e., the repro from https://github.com/cashapp/paparazzi/issues/1016#issuecomment-1646342203) in the IDE and this layout:

```
<layout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools">

    <androidx.constraintlayout.widget.ConstraintLayout
        android:layout_height="match_parent"
        tools:context=".ui.dashboard.DashboardFragment">

        <TextView
            android:id="@+id/text_dashboard"
            ... />
    </androidx.constraintlayout.widget.ConstraintLayout>

</layout>
```

results in this TagSnapshot hierarchy:

```
TagSnapshot{
  androidx.constraintlayout.widget.ConstraintLayout, 
  attributes=[
    AttributeSnapshot{layout_width="match_parent"}, 
    AttributeSnapshot{layout_height="match_parent"}, 
    AttributeSnapshot{context=".ui.dashboard.DashboardFragment"}, 
    AttributeSnapshot{tag="layout/fragment_dashboard_0"}
  ], 
  children=[
    TagSnapshot{
      TextView, 
        attributes=[
          AttributeSnapshot{id="@+id/text_dashboard"}, 
          ...
	], 
	children=[]
     }
  ]
}
```

In particular, note the `AttributeSnapshot{tag="layout/fragment_dashboard_0"}`, which reminds me of the work we did in https://github.com/cashapp/paparazzi/pull/196 to support Cash Card gradients.

Issue #1304, opened 2024-02-25
Snapshotting fails if `Dispatchers.setMain` and `Dispatchers.resetMain` have been called


ansman:
**Description**
We're seeing random failures using the latest paparazzi version due to another test setting the main dispatcher.

**Steps to Reproduce**
Here is an example test that showcase the issue:
```kotlin
class ExampleTest {
    init {
        // This is actually happening in another test class that happens to run before this test
        // Commenting out these two lines makes the test pass
        Dispatchers.setMain(UnconfinedTestDispatcher())
        Dispatchers.resetMain()
    }

    @get:Rule
    val paparazzi = Paparazzi()

    @Test
    fun test() {
        paparazzi.snapshot { Box(Modifier.size(10.dp).background(Color.Red)) }
    }
}
```

**Expected behavior**
Paparazzi should be able to record regardless of if main has previously been changed.

**Additional information:**
- Paparazzi Version: 1.3.2
- OS: macOS 14.2
- Compile SDK: 34
- Gradle Version: 8.6
- Android Gradle Plugin Version: 8.2.2

TWiStErRob:
Duplicate of https://github.com/cashapp/paparazzi/issues/1198 ?

ansman:
> Duplicate of #1198 ?

Probably same root cause if I had to guess but it’s a different problem (I’m getting exceptions during paparazzi tests). I’ll let someone else decide if they want to merge the tickets or not

TWiStErRob:
You might be right, I misread the version numbers.

Anyway, the solution might be similar: can you please check if [1.3.3-SNAPSHOT](https://github.com/cashapp/paparazzi?tab=readme-ov-file#releases) makes a difference? There have been [a lot of changes since 1.3.2](https://github.com/cashapp/paparazzi/compare/1.3.2...master).

kevinsunhg:
is this issue indeed fixed? we are on 1.3.3 and still seeing random failures without manually setting main dispatcher

gamepro65:
The exception thrown in this issue has been fixed since we no longer use Dispatchers.Main.

What specifically are you seeing? Please open a new issue with the failure + any repro / implementation specifics and I can revisit.

cpaleop:
False alarm, I forgot to apply again the Guava workaround since we skipped version 1.3.2. My bad folks 🤦🏻‍♂️

PR #1306, opened 2024-02-26
Use our own HandlerDispatcher for ComposeUi tests

gamepro65:
coroutines-test can put `Dispatchers.Main` in an unpredictable state. Instead we create our own HandlerDispatcher which prevents this dependency from interfering with Paparazzi.

Resolves: https://github.com/cashapp/paparazzi/issues/1304


jrodbx (approved):
Nice and straightforward!

Issue #1067, opened 2023-08-30
Generated strings not rendering correctly in new resource loading mechanism


mshearer123:
**Description**
After updating to paparazzi 1.3.1, strings do not render from R.strings correctly anymore.
using `stringResource(id = R.string.x)`

We use compose public Previews in our paparazzi tests.   The strings render fine in the preview.

**Steps to Reproduce**
Provide a sample project, failing test, or steps to reproduce.

**Expected behavior**
String would be rendered from R resources correctly

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: mac (m1)
- Compile SDK: 33
- Gradle Version: 8.0.2
- Android Gradle Plugin Version: 8.0.2

**Screenshots**
![Screenshot 2023-08-30 at 10 23 17](https://github.com/cashapp/paparazzi/assets/2699109/a44ced6a-f28f-467b-8595-8abc4f2758fa)

**update**
fixed using
```
app.cash.paparazzi.legacy.resource.loading=true
app.cash.paparazzi.legacy.asset.loading=true
```

jrodbx:
Closed by #1307

mshearer123:
This issue has resurfaced for us in v1.3.3
we had been using
```
app.cash.paparazzi.legacy.resource.loading=false
app.cash.paparazzi.legacy.asset.loading=false
```

kevinzheng-ap:
@mshearer123 I can definitely help to look into more details. 

1. Could you provide how these generated strings are declared in your gradle file?
2. Could you check if the generated string are in `build/generated/res/resValues/debug/values/gradleResValues.xml`? And if `build/generated/res/resValues/debug/values` exists in `build/intermediates/paparazzi/debug/resources.json`?

mshearer123:
Thanks @kevinzheng-ap ,

We don't have `gradleResValues.xml` files, we have a translation task that generates

kevinzheng-ap:
@mshearer123 How do you ask gradle to use the resources from these files in your task?

mshearer123:
i believe with these functions
```
  @TaskAction
  fun generateStringsXml() {
    val englishMap = Yaml().load<Map<String, Any>>(File("${langFolder.get().asFile.absolutePath}/en.lyaml").inputStream())
    val reference = englishMap["en"] as LinkedHashMap<String, Any>

    listOf(
      "en" to "values",
      "da" to "values-da",
      "de" to "values-de",
      "es" to "values-es",
[...]

PR #1307, opened 2024-02-28
Fixes Issue About Generated Strings  Rendering 

kevinzheng-ap:
Related to https://github.com/cashapp/paparazzi/issues/1067

Fixes issue that generated strings are not rendered correctly in new resource loading mechanism

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -26,6 +26,7 @@ import com.android.build.gradle.internal.api.TestedVariant
 import com.android.build.gradle.internal.dsl.BaseAppModuleExtension
 import com.android.build.gradle.internal.dsl.DynamicFeatureExtension
 import com.android.build.gradle.internal.publishing.AndroidArtifacts.ArtifactType
+import com.android.build.gradle.tasks.GenerateResValues
 import org.gradle.api.DefaultTask
 import org.gradle.api.DomainObjectSet
 import org.gradle.api.Plugin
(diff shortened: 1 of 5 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -97,8 +99,10 @@ public class PaparazziPlugin : Plugin<Project> {
       val reportOutputDir = project.extensions.getByType(ReportingExtension::class.java).baseDirectory.dir("paparazzi/${variant.name}")
       val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
 
+      val generateResValuesProvider = project.tasks.named("generate${variantSlug}ResValues") as TaskProvider<GenerateResValues>
```
kevinzheng-ap, line 102: Any better API to retrieve the generated res folder?

sample/build.gradle
```diff
@@ -9,6 +9,8 @@ android {
   compileSdk libs.versions.compileSdk.get() as int
   defaultConfig {
     minSdk libs.versions.minSdk.get() as int
+
+    resValue("string", "string_name_generated", "Test String generated")
```
jrodbx, line 13:
```suggestion
    resValue("string", "generated_string_name", "Generated Test String")
```

sample/src/main/java/app/cash/paparazzi/sample/ResourcesDemo.kt
```diff
@@ -124,5 +124,9 @@ fun ResourcesDemo() {
       resources.getStringArray(R.array.string_array_name).joinToString(),
       color = Color.Black
     )
+    Text(
+      resources.getString(R.string.string_name_generated),
```
jrodbx, line 128:
```suggestion
      resources.getString(R.string.generated_string_name),
```

sample/src/main/java/app/cash/paparazzi/sample/ResourcesDemoView.kt
```diff
@@ -59,6 +59,7 @@ class ResourcesDemoView(context: Context) : LinearLayout(context) {
     )
     addTextView(resources.getText(R.string.string_name_html_unescaped))
     addTextView(resources.getStringArray(R.array.string_array_name).joinToString())
+    addTextView(resources.getText(R.string.string_name_generated))
```
jrodbx, line 62:
```suggestion
    addTextView(resources.getText(R.string. generated_string_name))
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
kevinzheng-ap: It seems `InternalArtifactType.GENERATED_RES.[getOutputPath|getIntermediateOutputPath]` does not give the right path

Expected: build/generated/res/resValues/debug
Actual: build/[generated|intermediates]/generated_res/debug

Looking into getOutputPath or getIntermediateOutputPath, what it does inside is just to join paths

```
FileUtils.join(
    getOutputDir(buildDirectory.get().asFile),
    variantIdentifier,
    *paths,
    forceFilename.ifEmpty { getFileSystemLocationName() }
)
```

So it is static and different from actual generated res path. And generated res path could be set via `registerGeneratedResFolders`, so we should always read it from `GenerateResValues` task rather than constructing the static path programatically

jrodbx: Oh you're right.  I was confusing the two output directory types, good catch!

We still should try and avoid task references where possible, since that peeks into plugin implementation detail, but not a blocker for this.  Moving to AGP 8 apis, will change this logic to automatically be included as part of localResourceDirs.

jrodbx (approved):

PR #1315, opened 2024-03-01
Publish layout lib via GithubActions

geoff-powell:


BrianGardnerAtl (approved):

swankjesse (approved):

.github/workflows/publish-layoutlib.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/checkout@v4
+
+      - name: Configure JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+
+      - name: Publish Artifacts
+        run: ./gradlew publishMavenPublicationToMavenCentralRepository
```
jrodbx, line 31:
```suggestion
        run: ./gradlew publishMavenNativeLibraryPublicationToMavenCentralRepository
```
otherwise, this would publish a partial release of Paparazzi.

jrodbx (approved):

.github/workflows/publish-layoutlib.yml
geoff-powell: yep I had done the separate module deploys but that is way more concise. Updating

.github/workflows/publish-layoutlib.yml
```diff
@@ -0,0 +1,35 @@
+name: publish-layoutlib
+
+on:
+  workflow_dispatch:
+
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
```
jrodbx, line 10:
```suggestion
on:
  push:
    tags:
      - '**'
```

Nit: Might be better to copy release.yml to use tag-based releases.  Actually, no, that might conflict with the normal release process.  Maybe limit line 8 to use a special branch regex, i.e., something like `**/layoutlib**`.

geoff-powell: yes, sorry something is up with github. I updated this with a `layoutlib` regex but it seems github isn't updating although the code is pushed 🤔

jrodbx: @JakeWharton any ideas?

geoff-powell: https://github.com/cashapp/paparazzi/compare/master...gpowell/publish-layoutlib
shows 2 commits.

This pr only has my initial commit?

geoff-powell: nevermind, it just updated. Must have been a weird Github PR update delay IDK

.github/workflows/publish-layoutlib.yml
```diff
@@ -0,0 +1,30 @@
...
+  workflow_dispatch:
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
+
+jobs:
+
+  publish:
+    runs-on: ubuntu-latest
+    if: github.ref =~ 'refs/heads/**/layoutlib**'
```
jrodbx, line 13: interesting; what's the difference between this and using branch above?

jrodbx, line 13:
```suggestion
    if: github.ref =~ 'refs/heads/**/publish-layoutlib**'
```

just looked at my branch name, and thought it'd be useful to add this change so that the push to publish can be discerned from the push to bump its usage in the project.

jrodbx: Love it!

.github/workflows/publish-layoutlib.yml
TWiStErRob: Yeah, there was an incident around this time.

.github/workflows/publish-layoutlib.yml
TWiStErRob: @geoff-powell If it's manual, why is it necessary to put an arbitrary condition on the branch name? Only accounts with write access to the repo can trigger workflow_dispatch (because they need to push a branch in the repo), why can't it be triggered on any branch? Is this maybe a historical remnant from an earlier version where trigger was on tag push?

Issue #529, opened 2022-08-03
Feature Request: Warn and/or remove stale images


tir38:
My `images` directory was full of snapshots for tests that I have either renamed or deleted entirely. It would be helpful if Paparazzi could keep this directory clean by either warning me of golden images with no matching test, or outright deleting them for me.

alexvanyo:
It might make sense for this be a part of the `record` step: Delete all existing images first, since the recording will regenerate all of them. Any that were previously present (and were from a deleted or renamed test) would then not be regenerated, as expected.

BenedictP:
I think this feature is hard to do right. E.g., We are not running all tests when we write new screenshot tests or after we change a view and need new golden masters. We only run the screenshot tests we need. Otherwise, it will take a while ;)

tir38:
>We only run the screenshot tests we need.

Usually we do the same, except on CI. It would kinda suck to have a CI job fail because of unused test images. A separate `cleanup` task that could be run periodically would decouple your concerns I think.

kikin81:
@BenedictP interested to know how you manage running only the screenshot test you need? is there a specific gradle command? assuming you do per feature module?

borsini:
Just tested and this one is fixed now with the new `cleanRecordPaparazzi` command being introduced in [1.3.4](https://github.com/cashapp/paparazzi/releases/tag/1.3.4). Many thanks!

colinmarsch:
Closing per https://github.com/cashapp/paparazzi/issues/529#issuecomment-2157548176

PR #1316, opened 2024-03-01
Adding `cleanRecordPaparazzi...` task to help tidy up orphaned snapshots

nak5ive:
This adds a global task `deletePaparazziSnapshots` and variant-specific tasks `cleanRecordPaparazzi[Debug]`

A clean-record task helps identify and tidy-up orphaned snapshots after test coverage/naming changes.


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -77,6 +78,7 @@ public class PaparazziPlugin : Plugin<Project> {
   private fun <T> setupPaparazzi(project: Project, variants: DomainObjectSet<T>) where T : BaseVariant, T : TestedVariant {
     project.addTestDependency()
     val nativePlatformFileCollection = project.setupNativePlatformDependency()
+    val snapshotOutputDir = project.layout.projectDirectory.dir("src/test/snapshots")
```
gamepro65, line 81: I think this needs to read from the system property

`System.getProperty("paparazzi.snapshot.dir")`

geoff-powell (approved):

jrodbx (commented):
FYI, this will close https://github.com/cashapp/paparazzi/issues/529.

@jamesmosley55 explored this in the past and we had preferred an more incremental approach via https://docs.gradle.org/current/userguide/custom_tasks.html#incremental_tasks vs a delete-all-then-record approach.  Imagine changing the name of test method in FormBlocker, and deleting 1000+ images and rewriting 998?  Maybe I'm prematurely optimizing?

What James was exploring was creating an index of image paths when recording and checking that on rerun to do an in-memory diff, eliminating the paths that are no longer present.

Can we add tests as part of this PR?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -179,8 +191,15 @@ public class PaparazziPlugin : Plugin<Project> {
       val recordTaskProvider = project.tasks.register("recordPaparazzi$variantSlug", PaparazziTask::class.java) {
         it.group = VERIFICATION_GROUP
         it.description = "Record golden images for variant '${variant.name}'"
+        it.mustRunAfter(deleteSnapshots)
```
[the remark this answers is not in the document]
nak5ive: i may be able to set the ordering within the `cleanRecord` task itself and leave this task alone.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
gamepro65: Good call, didn't catch that. Should that have been this property too then?

gamepro65: Oh I see. So then this clean task would need to depend on that task so the property is setup for use

nak5ive: the definition of `snapshotOutputDir` isn't variant-specific, so should be safe to define outside of `variants.all{}`

nak5ive:
> Imagine changing the name of test method in FormBlocker, and deleting 1000+ images and rewriting 998? Maybe I'm prematurely optimizing?

i feel like if you're calling `cleanRecordPD` there's an expectation that it does this.

> Also, can we add tests as part of this PR?

yep! working on that

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: while it isn't today, it probably should be.  for example, see: https://github.com/cashapp/paparazzi/issues/595.

I propose moving this back into the variants.all block and making the `delete` task variant-aware as well (looks like `cleanRecordVariants` is fine as is).

nak5ive: i agree that it should be, but ATM if the `delete` task is made variant-aware, it will delete the same directory multiple times.
It would likely take some more sophisticated task graph wiring to ensure *all* `delete` tasks are run before *any* `record` task in case someone calls the global `cleanRecord`

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -181,6 +191,13 @@ public class PaparazziPlugin : Plugin<Project> {
         it.description = "Record golden images for variant '${variant.name}'"
       }
       recordVariants.configure { it.dependsOn(recordTaskProvider) }
+      val cleanRecordTaskProvider = project.tasks.register("cleanRecordPaparazzi$variantSlug") {
+        it.group = VERIFICATION_GROUP
+        it.description = "Clean and record golden images for variant '${variant.name}'"
+        it.dependsOn(deleteSnapshots, recordTaskProvider)
+        recordTaskProvider.get().mustRunAfter(deleteSnapshots)
```
nak5ive, line 198: it's possible to set the task ordering from within the `cleanRecord` task itself, but unsure if we should?

jrodbx: i'm uncertain as well, given that we have to eagerly get the record task in another task to do so.  Maybe let's just revert to the previous approach?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -707,6 +707,43 @@ class PaparazziPluginTest {
...
+    val snapshot = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete.png")
+    assertThat(snapshot.exists()).isFalse()
+
+    val snapshotWithKeep = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete_keep.png")
+    assertThat(snapshotWithKeep.exists()).isFalse()
+  }
+
+  @Test
+  fun cleanRecord() {
+    val fixtureRoot = File("src/test/projects/delete-snapshots")
```
jrodbx, line 729: i know it's tempting, but reusing projects across different tests may lead to subtle Gradle caching issues.  might be better to copy to a duplicate project?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -707,6 +707,43 @@ class PaparazziPluginTest {
...
+      .withArguments("deletePaparazziSnapshots", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    val snapshotsDir = File(fixtureRoot, "src/test/snapshots").registerForDeletionOnExit()
+
+    val snapshot = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete.png")
+    assertThat(snapshot.exists()).isFalse()
+
+    val snapshotWithKeep = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete_keep.png")
+    assertThat(snapshotWithKeep.exists()).isFalse()
```
jrodbx, line 724: move the file references to before the runner and add assertions that both existed prior to deletion?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -707,6 +707,43 @@ class PaparazziPluginTest {
...
+  @Test
+  fun cleanRecord() {
+    val fixtureRoot = File("src/test/projects/delete-snapshots")
+
+    val result = gradleRunner
+      .withArguments("cleanRecordPaparazziDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    assertThat(result.task(":deletePaparazziSnapshots")).isNotNull()
+    assertThat(result.task(":recordPaparazziDebug")).isNotNull()
```
jrodbx, line 736: nice

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -707,6 +707,43 @@ class PaparazziPluginTest {
...
+
+    val snapshotsDir = File(fixtureRoot, "src/test/snapshots").registerForDeletionOnExit()
+
+    val snapshot = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete.png")
+    assertThat(snapshot.exists()).isFalse()
+
+    val snapshotWithKeep = File(snapshotsDir, "images/app.cash.paparazzi.plugin.test_DeleteTest_delete_keep.png")
+    assertThat(snapshotWithKeep.exists()).isTrue()
```
jrodbx, line 744: similar to the other test method, about assertions before the command?

swankjesse (approved):

jrodbx (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -87,6 +89,15 @@ public class PaparazziPlugin : Plugin<Project> {
...
       it.description = "Record golden images for all variants"
     }
+    val cleanRecordVariants = project.tasks.register("cleanRecordPaparazzi") {
+      it.group = VERIFICATION_GROUP
+      it.description = "Clean and record golden images for all variants"
+    }
+    val deleteSnapshots = project.tasks.register("deletePaparazziSnapshots", Delete::class.java) {
+      it.group = VERIFICATION_GROUP
+      it.description = "Delete all golden images"
+      it.delete(project.fileTree(snapshotOutputDir))
```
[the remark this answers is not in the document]
jrodbx: I think it's fine as is, given that this folder is exclusively owned by Paparazzi (contrast that with say, wiping `src/main/resources`).

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
nak5ive: done

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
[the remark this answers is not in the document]
nak5ive: i went ahead and added includes for `.png` and `.mov`...

jrodbx (approved):

PR #902, opened 2023-05-29
Support verify gifs

BraisGabin:
Paparazzi didn't give support to verify `gifs`. This PR implement that.

PR #1322, opened 2024-03-02
Unify the code to get the path of a golden snapshot

BraisGabin:
Righ now at paparazzi there are 3 different implementations to get the path of a golden snapshot. They are equivalent but their implementations are different. That makes them easy to unsync and that if you need that path in other places you need to reimplement it.

This PR fixes that moving all the logic to a single extension funcion and adding tests to avoid regressions (those paths didn't have any test).

The reason of this PR is because it is needed for  #902 and #917. I assume that moving this to a separate PR will make it easier to review.


jrodbx:
Obsoleted partially by APNG work and 2.0 work

Issue #1277, opened 2024-02-01
Long Text in Accessibility Snapshot is cut off


hrcronin:
**Description**
When using `AccessibilityRenderExtension` and a particular composable has a very long string for the talkback reader, it is cut off in the recorded snapshot.

If there's some additional setting that I'm not using to enable these screenshots to be captured properly, please let me know. I've minimized the issue by doubling the width of the Device Config, but I'm still observing the issue.

**Steps to Reproduce**
This can be reproduced just by using a composable with text and a very long string. See Snippets below.

Composable to Test:
```
data class TextModel(
    val text: String,
    val color: Color,
    val textAlign: TextAlign = TextAlign.Start,
    val overflow: TextOverflow = TextOverflow.Clip,
    val maxLines: Int = Int.MAX_VALUE
)

@Composable
fun TitleText(
    textModel: TextModel
) {
    Text(
        text = textModel.text,
        style = MaterialTheme.typography.titleLarge.copy(
            color = textModel.color
        ),
        textAlign = textModel.textAlign,
        overflow = textModel.overflow,
        maxLines = textModel.maxLines
    )
}

@Composable
fun ColumnWithText(start: Int, end: Int) {
    Column(modifier = Modifier.fillMaxSize()) {
        var count = 0
        for (index in start..end) {
            val string = when (count) {
                1 -> stringResource(id = R.string.two)
                2 -> stringResource(id = R.string.three)
                3 -> stringResource(id = R.string.four)
                else -> stringResource(id = R.string.one)
            }
            TitleText(
                TextModel(
                    text = "$string $index",
                    color = MaterialTheme.colorScheme.primary
                )
            )
            count++
        }
        TitleText(
            textModel = TextModel(
                text = "This is a super duper very long string. Beware of the new line scroll.",
                color = MaterialTheme.colorScheme.primary
            )
        )
    }
}
```

Test Class
```
class PaparazziComposableTest {
    val talkbackPaparazzi = Paparazzi(
            renderExtensions = setOf(AccessibilityRenderExtension())
    )

    @Test
    fun test() {
        paparazzi.snapshot {
              ColumnWithText(3, 6)
        }
    }
```

**Expected behavior**
The full composable view should be visible and all text for the talkback reader should not be cut off. Currently I'm observing that at least a few characters are being cut off from the accessibility print out.

**Additional information:**
- Paparazzi Version: 1.3.0
- OS: Mac OS Ventura
- Compile SDK: Android 34 
[...]

PR #1332, opened 2024-03-08
Fix long content description cut off in Accessibility Snapshots

alexabenedict:
Related to #1277 - Fixes issue where content description gets cut off in Accessibility snapshots when text is long.


geoff-powell (approved):
Thanks for tackling this!

jrodbx (approved):

PR #1347, opened 2024-03-18
Fix gradle caching for Aar directories

### LINKED ISSUE #1342: Gradle Caching issue with AAR exploded resources when bumping dependencies.
**Description**
When bumping a gradle dependency version that is packaged as an AAR with resources, the `:preparePaparazziResources` task doesn't correctly rerun and uses an older cached version from local/remote gradle cache. This can cause issues loading resources when the exploded aar directory no longer exists in your local gradle home cache directory.

**Steps to Reproduce**
- Run sample with a dependency that is an AAR and refrence a resource like drawable or string. Using an older version then latest helps illustrate issue.
- Bump dependency version to latest version (test case is version `0.65.0`)
- Re-run same test and see error message similar to the following.
```
WARNING: Error reading ~/.gradle/caches/transforms-4/2ab825712fd47d3ee279f3575986818d/transformed/android-0.63.0/public.txt: java.nio.file.NoSuchFileException: ~/.gradle/caches/transforms-4/2ab825712fd47d3ee279f3575986818d/transformed/android-0.63.0/public.txt
```
- View `resources.json` file in `build/intermediates/paparazzi` and see old version references in `aarExplodedDirs` key

**Expected behavior**
Test regenerates resources file and tests always pass.

**Additional information:**
- Paparazzi Version: 1.3.3
- OS: Mac OS X
- Compile SDK: 33
- Gradle Version: 8.4
- Android Gradle Plugin Version: 8.2.2

geoff-powell:
Fixes #1342


paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/build.gradle
```diff
@@ -17,6 +17,9 @@ android {
   kotlinOptions {
     jvmTarget = libs.versions.javaTarget.get()
   }
+  dependencies {
+    implementation "androidx.core:core:1.10.0"
+  }
```
geoff-powell, line 22: decided to add the dep here instead of a separate test project.

gamepro65 (approved):
Nice! Only nit/suggestion would be instead of string replace on the actual file, make a 2nd build.gradle.bak or something and rename between the two instead

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/build.gradle.bak
```diff
@@ -0,0 +1,25 @@
...
+  }
+  compileOptions {
+    sourceCompatibility = libs.versions.javaTarget.get()
+    targetCompatibility = libs.versions.javaTarget.get()
+  }
+  kotlinOptions {
+    jvmTarget = libs.versions.javaTarget.get()
+  }
+  dependencies {
+    implementation "androidx.core:core:1.10.1"
```
geoff-powell, line 21: diff between resources in 1.10.0 and 1.10.1 are 0 so this is a good test of the aar directory change needing to cause the prepare task to run again and not use cache.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -69,6 +70,18 @@ public abstract class PrepareResourcesTask : DefaultTask() {
...
+  @get:Input
+  public val aarExplodedResourcesBaseDirs: Provider<List<String>> = project.provider {
+    aarExplodedDirs.relativize(gradleUserHomeDirectory)
+  }
+
+  // Used to require AAR base asset directories are inputs
+  @get:Input
+  public val aarExplodedAssetsBaseDirs: Provider<List<String>> = project.provider {
+    aarAssetDirs.relativize(gradleUserHomeDirectory)
+  }
```
jrodbx, line 83: Nice!

It looks like the original `aarExplodedDirs` and `aarAssetDirs` property types should change instead of adding more properties, so how about removing these changes and instead modifying the originals like so:

```
  @get:Input
  public abstract val aarExplodedDirs: ListProperty<String>
```

then move that logic into the Plugin, there's a bit of prior art there!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -98,9 +111,9 @@ public abstract class PrepareResourcesTask : DefaultTask() {
       resourcePackageNames = resourcePackageNames,
       projectResourceDirs = projectResourceDirs.get(),
       moduleResourceDirs = moduleResourceDirs.get(),
-      aarExplodedDirs = aarExplodedDirs.relativize(gradleUserHomeDirectory),
+      aarExplodedDirs = aarExplodedResourcesBaseDirs.get(),
```
jrodbx, line 114: per the above suggestion, this would become:
```suggestion
      aarExplodedDirs = aarExplodedDirs.get(),
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -98,9 +111,9 @@ public abstract class PrepareResourcesTask : DefaultTask() {
...
       projectAssetDirs = projectAssetDirs.get(),
-      aarAssetDirs = aarAssetDirs.relativize(gradleUserHomeDirectory)
+      aarAssetDirs = aarExplodedAssetsBaseDirs.get()
```
jrodbx, line 116: per the above suggestion, this would become:

```suggestion
      aarAssetDirs = aarAssetDirs.get()
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -188,6 +188,58 @@ class PaparazziPluginTest {
...
+
+      resourcesFile = File(fixtureRoot, "build/intermediates/paparazzi/debug/resources.json")
+      assertThat(resourcesFile.exists()).isTrue()
+      resourceFileContents = resourcesFile.readLines()
+      assertThat(resourceFileContents.any { it.contains("core-1.10.1") }).isTrue()
+    } finally {
+      assertThat(buildGradleFile.renameTo(buildGradleFileNew)).isTrue()
+      assertThat(buildGradleFileOriginalBackup.renameTo(buildGradleFile)).isTrue()
+    }
+  }
```
jrodbx, line 241: this looks like what `verifyResourcesUpdatedWhenExternalDependencyChanges` intended to test, but maybe the versioning was the key part?  can we fix/reuse that test per these changes, or is there a noteworthy difference to justify another test?

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/build.gradle
jrodbx: see `verify-update-aar-resources-change/build.gradle`

paparazzi-gradle-plugin/src/test/projects/prepare-resources-task-caching/build.gradle.bak
```diff
@@ -0,0 +1,25 @@
+plugins {
```
jrodbx, line 1: this file is repetitive, let's use the system property approach like `verify-update-aar-resources-change/build.gradle` or even better, let's update that test project instead adding a new one.

....unless there's a clear difference between a local aar and repository aar project?  If so, I'd still like the tests to clearly surface that difference, if possible.

jrodbx (commented):
Looks great, thanks.  Silly relative directories getting me again!  I have a few comments.

paparazzi-gradle-plugin/api/paparazzi-gradle-plugin.api
```diff
@@ -12,7 +12,9 @@ public abstract class app/cash/paparazzi/gradle/PaparazziPlugin$PaparazziTask :
 public abstract class app/cash/paparazzi/gradle/PrepareResourcesTask : org/gradle/api/DefaultTask {
 	public fun <init> ()V
 	public abstract fun getAarAssetDirs ()Lorg/gradle/api/file/ConfigurableFileCollection;
+	public final fun getAarExplodedAssetsBaseDirs ()Lorg/gradle/api/provider/Provider;
 	public abstract fun getAarExplodedDirs ()Lorg/gradle/api/file/ConfigurableFileCollection;
+	public final fun getAarExplodedResourcesBaseDirs ()Lorg/gradle/api/provider/Provider;
```
jrodbx, line 17:
```suggestion
	public abstract fun getAarAssetDirs ()Lorg/gradle/api/provider/ListProperty;
	public abstract fun getAarExplodedDirs ()Lorg/gradle/api/provider/ListProperty;
```

paparazzi-gradle-plugin/src/test/projects/verify-update-aar-resources-change/build.gradle
```diff
@@ -20,7 +20,9 @@ dependencies {
   def isFirstRun = System.getProperty("isFirstRun", "false").toBoolean()
   if (isFirstRun) {
     implementation files('libs/external1.aar')
+    implementation "androidx.core:core:1.10.0"
```
jrodbx, line 23:
```suggestion
    implementation 'androidx.core:core:1.10.0'
```

paparazzi-gradle-plugin/src/test/projects/verify-update-aar-resources-change/build.gradle
```diff
@@ -20,7 +20,9 @@ dependencies {
...
   } else {
     implementation files('libs/external2.aar')
+    implementation "androidx.core:core:1.10.1"
```
jrodbx, line 26:
```suggestion
    implementation 'androidx.core:core:1.10.1'
```

jrodbx (approved):

jrodbx (commented):
one thought: if the underlying issue is with version bumps to dependencies with 0 diff in res content, then perhaps updating external2.aar to contain the same resource content as external1.aar in `verify-update-aar-resources-change` might be enough?  

fine to do it in a follow-up, but a nice benefit to keeping it a local aar solution is avoiding the extra network call for a dependency fetch in plugin tests where possible.

PR #1351, opened 2024-03-20
Add ability to transform aar dependencies to jar in Paparazzi jvm project

### LINKED ISSUE #783: Switch from checking in jars to unpacking aars
We always wanted to do this but were too ~lazy~ rushed to write the artifact transform at the time.

Procrastination _always_ pays off. Now we can copy someone else's work instead: https://github.com/Kotlin/kotlinx.coroutines/blob/master/buildSrc/src/main/kotlin/UnpackAar.kt.

geoff-powell:
Fixes #783 so we no longer need to manually include androidX jars in project to compile


buildSrc/src/main/kotlin/AarAndJarCompatible.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.gradle.api.attributes.AttributeCompatibilityRule
+import org.gradle.api.attributes.CompatibilityCheckDetails
+import org.gradle.api.attributes.LibraryElements
+
+@CacheableRule
+abstract class AarAndJarCompatible : AttributeCompatibilityRule<LibraryElements> {
+  override fun execute(t: CompatibilityCheckDetails<LibraryElements>) {
+    val consumer = t.consumerValue ?: return
+    val producer = t.producerValue ?: return
+    if (consumer.name == LibraryElements.JAR && producer.name == "aar") {
```
geoff-powell, line 13: Fixes inner dependencies that are `aar` and jvm project requires `jar`

buildSrc/src/main/kotlin/AndroidJvmAndJvmCompatible.kt
```diff
@@ -0,0 +1,17 @@
...
+
+@CacheableRule
+abstract class AndroidJvmAndJvmCompatible : AttributeCompatibilityRule<KotlinPlatformType> {
+  override fun execute(t: CompatibilityCheckDetails<KotlinPlatformType>) {
+    val consumer = t.consumerValue ?: return
+    val producer = t.producerValue ?: return
+    if (consumer == KotlinPlatformType.jvm && producer == KotlinPlatformType.androidJvm) {
+      t.compatible()
+    }
+  }
```
geoff-powell, line 16: Allows for jvm projects to continue to use aar dependencies built for android on jvm project

buildSrc/build.gradle
```diff
@@ -0,0 +1,15 @@
+plugins {
+  id "org.gradle.kotlin.kotlin-dsl" version "4.3.0"
+}
+
+repositories {
+  gradlePluginPortal()
```
JakeWharton, line 6: nit: could be
```suggestion
  mavenCentral()
```

buildSrc/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+  google()
+}
+
+dependencies {
+  // Needed to fix issue with android kotlin project
+  implementation("com.android.tools.build:gradle:8.2.2")
+  implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.22")
```
JakeWharton, line 13: No `libs.versions.toml` in this repo? You can point at the enclosing project with https://github.com/cashapp/redwood/blob/trunk/build-support/settings.gradle#L6-L12

buildSrc/src/main/kotlin/UnpackAar.kt
```diff
@@ -0,0 +1,89 @@
...
+  @get:InputArtifact
+  abstract val inputArtifact: Provider<FileSystemLocation>
+
+  override fun transform(outputs: TransformOutputs) {
+    ZipFile(inputArtifact.get().asFile).use { zip ->
+      zip.entries().asSequence()
+        .filter { !it.isDirectory }
+        .filter { it.name.endsWith(".jar") }
+        .forEach {
+          val outputCacheDirectory = outputs.file(it.name).toString().split("caches/")[1]
```
JakeWharton, line 76: I'm a bit nervous about this as opposed to a local folder in `layout.buildDirectory`.

JakeWharton (approved):

buildSrc/src/main/kotlin/UnpackAar.kt
```diff
@@ -0,0 +1,89 @@
...
+abstract class UnpackAar : TransformAction<TransformParameters.None> {
+
+  @get:InputArtifact
+  abstract val inputArtifact: Provider<FileSystemLocation>
+
+  override fun transform(outputs: TransformOutputs) {
+    ZipFile(inputArtifact.get().asFile).use { zip ->
+      zip.entries().asSequence()
+        .filter { !it.isDirectory }
+        .filter { it.name.endsWith(".jar") }
```
JakeWharton, line 74: This is too broad. You want to match on `classes.jar` _exactly_. The aar spec allows other `.jar`s that you definitely don't want on your classpath: https://developer.android.com/studio/projects/android-library#aar-contents

gradle/libs.versions.toml
```diff
@@ -21,7 +21,14 @@ layoutlibPrebuiltSha = "a73a761"
...
 androidx-appcompat = { module = "androidx.appcompat:appcompat", version = "1.6.1" }
-androidx-lifecycleCommon = { module = "androidx.lifecycle:lifecycle-common", version = "2.7.0" }
+androidx-activity = { module = "androidx.activity:activity", version = "1.5.0" }
+androidx-savedstate = { module = "androidx.savedstate:savedstate", version = "1.2.1" }
+androidx-lifecycle-common = { module = "androidx.lifecycle:lifecycle-common", version = "2.7.0" }
+androidx-compose-ui-geometry = { module = "androidx.compose.ui:ui-geometry", version = "1.5.0" }
+androidx-compose-ui-text = { module = "androidx.compose.ui:ui-text", version = "1.5.0" }
+androidx-compose-ui-android = { module = "androidx.compose.ui:ui-android", version = "1.5.0" }
+androidx-compose-runtime-android = { module = "androidx.compose.runtime:runtime-android", version = "1.5.0" }
+androidx-lifecycle-runtime = { module = "androidx.lifecycle:lifecycle-runtime", version = "2.6.1" }
```
jrodbx, line 31: I believe some of these can be removed

buildSrc/src/main/kotlin/UnpackAar.kt
```diff
@@ -0,0 +1,96 @@
...
+import java.util.zip.ZipEntry
+import java.util.zip.ZipFile
+
+// Attributes used by aar dependencies
+private val artifactType = Attribute.of("artifactType", String::class.java)
+private val unpackedAar = Attribute.of("unpackedAar", Boolean::class.javaObjectType)
+
+fun Project.configureAar() =
+  configurations.configureEach {
+    // Causes issue with spotless
```
jrodbx, line 26: Is there a tracking issue on spotless?

buildSrc/src/main/kotlin/AarAndJarCompatible.kt
```diff
@@ -0,0 +1,20 @@
+package app.cash.paparazzi.gradle
+
+import org.gradle.api.artifacts.CacheableRule
+import org.gradle.api.attributes.AttributeCompatibilityRule
+import org.gradle.api.attributes.CompatibilityCheckDetails
+import org.gradle.api.attributes.LibraryElements
+
+@CacheableRule
+abstract class AarAndJarCompatible : AttributeCompatibilityRule<LibraryElements> {
```
jrodbx, line 9: I believe compatibility rules can be avoided, if you create custom configurations set up with the appropriate variant attribute matching.  This would avoid messing with the existing configurations and is cleaner.

buildSrc/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+}
+
+repositories {
+  mavenCentral()
+  google()
+}
+
+dependencies {
+  // Needed to fix issue with android kotlin project
+  implementation libs.plugin.android
```
jrodbx, line 12: If we can eliminate the compatibility rules, we can avoid buildSrc altogether and instead inline the transforms and wiring into build.gradle.

buildSrc/src/main/kotlin/AarAndJarCompatible.kt
geoff-powell: Hmm, my gradle knowledge is lacking here with how variant attribute matching works with a custom configuration.

geoff-powell: I was able to get this to work. One difference in my current implementation is that child dependencies don't get resolved and transformed like they do with the compatibility rules.

I still think this method makes more logical sense from a configuration standpoint

geoff-powell: ☝🏽 scratch that. I had set the `isTransitive` property of the configuration to be false. Setting it to true matches the desired implementation of the compatibility rules attempt.

geoff-powell:
Updated this to make a simple configuration defined in a separate gradle file.

BrianGardnerAtl (commented):
I think it looks good but my gradle knowledge is lacking for a full approval

paparazzi/build.gradle
```diff
@@ -1,10 +1,10 @@
-import app.cash.paparazzi.gradle.UnpackAarKt
 import org.jetbrains.kotlin.gradle.plugin.AbstractKotlinPluginKt
 
 apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'org.jetbrains.dokka'
 apply plugin: 'com.vanniktech.maven.publish'
 apply plugin: 'dev.drewhamilton.poko'
+apply from: "customConfiguration.gradle"
```
geoff-powell, line 7: looks like gradle file translation from kotlin is not working as expected. Causes issue with downstream IR kotlin compilation.

geoff-powell: required me to add `implementation libs.plugin.kotlin` to the buildSrc/build.gradle file to get compilation working.

nak5ive (approved):

gradle/aarToJarConfiguration.gradle
```diff
@@ -0,0 +1,125 @@
...
+  project.configurations.compileOnly.dependencies.add(
+    create(
+      project.configurations.compileOnlyAarAsJar.incoming.artifactView {
+        it.attributes.attribute(
+          artifactType, "aarAsJar"
+        )
+      }.files
+    )
+  )
+}
```
jrodbx, line 64: we can dry this up a bit:

```
  def aarAsJar =  project.configurations.create("${configuration}AarAsJar") {
      attributes.attribute(...)
    }
    .incoming
    .artifactView { viewConfiguration ->
      viewConfiguration.attributes.attribute(...)
    }
    .files

  ...
  project.configurations
    .getByName(configuration)
    .dependencies
    .add(project.dependencies.create(aarAsJar))
```

gradle/aarToJarConfiguration.gradle
```diff
@@ -0,0 +1,125 @@
+//file:noinspection GroovyAssignabilityCheck
+import com.android.build.api.attributes.BuildTypeAttr
+import org.gradle.work.DisableCachingByDefault
```
jrodbx, line 3: can we inline this file into `paparazzi/build.gradle` and get rid of `buildSrc` altogether?  the SDK should be the only consumer of this `aarAsjar` transformation.

buildSrc/build.gradle
jrodbx: is `BuildTypeAttr` necessary?  would setting [AndroidArtifacts.ARTIFACT_TYPE, AndroidArtifacts.TYPE_CLASSES_JAR] as a key-value to the attribute container be enough to assist variant matching?

jrodbx (commented):
looking good!

I think we can reduce this a bit and keep the logic to `paparazzi/build.gradle`.

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziViewOwners.kt
```diff
@@ -25,6 +25,6 @@ internal class PaparazziSavedStateRegistryOwner(
 internal class PaparazziOnBackPressedDispatcherOwner(
   private val lifecycleOwner: LifecycleOwner
 ) : OnBackPressedDispatcherOwner, LifecycleOwner by lifecycleOwner {
-  override val onBackPressedDispatcher: OnBackPressedDispatcher
-    get() = OnBackPressedDispatcher { /* Swallow all back-presses. */ }
+  override fun getOnBackPressedDispatcher(): OnBackPressedDispatcher =
+    OnBackPressedDispatcher { /* Swallow all back-presses. */ }
```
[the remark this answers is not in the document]
geoff-powell: Yep, we were inadvertently bumping dependeencies by default. Keeping this in this pr and then going to bump the androidX deps in a follow up

gradle/aarToJarConfiguration.gradle
geoff-powell: yep, now its simplified

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/AarTestUtils.kt
```diff
@@ -68,16 +71,19 @@ internal fun makeAarRepositoryFromAarArtifact(
...
     ZipOutputStream(aarPath.outputStream().buffered()).use { zip ->
-      sourceDirectory.visitFileTree {
-        onVisitFile { file, _ ->
-          val relativePath = sourceDirectory.relativize(file).invariantSeparatorsPathString
-          val entry = ZipEntry(relativePath)
-          zip.putNextEntry(entry)
-          zip.write(file.readBytes())
-          zip.closeEntry()
-          return@onVisitFile CONTINUE
+      Files.walkFileTree(
```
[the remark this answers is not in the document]
geoff-powell: yep, unfortunately a side-effect of the older androidX deps we pulled. I have a follow up here that updates those deps and reverts these changes. https://github.com/cashapp/paparazzi/pull/1373

paparazzi/build.gradle
```diff
@@ -154,3 +164,91 @@ tasks.withType(Test).configureEach {
...
+    try {
+      def file = getPrimaryInput().get().asFile
+      def aarFilename = file.name.substring(0, file.name.lastIndexOf('.'))
+      zip = new ZipFile(file)
+      def entries = zip.entries()
+      while (entries.hasMoreElements()) {
+        def item = entries.nextElement()
+        if (item.isDirectory()) continue
+
+        if (item.name.endsWith("classes.jar")) {
```
jrodbx, line 235:
```suggestion
        if (item.name == "classes.jar") {
```

geoff-powell:
Merging this to unblock some of Al's work. Happy to put up any fixes in a follow up PR

PR #1355, opened 2024-03-22
Add support for render mode SHRINK for AccessibilityRenderExtension

### LINKED ISSUE #1350: `AccessibilityRenderExtension` doesn't work correctly with `RenderingMode.SHRINK`
**Description**
Using the `AccessibilityRenderExtension` is useful for analyzing accessibility information, but the information added to the side of the screenshot is cropped out when the `SHRINK` rendering mode is used.

**Steps to Reproduce**
Here is an example test:
```kotlin
class ExampleTest {

    @get:Rule
    val paparazzi = Paparazzi(
        renderingMode = RenderingMode.SHRINK,
    )

    @Test
    fun exampleTest() {
        paparazzi.snapshot {
            Box(
                modifier = Modifier
                    .background(Color.White)
                    .padding(16.dp)
            ) {
                Text(text = "Hello, world!")
            }
        }
    }

}
```

**Expected behavior**
I would expect for the accessibility information on the side of the screenshot to be included in the cropped screenshot, that is to say that the crop would take into account the size of the accessibility information, even if it was taller than the actual content.

**Additional information:**
- Paparazzi Version: 1.3.3
- OS: MacOS
- Compile SDK: 34
- Gradle Version: 8.6
- Android Gradle Plugin Version: 8.3.0

**Screenshots**
Screenshot using `SHRINK`:
image

Screenshot using `NORMAL`:
image

I've given fixing this a look and it seems like one of the challenges here is that the accessibility information is a canvas element which can't easily use things like `WRAP_CONTENT`.

geoff-powell:
Fixes #1350


paparazzi-gradle-plugin/src/test/projects/verify-accessibility-render-extension/src/test/resources/accessibility.png
jrodbx, line 1: Curious: why move from the unit test to the plugin test?  It'd be nice keeping the former, unless there's a good reason.

Basically, the commit prior: https://github.com/cashapp/paparazzi/compare/c3e10703ecd6daab2a6f30e8118c511680437481..39280d0a467518e4b36799cfc25ec6c8ea16ad19

geoff-powell: Cool, got a comment in slack about moving it. I reverted that change to make the change smaller and straight forward

VirtualParticle:
Thank you for working on this! One thing I'll note is that it seems like this still runs into an issue if the content description takes up more space than the actual content. Take this test, for example:
```kotlin
@Test
fun `verify long content description`() {
  paparazzi.snapshot {
    Text(
      text = "Text",
      modifier = Modifier
        .padding(10.dp)
        .semantics {
          contentDescription = "This is a very long content description"
        },
    )
  }
}
```
It produces this screenshot:
image
Ideally, the test would recognize that the accessibility info is bigger and the screenshot would expand to fit the content description. Might be out-of-scope for this though.

paparazzi/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionShrinkTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import app.cash.paparazzi.SnapshotHandler
+import app.cash.paparazzi.internal.ImageUtils
+import com.android.ide.common.rendering.api.SessionParams
+import org.junit.Rule
+import org.junit.Test
+import java.awt.image.BufferedImage
+import java.io.File
+import javax.imageio.ImageIO
+
+class AccessibilityRenderExtensionShrinkTest {
```
jrodbx, line 16: nit: can we combine this with `AccessibilityRenderExtensionTest` + TestParameterInjector?  Looks like `renderingMode` is the only argument that differs between the two.

jrodbx (approved):

geoff-powell:
> Thank you for working on this! One thing I'll note is that it seems like this still runs into an issue if the content description takes up more space than the actual content. Take this test, for example:
> 
> ```kotlin
> @Test
> fun `verify long content description`() {
>   paparazzi.snapshot {
>     Text(
>       text = "Text",
>       modifier = Modifier
>         .padding(10.dp)
>         .semantics {
>           contentDescription = "This is a very long content description"
>         },
>     )
>   }
> }
> ```
> 
> It produces this screenshot: image Ideally, the test would recognize that the accessibility info is bigger and the screenshot would expand to fit the content description. Might be out-of-scope for this though.

Ya so that issue is due to the text having to be rendered after draw and not being part of the layout scope. Currently paparazzi render extensions execute on 1 layout pass. Unfortunately, we can only gather compose semantics after layout. Still thinking through other ways to fix this limitation

PR #1314, opened 2024-03-01
Bump LayoutLib to Iguana

jrodbx:
Iguana provides API 34 support!

PR #1360, opened 2024-03-28
Bump to AGP 8.3.1

jrodbx:
Intentionally bumping only AGP, as this moves some of the changeset out of https://github.com/cashapp/paparazzi/pull/1314.


geoff-powell:
Nice 🎉

PR #1228, opened 2024-01-14
use Google's published layoutlib

gabrielittner:
Closes #1221.
Closes #884.
Closes #717.

- Replaces the manually published layoutlib modules with the new artifacts published by Google.
- In 14.0.3 a new `layoutlib-resources` artifact was introduced. This contains the icu, keyboard and font files that previously were in the main layoutlib artifacts. It also contains a `res` folder which can be used instead of reading resources from the SDK platform directory.
- In 14.0.4 icu, keyboard and font files moved back to the main layoutlib artifacts. `layoutlib-resources` now only contains the `res` folder which moved to a top level directory inside the jar.
- ~Initially I've also upgraded `layoutlib-api` to the Jellyfish preview versions. However it's not actually needed for everything to work, so I downgraded it again. Because of that this PR does not depend on any alpha/beta versions, but I guess the `14.0.x` layoutlib artifacts could still be considered as not stable until AS Jellyfish is stable.~

PR #1362, opened 2024-03-28
Migrate MacOS runner from macOS 12 Intel to macOS 14 M1

jrodbx:
The divergence between MacOS Intel and M1 renderings is unreasonably high (@gabrielittner points that out here: https://github.com/cashapp/paparazzi/pull/1228/files#r1473442070).

Eventually, this will be GHA's default latest Mac runner offering (see comment in build.yml), so feels OK to bump now.

Note that this force bumps the tests in the `:paparazzi` module to platform 33, because platform 31 is not part into the macOS-14 runner.

See:
* https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#android
* https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#android

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          macOS-latest,
+          # Migrate back to macOS-latest once 'latest' workflow label is updated (April – June 2024)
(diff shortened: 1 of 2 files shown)


gamepro65 (approved):
lgtm

PR #1363, opened 2024-03-29
Extract FrameHandler from PaparazziSdk

gamepro65:
We are calling `handler.use { }` inside `takeSnapshots` as well as `frameHandler?.close()` in sdk's teardown causing us to double close the frameHandler.

Instead of continuing to use FrameHandler directly in PaparazziSdk this PR aims to expose a frame callback that can be utilized by Paparazzi to forward to it's self owned FrameHandler avoiding any accidental cleanup during sdk's teardown.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
@@ -40,7 +40,6 @@ import androidx.compose.ui.platform.createLifecycleAwareWindowRecomposer
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.setViewTreeLifecycleOwner
 import androidx.savedstate.setViewTreeSavedStateRegistryOwner
-import app.cash.paparazzi.SnapshotHandler.FrameHandler
 import app.cash.paparazzi.accessibility.AccessibilityRenderExtension
 import app.cash.paparazzi.agent.InterceptorRegistrar
 import app.cash.paparazzi.internal.ImageUtils
@@ -93,12 +92,12 @@ public class PaparazziSdk @JvmOverloads constructor(
   private val renderExtensions: Set<RenderExtension> = setOf(),
   private val supportsRtl: Boolean = false,
   private val showSystemUi: Boolean = false,
-  private val validateAccessibility: Boolean = false
+  private val validateAccessibility: Boolean = false,
+  private val onNewFrame: (BufferedImage) -> Unit
 ) {
   private val logger = PaparazziLogger()
   private lateinit var renderSession: RenderSessionImpl
   private lateinit var bridgeRenderSession: RenderSession
-  private var frameHandler: FrameHandler? = null
 
   public val layoutInflater: LayoutInflater
     get() = RenderAction.getCurrentContext().getSystemService("layout_inflater") as BridgeInflater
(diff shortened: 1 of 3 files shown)


geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -165,34 +163,29 @@ public class PaparazziSdk @JvmOverloads constructor(
...
 
-  public fun setFrameHandler(handler: FrameHandler) {
-    frameHandler = handler
-  }
-
   public fun <V : View> inflate(@LayoutRes layoutId: Int): V =
     layoutInflater.inflate(layoutId, null) as V
 
-  public fun snapshot(composable: @Composable () -> Unit) {
+  public fun snapshot(frameCallback: (BufferedImage) -> Unit, composable: @Composable () -> Unit) {
```
jrodbx, line 174: I'm not sure that passing the frame callback to every snapshot/gif every method is ideal.

preferred:
```
val frameCallback = {...}

sdk = SDK(frameCallback)
// or
// sdk.setFrameCallback(frameCallback)
sdk.setup()
sdk.snapshot(view, ...)
// update view model
sdk.snapshot(view, ...)
sdk.shutdown()
```

vs:
```
val frameCallback = {...}

sdk = SDK()
sdk.setup()
sdk.snapshot(view, ..., frameCallback)
// update view model
sdk.snapshot(view, ..., frameCallback) // <== why pass this again?
sdk.shutdown()
```

Now I realize it's also not ideal to pass the frame callback via a setter vs a constructor argument, but this is primarily because the current use case is the Junit Rule, i.e., the frame handler takes the name of the method which is only known after the rule (and therefore the sdk) is created, thus the deferred state management.

I *do* like moving `handler.use` out of the SDK, but it looks like:
* leaving the field variable
* keeping the IllegalStateException check in takeSnapshots
* removing `frameHandler?.close()`

fixes the double close problem.

gamepro65: > preferred:
> 
> ```
> val frameCallback = {...}
> 
> sdk = SDK(frameCallback)
> // or
> // sdk.setFrameCallback(frameCallback)
> sdk.setup()
> sdk.snapshot(view, ...)
> // update view model
> sdk.snapshot(view, ...)
> sdk.shutdown()
> ```

Took a stab at the preferred path, lmk what you think @jrodbx

jrodbx: nice!

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -95,12 +94,12 @@ public class PaparazziSdk @JvmOverloads constructor(
   private val renderExtensions: Set<RenderExtension> = setOf(),
   private val supportsRtl: Boolean = false,
   private val showSystemUi: Boolean = false,
-  private val validateAccessibility: Boolean = false
+  private val validateAccessibility: Boolean = false,
+  private val frameCallback: (BufferedImage) -> Unit
```
jrodbx, line 98:
```suggestion
  private val onNewFrame: (BufferedImage) -> Unit
```

Issue #1060, opened 2023-08-25
Resources$NotFoundException. If call resource (that res file is in app module) from dynamic feature module


juckrit:
**Description**
I have composable function in dynamicFeature Module that show Image that image file is in app module
when I run "gradle recordPaparazziDebug" the test fail with Resources$NotFoundException

**Steps to Reproduce**
1. clone this sample project https://github.com/juckrit/Poc_Paparazzi
2. run gradle task  "recordPaparazziDebug"

**Expected behavior**
The test should pass

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: Mac Os 13.3.1 (22E261)
- Compile SDK:  33
- Gradle Version: 8.0
- Android Gradle Plugin Version: 8.1.0

**Screenshots**
<img width="1440" alt="Screenshot 2566-08-25 at 20 42 50" src="https://github.com/cashapp/paparazzi/assets/35564862/61a5ce98-2740-4044-90e7-b820e40d61d6">

jacobtabak:
I have a similar issue.

I'm not using a dynamic feature module, but the resources being referenced are in a different module.

The problem manifested after upgrading to Gradle 8.3 and AGP 8.1.1. 

`android.content.res.Resources$NotFoundException: Could not resolve resource value: 0x7F06027E.
	at app//android.content.res.Resources_Delegate.throwException(Resources_Delegate.java:1161)
	at app//android.content.res.Resources_Delegate.throwException(Resources_Delegate.java:1137)
	at app//android.content.res.Resources_Delegate.throwException(Resources_Delegate.java:1141)
	at app//android.content.res.Resources_Delegate.getColorStateList(Resources_Delegate.java:264)
	at app//android.content.res.Resources_Delegate.getColor(Resources_Delegate.java:242)
	at app//android.content.res.Resources.getColor(Resources.java:1065)
	at app//androidx.compose.ui.res.ColorResourceHelper.getColor-WaAFU9c(ColorResources.android.kt:51)
	at app//androidx.compose.ui.res.ColorResources_androidKt.colorResource(ColorResources.android.kt:40)
	at app//com.seatgeek.android.compose.theme.SeatGeekTheme$colorsCache$1.invoke(SeatGeekTheme.kt:64)
	at app//com.seatgeek.android.compose.theme.SeatGeekTheme$colorsCache$1.invoke(SeatGeekTheme.kt:61)
[...]

PR #1367, opened 2024-04-01
Fixes Resources$NotFoundException when accessing app module resource from dynamic feature module

kevinzheng-ap:
Related to https://github.com/cashapp/paparazzi/issues/1060

Fixes Resources$NotFoundException when accessing app module resource from dynamic feature module.

The cause is in [PublishingSpecs.kt](build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt), `PACKAGED_RES` and `SYMBOL_LIST_WITH_PACKAGE_NAME` are not properly associated with correspondent ArtifactType when publishing as `ComponentTypeImpl.BASE_APK` from `application` module. So these artifacts are not available for `dynamic feature` module during the execution of `preparePaparazziDebugResources`.

The proposed solution is to replicate the logic in [ArtifactPublishingUtil.kt](https://cs.android.com/android-studio/platform/tools/base/+/65c212b013f1c94d8db8f488f50a53d6b3ace64f:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/scope/ArtifactPublishingUtil.kt;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f;l=101) to map above artifacts to their respective ArtifactType like what [ComponentTypeImpl.LIBRARY](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt;drc=83b16ff12bba875684d4074db6f8056293bb9036;l=229) does. However, this approach requires application module to apply `PaparazziPlugin`, even if it does not have any snapshot tests of its own.


paparazzi-gradle-plugin/src/test/projects/supports-dynamic-feature-modules/app/build.gradle
```diff
@@ -15,3 +17,9 @@ android {
 
   dynamicFeatures = [':dynamic_feature']
 }
+
+dependencies {
+  implementation files('libs/external.aar')
+}
+
```
kevinzheng-ap, line 24: Unlike `library` module, `application` module does not generate `symbol_list_with_package_name/[variant]/package-aware-r.txt` when there is no dependency. So I have to add one dependency in the test. 

In reality, `application` module will always have dependencies, so there is no issue about that. We do not have to check if `symbol_list_with_package_name/[variant]/package-aware-r.txt` exists

geoff-powell (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -333,3 +450,20 @@ public class PaparazziPlugin : Plugin<Project> {
...
+)
+
+private val API_AND_RUNTIME_ELEMENTS: List<AndroidArtifacts.PublishedConfigType> =
+  listOf(API_ELEMENTS, RUNTIME_ELEMENTS)
+private val RUNTIME_ELEMENTS_ONLY: List<AndroidArtifacts.PublishedConfigType> =
+  listOf(RUNTIME_ELEMENTS)
+
+private fun getApplicationAdditionalSpec() = setOf(
+  OutputSpec(InternalArtifactType.PACKAGED_RES, ArtifactType.ANDROID_RES, RUNTIME_ELEMENTS_ONLY),
+  OutputSpec(InternalArtifactType.SYMBOL_LIST_WITH_PACKAGE_NAME, ArtifactType.SYMBOL_LIST_WITH_PACKAGE_NAME)
```
jrodbx, line 468: inline

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -333,3 +450,20 @@ public class PaparazziPlugin : Plugin<Project> {
...
+  val outputType: Artifact.Single<out FileSystemLocation>,
+  val artifactType: ArtifactType,
+  val publishedConfigTypes: List<AndroidArtifacts.PublishedConfigType> = API_AND_RUNTIME_ELEMENTS,
+  val libraryElements: String? = null
+)
+
+private val API_AND_RUNTIME_ELEMENTS: List<AndroidArtifacts.PublishedConfigType> =
+  listOf(API_ELEMENTS, RUNTIME_ELEMENTS)
+private val RUNTIME_ELEMENTS_ONLY: List<AndroidArtifacts.PublishedConfigType> =
+  listOf(RUNTIME_ELEMENTS)
```
jrodbx, line 464: inline

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -77,6 +103,97 @@ public class PaparazziPlugin : Plugin<Project> {
     }
   }
 
+  /**
+   * Ported from: [ArtifactPublishingUtil.kt](https://cs.android.com/android-studio/platform/tools/base/+/65c212b013f1c94d8db8f488f50a53d6b3ace64f:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/scope/ArtifactPublishingUtil.kt;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f;l=101)
+   *
+   * Publish intermediate artifacts in the BuildArtifactsHolder in addition to default PublishingSpecs.
+   *
+   * ignore original publishInfo: VariantPublishingInfo? parameter as we are not using publication config
```
jrodbx, line 111: this can be passed as `null`, since the signature of `com.android.build.gradle.internal.scope.ArtifactPublishingUtil.kt#publishBuildArtifacts` is:

```
fun publishBuildArtifacts(
    creationConfig: ComponentCreationConfig,
    publishInfo: VariantPublishingInfo?
) {
```

jrodbx (commented):
Ok, so this is an awesome solution given your source digging into AGP internals, but I'm concerned this depends on internals too much and will break every time we bump AGP.

I'm curious if there is a DSL one level up that exposes this to plugin or build script authors.  In other words,

Looking at line 105 of `ArtifactPublishingUtil` in AGP 8.2.2:
```
fun publishBuildArtifacts(
    creationConfig: ComponentCreationConfig,
    publishInfo: VariantPublishingInfo?
) {
    for (outputSpec in PublishingSpecs.getVariantPublishingSpec(creationConfig.componentType).outputs) {
```
ultimately, it looks like this approach replaces `publishInfo` with the `additionalSpecs` needed to drive this business logic correctly.  Is there a way via DSL to configure the variant publishing spec such that:
```
PublishingSpecs.getVariantPublishingSpec(creationConfig.componentType).outputs
```
in AGP internal will execute the desired path in your implementation of `publishBuildArtifacts`?

jrodbx:
> The proposed solution is to replicate the logic in [ArtifactPublishingUtil.kt](https://cs.android.com/android-studio/platform/tools/base/+/65c212b013f1c94d8db8f488f50a53d6b3ace64f:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/scope/ArtifactPublishingUtil.kt;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f;l=101) to map above artifacts to their respective ArtifactType like what [ComponentTypeImpl.LIBRARY](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt;drc=83b16ff12bba875684d4074db6f8056293bb9036;l=229) does. 

Why mimic what ComponentTypeImpl.LIBRARY does?  Is a dynamic feature considered a LIBRARY component type for the purposes of publishing?

> However, this approach requires application module to apply `PaparazziPlugin`, even if it does not have any snapshot tests of its own.

How come?  Any ideas how we can avoid this?  Android Studio doesn't require this, so it'd be an odd user experience to require this.

kevinzheng-ap:
> Ok, so this is an awesome solution given your source digging into AGP internals, but I'm concerned this depends on internals too much and will break every time we bump AGP.
> 
> I'm curious if there is a DSL one level up that exposes this to plugin or build script authors. In other words,
> 
> Looking at line 105 of `ArtifactPublishingUtil` in AGP 8.2.2:
> 
> ```
> fun publishBuildArtifacts(
>     creationConfig: ComponentCreationConfig,
>     publishInfo: VariantPublishingInfo?
> ) {
>     for (outputSpec in PublishingSpecs.getVariantPublishingSpec(creationConfig.componentType).outputs) {
> ```
> 
> ultimately, it looks like this approach replaces `publishInfo` with the `additionalSpecs` needed to drive this business logic correctly. Is there a way via DSL to configure the variant publishing spec such that:
> 
> ```
> PublishingSpecs.getVariantPublishingSpec(creationConfig.componentType).outputs
> ```
> 
> in AGP internal will execute the desired path in your implementation of `publishBuildArtifacts`?

[PublishingSpecs.getVariantPublishingSpec](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt;l=373-375;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f?q=PublishingSpecs&ss=android-studio%2Fplatform%2Ftools%2Fbase#:~:text=372-,373,-374) is referring to a private static map which is initialized with [hardcoded builder](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt;l=136-175;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f?q=PublishingSpecs&ss=android-studio%2Fplatform%2Ftools%2Fbase#:~:text=135-,136,-137). I did not find a way to intercept the initialization and add more artifact type there. And it is immutable after initialization.

kevinzheng-ap:
> > The proposed solution is to replicate the logic in [ArtifactPublishingUtil.kt](https://cs.android.com/android-studio/platform/tools/base/+/65c212b013f1c94d8db8f488f50a53d6b3ace64f:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/scope/ArtifactPublishingUtil.kt;drc=65c212b013f1c94d8db8f488f50a53d6b3ace64f;l=101) to map above artifacts to their respective ArtifactType like what [ComponentTypeImpl.LIBRARY](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/publishing/PublishingSpecs.kt;drc=83b16ff12bba875684d4074db6f8056293bb9036;l=229) does.
> 
> Why mimic what ComponentTypeImpl.LIBRARY does? Is a dynamic feature considered a LIBRARY component type for the purposes of publishing?
> 
> > However, this approach requires application module to apply `PaparazziPlugin`, even if it does not have any snapshot tests of its own.
> 
> How come? Any ideas how we can avoid this? Android Studio doesn't require this, so it'd be an odd user experience to require this.

The root cause is not dynamic feature itself, but its dependency `application` module. `PACKAGED_RES` and `SYMBOL_LIST_WITH_PACKAGE_NAME` are not provided for `application` module who is published as ComponentTypeImpl.BASE_APK. Then these artifacts are not available for dynamic feature to consume. 

So the propose solution is actually to mimic what  ComponentTypeImpl.LIBRARY does for application module. In order to do that, we have to apply `PaparazziPlugin` to `application` module to provide these additional artifacts. I understand it is weird user experience to do that approach. I'm open to hearing other options.

kevinzheng-ap:
Close this one. Will open a new one when it is fixed properly in AGP. More in https://github.com/cashapp/paparazzi/issues/1060#issuecomment-2154361319

PR #1373, opened 2024-04-09
Revert changes made from older apis

geoff-powell:
Reverts changes made for older apis in #1351


geoff-powell:
> Looks good to me. Were the changes in the `PaparazziViewOwners` class necessitated by the version bumps?

yep, we were using an older version of androidX activity dependency that required a different method to override.

jrodbx (approved):

PR #1377, opened 2024-04-10
Clean up and refactor some things prior to gif support

jrodbx:
https://github.com/cashapp/paparazzi/pull/1186 has a few cleanup items embedded in it; pulling that out to make git history a little cleaner.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
@@ -60,7 +60,8 @@ public class SnapshotVerifier @JvmOverloads constructor(
           relativePath = expected.path,
           image = image,
           goldenImage = goldenImage,
-          maxPercentDifferent = maxPercentDifference
+          maxPercentDifferent = maxPercentDifference,
+          failureDir = failureDir
         )
       }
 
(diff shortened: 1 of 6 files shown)


paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
```diff
@@ -69,4 +70,15 @@ public class SnapshotVerifier @JvmOverloads constructor(
   }
 
   override fun close(): Unit = Unit
+
+  private companion object {
+    /** Directory where to write the thumbnails and deltas. */
+    private val failureDir: File
+      get() {
+        val buildDirString = System.getProperty("paparazzi.build.dir")
+        val failureDir = File(buildDirString, "paparazzi/failures")
```
gabrielittner, line 79: Since you're generally trying to keep Gradle and the runtime separate so that it could be used with other build systems, would it make sense to introduce a `paparazzi.failure.dir` property that directly points to `build/paparazzi/failures`? `paparazzi.build.dir` is only used here and for reading `appTestDir` but `appTestDir` is unused.

jrodbx: Few things to unpack here:
1) This is using Java system properties, not Gradle properties, so other build systems should still work here?
2) Good catch on unused appTestDir, that should be removed!
3) No strong opinion on introducing `paparazzi.failure.dir`; however, this would allow implementers to point to a failure directory outside of build, whereas the current approach forces it to live under `build.  Would that be desirable?  It used to live outside and I think the general feedback was that it wasn't helpful living alongside source-controlled sources.

gamepro65 (approved):

PR #1380, opened 2024-04-15
Fix predraw callbacks on every frame

gamepro65:
Noticed onPreDraw stopped getting callbacks for subsequent frames in an animation. Previously RenderSessionImpl manually called `AttachInfo_Accessor.dispatchOnPreDraw(viewRoot);` however that logic has been removed in favor of [ThrededRenderer](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ThreadedRenderer.java;l=742?q=ThreadedRenderer&ss=android%2Fplatform%2Fsuperproject%2Fmain) handling it.

Poking around AS I noticed their subsequent onPreDraw callbacks get invoked because they go down a different render code path using [FLAG_KEY_DISABLE_BITMAP_CACHING](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/layoutlib/bridge/src/com/android/layoutlib/bridge/impl/RenderSessionImpl.java;l=518?q=RenderSessionImpl)

Looking at [RenderTask](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/RenderTask.java;l=676?q=FLAG_KEY_DISABLE_BITMAP_CACHING&ss=android-studio) we can see they set it.

### DIFF
--- a/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
@@ -218,6 +218,29 @@ class PaparazziTest {
     assertThat(thrown).isTrue()
   }
 
+  @Test
+  fun preDrawOnEveryFrame() {
+    val log = mutableListOf<String>()
+
+    val view = object : View(paparazzi.context) {
+      override fun onAttachedToWindow() {
+        viewTreeObserver.addOnPreDrawListener {
+          log += "predraw"
+          true
+        }
+      }
+
+      override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+        log += "draw"
+      }
+    }
+
+    paparazzi.gif(view, fps = 4)
+
+    assertThat(log).isEqualTo(listOf("draw", "predraw", "predraw", "predraw"))
+  }
+
   private val time: Long
     get() {
       return TimeUnit.NANOSECONDS.toMillis(System_Delegate.nanoTime())
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
@@ -74,6 +74,7 @@ internal class Renderer(
       )
     )
       .plusFlag(RenderParamsFlags.FLAG_DO_NOT_RENDER_ON_CREATE, true)
+      .plusFlag(RenderParamsFlags.FLAG_KEY_DISABLE_BITMAP_CACHING, true)
       .withTheme("AppTheme", true)
 
     val platformDataRoot = System.getProperty("paparazzi.platform.data.root")


paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -218,6 +218,27 @@ class PaparazziTest {
     assertThat(thrown).isTrue()
   }
 
+  @Test
+  fun preDrawOnEveryFrame() {
+    val order = mutableListOf<String>()
+    val view = object : View(paparazzi.context) {
```
jrodbx, line 224:
```suggestion
  fun predrawOnEveryFrame() {
    val log = mutableListOf<String>()
    
    val view = object : View(paparazzi.context) {
```

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -218,6 +218,27 @@ class PaparazziTest {
...
+
+      override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+        order.add("draw")
+      }
+    }
+
+    paparazzi.gif(view, fps = 4)
+
+    assertThat(order).isEqualTo(listOf("draw", "predraw", "predraw", "predraw"))
```
jrodbx, line 239: > Noticed onPreDraw stopped getting callbacks for subsequent frames in an animation. 

I might be confused, so I'll ask a clarifying question.

This unit test isn't animating the view, so wouldn't pre-caching the bitmap and avoiding unnecessary predraws make sense?

Put another way, what if this view were animating?  Would it then show the predraws without the RenderParamsFlag change?

Also, why isn't there a "predraw" prior to the 1st "draw"?

gamepro65: The preDrawListener is added to AttachInfo's ViewTreeObserver which is supposed to be invoked for every frame, it isn't a drawing call per-se rather a hey, we are about to render to the screen. 

Yes we are not animating the view which is why we don't expect subsequent draw's but the preDraw still should apply before each frame.

Because of the changes, [this](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ThreadedRenderer.java;l=700?q=ThreadedRenderer&ss=android%2Fplatform%2Fsuperproject%2Fmain) call causes draw to exectue, lower you can see [here](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ThreadedRenderer.java;l=742?q=ThreadedRenderer&ss=android%2Fplatform%2Fsuperproject%2Fmain) is where the preDraw is called before rendering the root's RenderingNode into the canvas. If preDraw causes an invalidation of the view, then [this](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/view/ThreadedRenderer.java;l=745?q=ThreadedRenderer&ss=android%2Fplatform%2Fsuperproject%2Fmain) line will update the dirty list and render the final changes within the first frame.

If we updated the test to invalidate in preDraw we see the end results being

[draw, predraw, draw, predraw, draw, predraw, draw]

Notice 3 predraws but 4 draws. This is because invalidating the first frame in preDraw causes a 2nd draw phase before the first draw pass renders. Can update the test to invalidate to showcase that behavior. The pre iguana LayoutLib however it was as you'd be expecting in order `["predraw", "draw", "predraw", "predraw"]`

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -218,6 +218,29 @@ class PaparazziTest {
...
 
+  @Test
+  fun preDrawOnEveryFrame() {
+    val log = mutableListOf<String>()
+
+    val view = object : View(paparazzi.context) {
+      override fun onAttachedToWindow() {
+        viewTreeObserver.addOnPreDrawListener {
+          invalidate()
+          log.add("predraw")
```
jrodbx, line 229:
```suggestion
          log += "predraw"
```

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -218,6 +218,29 @@ class PaparazziTest {
...
+        }
+      }
+
+      override fun onDraw(canvas: Canvas?) {
+        super.onDraw(canvas)
+        log.add("draw")
```
jrodbx, line 235:
```suggestion
        log += "draw"
```

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -218,6 +218,29 @@ class PaparazziTest {
...
   }
 
+  @Test
+  fun preDrawOnEveryFrame() {
+    val log = mutableListOf<String>()
+
+    val view = object : View(paparazzi.context) {
+      override fun onAttachedToWindow() {
+        viewTreeObserver.addOnPreDrawListener {
+          invalidate()
```
jrodbx, line 228:
```suggestion
```

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
jrodbx: > Yes we are not animating the view which is why we don't expect subsequent draw's but the preDraw still should apply before each frame.

Ah, got it.  TIL.

Btw, no need to add the `invalidate` to force the extra "draw"s, your explanation makes sense.

jrodbx (approved):

PR #1382, opened 2024-04-16
Add boolean Flag to decide if image should be original size or not

emuguy1:
Inspired by different other comments, Issues and another Pull Request by @sdoward #543, I tried to add a more simpler solution to the problem I and other developers are facing.
As I want to use the created Screenshots to also upload it to the PlayStore and need full Res Images for that the current limit is blocking me from using the images.

Therefore I added an optional boolean flag 'useDeviceResolution', that when set to true, lets the resulting image be the deviceConfigs scale.

In the default case the image still gets downscaled to decrease file size.

I would really appreciate a feedback and if possible a fast inclution in the project, as I really want/need this feature.


swankjesse (approved):

emuguy1:
@swankjesse  Is there any update on when this can be included in an release? What would be the process to include something like this. Is there something I can help to get this issue resolved?

mehrmann:
We're facing the same issue, would want this to be merged so we can choose to have high resolution screenshots.

jrodbx:
@emuguy1  I'm ambivalent about adding this as it will likely be rewritten and break in the future; that being said, it seems like a simple enough thing to include now.  Mind adding a test that varies the boolean value using TestParameterInjector (see `paparazzi-gradle-plugin/src/test/projects/locale-qualifier` for an example)?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1608,6 +1608,31 @@ class PaparazziPluginTest {
     assertThat(nexus7SnapshotImage).isSimilarTo(nexus7GoldenImage).withDefaultThreshold()
   }
 
+  @Test
+  fun deviceResolutionQualifier() {
```
jrodbx, line 1612:
```suggestion
  fun scaledVersusFullDeviceResolution() {
```

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1608,6 +1608,31 @@ class PaparazziPluginTest {
...
+    val fixtureRoot = File("src/test/projects/device-resolution-qualifier")
```
jrodbx, line 1613:
```suggestion
    val fixtureRoot = File("src/test/projects/device-resolution")
```

paparazzi-gradle-plugin/src/test/projects/device-resolution-qualifier/build.gradle
jrodbx, line 1: rename test project to `device-resolution` since this isn't using qualifiers (like locales do)

paparazzi-gradle-plugin/src/test/projects/device-resolution-qualifier/src/test/java/app/cash/paparazzi/plugin/test/LocaleQualifierTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import app.cash.paparazzi.DeviceConfig
+import app.cash.paparazzi.Paparazzi
+import com.google.testing.junit.testparameterinjector.TestParameter
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+class LocaleQualifierTest(
```
jrodbx, line 12:
```suggestion
class DeviceResolutionTest(
```

paparazzi-gradle-plugin/src/test/projects/device-resolution-qualifier/src/test/java/app/cash/paparazzi/plugin/test/LocaleQualifierTest.kt
```diff
@@ -0,0 +1,30 @@
...
+  @TestParameter useDeviceResolutionFlag: OriginalDeviceResolutionFlag
```
jrodbx, line 13:
```suggestion
  @TestParameter useDeviceResolution: Boolean
```

sample/src/test/java/app/cash/paparazzi/sample/TestParameterInjectorTest.kt
```diff
@@ -34,7 +34,7 @@ class TestParameterInjectorTest(
   }
 
   @get:Rule
-  val paparazzi = Paparazzi(deviceConfig = config.deviceConfig)
+  val paparazzi = Paparazzi(deviceConfig = config.deviceConfig, useDeviceResolution = true)
```
jrodbx, line 37: revert?

paparazzi-gradle-plugin/src/test/projects/device-resolution-qualifier/src/test/resources/resolution_default.png
jrodbx, line 1: nit: would be nice to use another view layout such that the resolution difference is more apparent, but not a blocker to merge

jrodbx, line 1: hmm, does this file need to be re-recorded due to the current test failure on CI?

emuguy1: Is there a easy way to do this? Currently unable to run the test environment locally.

emuguy1: Can look into that. Just would like to first get it running and then improve. But thats actually a good suggestion

emuguy1: So I again recreated them, but they dont change so is it possible that the order is now switched? Will try this

emuguy1: Ok so this wasnt the solution. Is there a way, to see the pictures it generates. I locally created the exact same test in samples and let it run but it created the exact same pictures so it seems there is anything interferring with the test. Is there a way to see the pictures, that the test-run creates.

emuguy1:
@swankjesse @jrodbx can maybe one of you take a look at this. It fails with a gradle build error saying it cant find the guava-fix.gradle

```
Build file '/Users/runner/work/paparazzi/paparazzi/paparazzi-gradle-plugin/src/test/projects/device-resolution/build.gradle' line: 28

* What went wrong:
A problem occurred evaluating root project 'device-resolution'.
> Could not read script '/Users/runner/work/paparazzi/paparazzi/paparazzi-gradle-plugin/src/test/projects/guava-fix.gradle' as it does not exist.
```

I did find the error with the image difference. But this one I cant work out. I definitely see the file, the build gradle is the same as in the other projects and there it works.

Thanks for the help in advance.

TWiStErRob:
Update from main and remove the apply:
https://github.com/cashapp/paparazzi/commit/d20664920ee8b46c1741cecb5f3804b7d9ff1d73

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -381,8 +382,8 @@ public class PaparazziSdk @JvmOverloads constructor(
 
   private fun scaleImage(image: BufferedImage): BufferedImage {
     val scale = ImageUtils.getThumbnailScale(image)
-    // Only scale images down so we don't waste storage space enlarging smaller layouts.
-    return if (scale < 1f) ImageUtils.scale(image, scale, scale) else image
+    // Only scale images down, so we don't waste storage space enlarging smaller layouts.
+    return if (scale < 1f && !useDeviceResolution) ImageUtils.scale(image, scale, scale) else image
   }
```
Lutzifer, line 387: @emuguy1 maybe switch the if and else here to avoid the negation

jrodbx (approved):

emuguy1:
it works!!!!!! nice thanks a lot.

PR #1384, opened 2024-04-17
Fix negative SystemClock.uptimeMillis()

gamepro65:
Uptime is calculated based on bootTime which if not set defaults to the System's bootTime which combined with our overriding nanoTime causes SystemClock.uptimeMillis() to return a negative number.


jrodbx (approved):
nice!

PR #1388, opened 2024-04-22
Fix issue where multiple snapshot calls when using render extension fails as the view never is removed from its parent

geoff-powell:
Calling `Paparazzi.snapshot` multiple times on a view would cause an exception. 

```
java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
```

This fixes the issue by removing the correct view from its parent after a snapshot is taken.


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -246,8 +246,8 @@ public class PaparazziSdk @JvmOverloads constructor(
     frameCount: Int
   ) {
     val viewGroup = bridgeRenderSession.rootViews[0].viewObject as ViewGroup
-    val modifiedView = renderExtensions.fold(view) { view, renderExtension ->
-      renderExtension.renderView(view)
+    val modifiedView = renderExtensions.fold(view) { currentView, renderExtension ->
+      renderExtension.renderView(currentView)
```
geoff-powell, line 250: readability

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -303,7 +303,8 @@ public class PaparazziSdk @JvmOverloads constructor(
         }
       }
     } finally {
-      viewGroup.removeView(modifiedView)
+      // Remove original view from parent, even if there aren't render extensions applied
+      (view.parent as ViewGroup).removeView(view)
```
geoff-powell, line 307: nit: in this case any views created by a render extension will not be removed. Is this something we should case about?

JakeWharton: Yeah I think you need to do both still. Remove the outermost view from the view group, and remove the user view from it's parent. These might be the same so you could make it conditional on `view !== modifierView`.

JakeWharton (approved):

PR #1407, opened 2024-05-02
Update snapshots of tests affected by Compose 1.6 font padding change

gabrielittner:
Fixes these 4 tests 
```
PaparazziPluginTest.compose
PaparazziPluginTest.composeA11y
PaparazziPluginTest.composeWear
PaparazziPluginTest.nightModeCompose
```

When I run these locally or through Github Actions on our fork ([example 1](https://github.com/freeletics/paparazzi/actions/runs/8899464793), [example 2](https://github.com/freeletics/paparazzi/actions/runs/8929710663)) they are failing. Compose 1.6 changed the default value for `includeFontPadding` which is visible in the re-generated screenshots. It looks like there is a build cache issue that's causing them not to fail in this repo.


jrodbx (approved):
thanks!

PR #1419, opened 2024-05-09
Simplify api for asserting generated images to golden images for plugin

geoff-powell:
This should help when adding test projects with multiple images that are generated.

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -695,8 +695,8 @@ class PaparazziPluginTest {
       .runFixture(fixtureRoot) { build() }
 
     val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/debug/images")
-    val snapshots = snapshotsDir.listFiles().apply { sortBy { it.lastModified() } }
-    assertThat(snapshots!!).hasLength(4)
+    val snapshots = snapshotsDir.listFilesSorted()
+    assertThat(snapshots!!).hasSize(4)
 
     val normal = File(fixtureRoot, "src/test/resources/normal.png")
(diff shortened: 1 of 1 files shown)


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -777,8 +777,8 @@ class PaparazziPluginTest {
       .runFixture(fixtureRoot) { build() }
 
     val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/debug/images")
-    val snapshots = snapshotsDir.listFiles()
-    assertThat(snapshots!!).hasLength(3)
+    val snapshots = snapshotsDir.listFilesSorted()
```
jrodbx, line 780: nit: this adds an unnecessary array copy now.  I don't see the value in adding this everywhere, since it still needs to be a thing to remember when adding a new test.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1633,7 +1633,7 @@ class PaparazziPluginTest {
       .runFixture(fixtureRoot) { build() }
 
     val snapshotsDir = File(fixtureRoot, "build/reports/paparazzi/debug/images")
-    val snapshots = snapshotsDir.listFiles()?.sortedBy { it.lastModified() }
+    val snapshots = snapshotsDir.listFilesSorted()
```
jrodbx, line 1636: fine for the regular cases

jrodbx (approved):

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
geoff-powell: sure happy to only update the ones that have multiple images to compare instead.

PR #1423, opened 2024-05-13
Delete MatrixMultiplication Interceptor

gamepro65:
Regression caught in Cash, looks like Matrix is now provided so now we are unintentionally were messing with these operations 😅


jrodbx (approved):
Yay!  Any tests to delete?

gamepro65:
> Yay! Any tests to delete?

I didn't see any, looked at when we added these interceptors but no tests in that PR

TWiStErRob:
@gamepro65 how did this manifest?

gamepro65:
> @gamepro65 how did this manifest?

We've got a few tests in Cash which use `multiplyMV` and `multiplyMM`. Running those tests resulted in these operations applying twice. Could make a unit test to catch regressions if you were looking for a more concrete example.

TWiStErRob:
Nice. Up to you on the regression, I guess this was triggered by a layoutlib bump? If so, do we see a chance of changing again?

Issue #1376, opened 2024-04-10
1.3.4-SNAPSHOT causes configuration cache error when used with Google Services


joharei:
**Description**
This issue is very similar to #1369, but it is now failing because of Google Services. After updating to 1.3.4-20240409.090340, I am getting an error that seems like it's related to Gradle's configuration cache and a clash with the Google Services plugin:
```
Configuration cache state could not be cached: field `__projectResourceDirs__` of task `:app:preparePaparazziDebugResources` of type `app.cash.paparazzi.gradle.PrepareResourcesTask`: error writing value of type 'org.gradle.api.internal.provider.DefaultListProperty'
> Querying the mapped value of flatmap(provider(task 'processDebugGoogleServices', class com.google.gms.googleservices.GoogleServicesTask)) before task ':app:processDebugGoogleServices' has completed is not supported
```

If I run the task without configuration cache it runs fine:
```bash
./gradlew --no-configuration-cache verifyPaparazziDebug
```

This error does not happen on 1.3.3.

**Steps to Reproduce**
1. Add the Google Services plugin to the project
2. Update to Paparazzi 1.3.4-20240409.090340 (SNAPSHOT)
3. Run `verifyPaparazzi`

**Expected behavior**
Screenshot tests should run.

**Additional information:**
- Paparazzi Version: 1.3.4-20240409.090340
- OS: macOS 14.4.1
- Compile SDK: 34
- Gradle Version: 8.6
- Android Gradle Plugin Version: 8.3.1

PR #1424, opened 2024-05-13
Fix configuration cache bug when generating sources

jrodbx:
Defers fetching of file collection elements during task configuration. Closes https://github.com/cashapp/paparazzi/issues/1376.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -156,18 +156,24 @@ public class PaparazziPlugin : Plugin<Project> {
         task.targetSdkVersion.set(android.targetSdkVersion())
         task.compileSdkVersion.set(android.compileSdkVersion())
         task.projectResourceDirs.set(
-          project.provider {
-            val resourcesComputer = variant.mergeResourcesProvider?.get()?.resourcesComputer
-            val generateResValuesDirs = resourcesComputer?.generatedResOutputDir ?: project.files()
-            val extraGeneratedResDirs = resourcesComputer?.extraGeneratedResFolders ?: project.files()
-
-            extraGeneratedResDirs.map(projectDirectory::relativize) + localResourceDirs.relativize(projectDirectory) + generateResValuesDirs.map(projectDirectory::relativize)
+          run {
+            val resourcesComputer = variant.mergeResourcesProvider
+            val extraGeneratedResDirs =
+              resourcesComputer?.map { it.resourcesComputer.extraGeneratedResFolders }
+                ?: project.provider { project.files() }
+            val generatedResOutputDirs =
+              resourcesComputer?.map { it.resourcesComputer.generatedResOutputDir }
+                ?: project.provider { project.files() }
+            extraGeneratedResDirs
+              .zip(project.provider { localResourceDirs }, FileCollection::plus)
+              .zip(generatedResOutputDirs, FileCollection::plus)
+              .flatMap { it.relativize(projectDirectory) }
           }
         )
-        task.moduleResourceDirs.set(project.provider { moduleResourceDirs.relativize(projectDirectory) })
-        task.aarExplodedDirs.set(project.provider { aarExplodedDirs.relativize(gradleHomeDir) })
-        task.projectAssetDirs.set(project.provider { localAssetDirs.plus(moduleAssetDirs).relativize(projectDirectory) })
-        task.aarAssetDirs.set(project.provider { aarAssetDirs.relativize(gradleHomeDir) })
+        task.moduleResourceDirs.set(moduleResourceDirs.relativize(projectDirectory))
+        task.aarExplodedDirs.set(aarExplodedDirs.relativize(gradleHomeDir))
+        task.projectAssetDirs.set(localAssetDirs.plus(moduleAssetDirs).relativize(projectDirectory))
+        task.aarAssetDirs.set(aarAssetDirs.relativize(gradleHomeDir))
         task.paparazziResources.set(buildDirectory.file("intermediates/paparazzi/${variant.name}/resources.json"))
       }
 
--- a/paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
+++ b/paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
@@ -1,0 +1,37 @@
+plugins {
+  id 'com.android.library'
+  id 'kotlin-android'
+  id 'app.cash.paparazzi'
+}
+
+android {
+  namespace 'app.cash.paparazzi.plugin.test'
+  compileSdk libs.versions.compileSdk.get() as int
+  defaultConfig {
+    minSdk libs.versions.minSdk.get() as int
+  }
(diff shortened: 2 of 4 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -157,17 +157,23 @@ public class PaparazziPlugin : Plugin<Project> {
...
+              resourcesComputer?.map { it.resourcesComputer.extraGeneratedResFolders }
+                ?: project.provider { project.files() }
+            val generatedResOutputDirs =
+              resourcesComputer?.map { it.resourcesComputer.generatedResOutputDir }
+                ?: project.provider { project.files() }
+            extraGeneratedResDirs
+              .zip(project.provider { localResourceDirs }) { t, u -> t.plus(u) }
+              .zip(generatedResOutputDirs) { t, u -> t.plus(u) }
+              .flatMap { it.relativize(projectDirectory) }
+          }.get()
```
TWiStErRob, line 171: This looks fishy, why create a `.provider { }` just to `.get()` it? Am I reading this wrong?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -157,17 +157,23 @@ public class PaparazziPlugin : Plugin<Project> {
...
+            val extraGeneratedResDirs =
+              resourcesComputer?.map { it.resourcesComputer.extraGeneratedResFolders }
+                ?: project.provider { project.files() }
+            val generatedResOutputDirs =
+              resourcesComputer?.map { it.resourcesComputer.generatedResOutputDir }
+                ?: project.provider { project.files() }
+            extraGeneratedResDirs
+              .zip(project.provider { localResourceDirs }) { t, u -> t.plus(u) }
+              .zip(generatedResOutputDirs) { t, u -> t.plus(u) }
+              .flatMap { it.relativize(projectDirectory) }
```
TWiStErRob, line 170: this might read better (if compiles)
```suggestion
            extraGeneratedResDirs
              .zip(project.provider { localResourceDirs }, FileCollection::plus)
              .zip(generatedResOutputDirs, FileCollection::plus)
              .flatMap { it.relativize(projectDirectory) }
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -157,17 +157,23 @@ public class PaparazziPlugin : Plugin<Project> {
...
-          }
+            val resourcesComputer = variant.mergeResourcesProvider
+            val extraGeneratedResDirs =
+              resourcesComputer?.map { it.resourcesComputer.extraGeneratedResFolders }
+                ?: project.provider { project.files() }
+            val generatedResOutputDirs =
+              resourcesComputer?.map { it.resourcesComputer.generatedResOutputDir }
+                ?: project.provider { project.files() }
+            extraGeneratedResDirs
+              .zip(project.provider { localResourceDirs }) { t, u -> t.plus(u) }
```
TWiStErRob, line 168: This `provider { }` might be highlighting that `localResourceDirs`/`localAssetDirs` is being too eagerly calculated too? What happens if someone adds a source folder after this plugin is applied? (e.g. in a `variants.all { }`)

paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+  kotlinOptions {
+    jvmTarget = libs.versions.javaTarget.get()
+  }
+}
+
+androidComponents {
+  onVariants(selector().all()) { variant ->
+    def sourceGen =
+      project.tasks.register("do${variant.name.capitalize()}SourceGen", SourceGenTask) {
+        it.outputDirectory.set(null)
```
TWiStErRob, line 26: Why do you explicitly clear it? AGP will overwrite it regardless of current value, right?

paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+  }
+  kotlinOptions {
+    jvmTarget = libs.versions.javaTarget.get()
+  }
+}
+
+androidComponents {
+  onVariants(selector().all()) { variant ->
+    def sourceGen =
+      project.tasks.register("do${variant.name.capitalize()}SourceGen", SourceGenTask) {
```
TWiStErRob, line 25: Should this have a `doFirst` that actually writes at least an empty source file? Just to prevent accidental NO-SOURCE at any point? e.g. the main source set has no code in this project.

TWiStErRob (commented):
Nice one, thanks for looking! A few thoughts, and questions.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Solely to keep this block of code together instead of inlining the respective parts above; it's going away soon anyway, so I took a shortcut

TWiStErRob: stdlib `run { }`?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: this is exactly what the added plugin test checks

paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
jrodbx: ah, leftover from earlier debugging.  removed.

paparazzi-gradle-plugin/src/test/projects/configuration-cache-generated-sources/build.gradle
jrodbx: what's the concern here?  that subsequent reruns will not be cached due to NO-SOURCE?

TWiStErRob: more like not having a close enough to real life test where a lot of code gets NOT executed due to not having no code. NO-SOURCE is an input check, the outputs will be just empty, so caching-wise it all the same.

jrodbx: i see, i try to favor minimal test setups.  the configuration cache bug wasn't surfaced by any actual source file gen, but rather the wiring of tasks/properties in SourceDirectoriesImpl.addGeneratedSourceDirectory, so imo, adding extra config, etc, to make this a real life test adds noise to anyone who reads the test and wants to understand the underlying issue better.

geoff-powell (approved):

PR #1427, opened 2024-05-14
Add @Paparazzi annotation api

geoff-powell:


gradle/aarAsJar.gradle
geoff-powell, line 1: extracted Aar2Jar logic for use in multiple modules

paparazzi-annotations/src/main/java/app/cash/paparazzi/annotations/PreviewModels.kt
```diff
@@ -0,0 +1,32 @@
+package app.cash.paparazzi.annotations
+
+import androidx.compose.runtime.Composable
+
+public sealed interface PaparazziPreviewData {
```
geoff-powell, line 5: states that exemplify previews

PR #1435, opened 2024-05-16
Extract aar2jar logic into shareable plugin

jrodbx:
Some minor things to clean up, but I think I like how this is looking.

### DIFF
--- a/build-logic/aar2jar/build.gradle
+++ b/build-logic/aar2jar/build.gradle
@@ -1,0 +1,21 @@
+apply plugin: 'org.jetbrains.kotlin.jvm'
+apply plugin: 'java-gradle-plugin'
+
+repositories {
+  mavenCentral()
+  google()
+}
+
+dependencies {
+  compileOnly gradleApi()
+  implementation libs.plugin.android
+}
+
+gradlePlugin {
+  plugins {
+    aar2jar {
+      id = 'aar2jar'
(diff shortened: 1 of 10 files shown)


geoff-powell (approved):

TWiStErRob:
@jrodbx Are you planning to publish this at one point?

jrodbx:
> @jrodbx Are you planning to publish this at one point?

Hadn't considered it, but could!  Would you find it useful?

TWiStErRob:
I would, if it solves the following (yet to try):

Recently I had to bump a module from kotlin-library to android-library:
https://github.com/TWiStErRob/net.twisterrob.astro/commit/14a0fbb314b1819c7a987c30d93e933b90fe888e#diff-e560ab5b481520f07c70edf4be9724d3b19aada99c2308249bdbefc047993c32R2

because a dependency (`:test-base-compose`) is an `android-library`:
https://github.com/TWiStErRob/net.twisterrob.astro/commit/14a0fbb314b1819c7a987c30d93e933b90fe888e#diff-ce5319610e382c1f6f53ed10bee8d7f6e19aa67e2771792e34707b4e9029f779

because it depends on Espresso Core, which is an AAR:
https://mvnrepository.com/artifact/androidx.test.espresso/espresso-core/3.5.1

As far as I understand if I apply this plugin on the `:test-base-compose`, I can keep it a `kotlin-library` and therefore have less AGP modules, especially low-level ones that are referenced everywhere.

PR #1437, opened 2024-05-18
Remove JUnit usages from files in main source set

jrodbx:
The `paparazzi` module has been designed as a JUnit rule, so it's been tempting to use Junit assertions in both test sources as well as main sources.  However, moving core classes out into an SDK non-JUnit module reveals the tight coupling in some classes, so let's fix that.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
@@ -16,9 +16,6 @@
 
 package app.cash.paparazzi.internal
 
-import org.junit.Assert.assertEquals
-import org.junit.Assert.assertTrue
-import org.junit.Assert.fail
 import java.awt.AlphaComposite
 import java.awt.Color
 import java.awt.Graphics2D
@@ -68,7 +65,7 @@ internal object ImageUtils {
       var message = "Unable to load golden thumbnail: $relativePath\n"
       message = saveImageAndAppendMessage(thumbnail, message, relativePath, failureDir)
       if (FAIL_ON_MISSING_THUMBNAIL) {
-        fail(message)
+        throw IllegalStateException(message)
       } else {
         println(message)
       }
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
```diff
@@ -68,7 +65,7 @@ internal object ImageUtils {
       var message = "Unable to load golden thumbnail: $relativePath\n"
       message = saveImageAndAppendMessage(thumbnail, message, relativePath, failureDir)
       if (FAIL_ON_MISSING_THUMBNAIL) {
-        fail(message)
+        throw AssertionError(message)
```
TWiStErRob, line 68: Have you considered using `AssertionFailedError` from https://github.com/ota4j-team/opentest4j ?

The reasoning would be that this is a testing framework, and is running on JVM so it fits the umbrella open test is trying to standardize.

gamepro65 (approved):
Makes sense

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
JakeWharton: Seems like there should be a failure policy interface that callers of the SDK can supply with a default instance implementation that throws AE. I would not take a hard dependency on that project.

JakeWharton (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
TWiStErRob: Is this another ISE, btw? as the test is broken, not failed.

JakeWharton: I don't know. I thought this case was when you add a snapshot call and run verify without having done a record.

jrodbx: Correct, per the other reasonings, I'd say this should also be an ISE.

Issue #1440, opened 2024-05-20
Incorrect contentDescriptions when using AccessibilityRenderExtension with clickable nodes


shorner5:
**Description**
When a node is clickable, and its child uses `clearAndSetSemantics`, semantics are not cleared. When the parent is NOT clickable, the semantics are cleared correctly. 

**Steps to Reproduce**
Add a test where the parent is clickable, and a child clears its children's semantics:

```
class ScreenshotTest {
    @get:Rule
    val paparazzi = Paparazzi(
        renderExtensions = setOf(AccessibilityRenderExtension())
    )

    @Test
    fun contentDescriptionTest() {
        paparazzi.snapshot {
            Box(modifier = Modifier.clickable {}) {
                Column(
                    modifier = Modifier.clearAndSetSemantics {
                        contentDescription = "OVERRIDDEN CONTENT DESCRIPTION"
                    }
                ) {
                    Text(text = "Text")
                }
            }
        }
    }
}
```

**Expected behavior**
The content description should be "OVERRIDDEN CONTENT DESCRIPTION":
![expected](https://github.com/cashapp/paparazzi/assets/15132447/8b8a6de6-ab49-4a36-a107-558d97f83c25)

**Actual result**
Paparazzi renders the content description as "Text, OVERRIDDEN CONTENT DESCRIPTION". 
![actual](https://github.com/cashapp/paparazzi/assets/15132447/b0aeff6b-1694-4f73-a09f-e00e96305f3d)

TalkBack announces only "OVERRIDDEN CONTENT DESCRIPTION", and Layout Inspector shows that the Box's contentDescription is correct. 
<img width="1048" alt="Screenshot 2024-05-20 at 1 35 37 PM" src="https://github.com/cashapp/paparazzi/assets/15132447/f1fefe00-7e7b-47a0-8381-4352e0bc6087">

**Additional information:**
- Paparazzi Version: 1.3.3
- OS: Mac
- Compile SDK: 34
- Gradle Version: 8.6
- Android Gradle Plugin Version: 8.4.0-beta01

PR #1445, opened 2024-05-23
Fix when clearAndSetSemantics is used to render content desc

geoff-powell:
Fixes https://github.com/cashapp/paparazzi/issues/1440

Explanation of why semantics aren't cleared in unmergedNodes. https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/node/SemanticsModifierNode.kt;l=38-50?q=isClearingSemantics


jrodbx (approved):

BrianGardnerAtl (approved):

geoff-powell:
Confirmed with other permutations nested at different layers that this displays exactly like talkback reads.

PR #1454, opened 2024-05-28
Update Kotlin and Compose to v2

jrodbx:
Closes #1441

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,11 +2,10 @@
 agp = "8.4.1"
 androidTools = "31.4.1" # == 23.0.0 + agp version
 bytebuddy = "1.14.16"
-composeCompiler = "1.5.14"
 coroutines = "1.8.1"
 javaTarget = "11"
-kotlin = "1.9.24"
-ksp = "1.9.24-1.0.20"
+kotlin = "2.0.0"
+ksp = "2.0.0-1.0.21"
 ktlint = "1.2.1"
 moshi = "1.15.1"
 
@@ -27,7 +26,6 @@ androidx-recyclerview = { module = "androidx.recyclerview:recyclerview", version
 bytebuddy-agent = { module = "net.bytebuddy:byte-buddy-agent", version.ref = "bytebuddy" }
 bytebuddy-core = { module = "net.bytebuddy:byte-buddy", version.ref = "bytebuddy" }
 
-compose-compiler = { module = "androidx.compose.compiler:compiler", version.ref = "composeCompiler" }
 composeUi-material = { module = "androidx.compose.material:material", version = "1.6.7" }
 composeUi-uiTooling = { module = "androidx.compose.ui:ui-tooling" }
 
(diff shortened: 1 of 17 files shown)


paparazzi/build.gradle
```diff
@@ -45,7 +45,7 @@ dependencies {
   api libs.junit
   api libs.androidx.annotations
   api libs.guava
-  api libs.kotlinx.coroutines.core
+  api libs.kotlinx.coroutines.android
```
jrodbx, line 48: it appears that there used to be a transitive dependency on coroutines.android and this bump caused this reference to [Handler.asCoroutineDispatcher](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt#L588) to fail.

gradle/libs.versions.toml
```diff
@@ -73,8 +74,9 @@ plugin-android = { module = "com.android.tools.build:gradle", version.ref = "agp
 plugin-buildConfig = { module = "com.github.gmazzo.buildconfig:plugin", version = "5.3.5" }
 plugin-dokka = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version = "1.9.20" }
 plugin-kotlin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
+plugin-kotlin-compose = { module = "org.jetbrains.kotlin:compose-compiler-gradle-plugin", version.ref = "kotlin" }
 plugin-kotlinApiDump = { module = "org.jetbrains.kotlinx:binary-compatibility-validator", version = '0.14.0' }
 plugin-ksp = { module = "com.google.devtools.ksp:com.google.devtools.ksp.gradle.plugin", version.ref = "ksp" }
 plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version = "0.28.0" }
-plugin-poko = { module = "dev.drewhamilton.poko:poko-gradle-plugin", version = "0.15.3" }
+plugin-poko = { module = "dev.drewhamilton.poko:poko-gradle-plugin", version = "0.16.0-beta01" }
```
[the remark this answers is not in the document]
JakeWharton: https://github.com/drewhamilton/Poko/issues/352

gradle/libs.versions.toml
```diff
@@ -27,14 +27,15 @@ androidx-recyclerview = { module = "androidx.recyclerview:recyclerview", version
 bytebuddy-agent = { module = "net.bytebuddy:byte-buddy-agent", version.ref = "bytebuddy" }
 bytebuddy-core = { module = "net.bytebuddy:byte-buddy", version.ref = "bytebuddy" }
 
-compose-compiler = { module = "androidx.compose.compiler:compiler", version.ref = "composeCompiler" }
+compose-compiler = { module = "org.jetbrains.kotlin:kotlin-compose-compiler-plugin-embeddable", version.ref = "kotlin" }
```
JakeWharton, line 30: This should not be needed anymore. Apply the Compose plugin instead.

jrodbx: I tried that, but paparazzi is a jvm module, so this line was failing when applying the Compose plugin:

```
add(AbstractKotlinPluginKt.PLUGIN_CLASSPATH_CONFIGURATION_NAME, libs.compose.compiler)
```

JakeWharton (approved):

gradle/libs.versions.toml
JakeWharton: Delete that line completely. The plugin will work fine in JVM modules.

jrodbx: Ah, you're right.  I guess my issue here was related to the plugin test fixture configuration setup that I later fixed.  Good catch!

PR #1456, opened 2024-05-29
Post-Jellybean cleanup

jrodbx:


renovate.json5
```diff
@@ -33,8 +32,7 @@
     {
       "description": "LayoutLib shouldn't auto-merge, because updating it is a complex process, see libs/README.md.",
```
TWiStErRob, line 33: Does this read me need updating? or the description here?

kevinzheng-ap (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
```diff
@@ -79,15 +79,15 @@ internal class Renderer(
...
-    val platformDataDir = File(platformDataRoot, "data")
+    val layoutlibRuntimeRoot = System.getProperty("paparazzi.layoutlib.runtime.root")
+      ?: throw RuntimeException("Missing system property for 'paparazzi.layoutlib.runtime.root'")
+    val buildProp = File(layoutlibRuntimeRoot, "build.prop")
+    val platformDataDir = File(layoutlibRuntimeRoot, "data")
     val fontLocation = File(platformDataDir, "fonts")
-    val nativeLibLocation = File(platformDataDir, getNativeLibDir())
     val icuLocation = File(platformDataDir, "icu" + File.separator + "icudt72l.dat")
     val keyboardLocation = File(platformDataDir, "keyboards" + File.separator + "Generic.kcm")
-    val buildProp = File(environment.platformDir, "build.prop")
```
gabrielittner, line 89: Nice! I missed that the build.prop is shipped inside the runtime jar. With this gone you can completely remove `platformDir` from `Environment` which means that Paparazzi doesn't need to care at all about the local SDK anymore

geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/Renderer.kt
jrodbx: correct, that's the next PR

jrodbx: probably also means that the `build-class-next-sdk` test is no longer needed since there isn't any way to point to experimental sdks anymore.

renovate.json5
jrodbx: done

PR #1462, opened 2024-05-29
Pull out SDK into its own module

jrodbx:
Note that this does not rename the artifacts as that would be a breaking change; this should remain binary compatible until 2.0


paparazzi-junit4/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("invisible_reference", "invisible_member")
```
jrodbx, line 1: allows "friend references", i.e., extending internal access across cross project-modules.  By default, kotlin only allows code within the same module to access `internal` members.

paparazzi-sdk/build.gradle
```diff
@@ -42,7 +42,6 @@ dependencies {
   api libs.tools.ninepatch
   api libs.tools.sdkCommon
   api libs.kxml2
-  api libs.junit
```
jrodbx, line 45: 🎉

paparazzi-junit4/build.gradle
```diff
@@ -0,0 +1,129 @@
...
+  )
+  systemProperty(
+    "paparazzi.layoutlib.runtime.root",
+    configurations.layoutlibRuntime.singleFile.absolutePath
+  )
+  systemProperty(
+    "paparazzi.layoutlib.resources.root",
+    configurations.layoutlibResources.singleFile.absolutePath
+  )
+}
```
jrodbx, line 129: a bit unfortunate that both the sdk and junit rule needs this; but it makes sense, the actual config in practice would come from the (gradle) plugin.  might make sense to evolve the sdk into a builder pattern; i'll mess around with this in a follow-up.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -329,7 +329,7 @@ public class PaparazziPlugin : Plugin<Project> {
 
   private fun Project.addTestDependency() {
     val dependency = if (isInternal()) {
-      dependencies.project(mapOf("path" to ":paparazzi"))
+      dependencies.project(mapOf("path" to ":paparazzi-junit4"))
```
jrodbx, line 332: deliberate for now.  can query gradle extension config for junit4 vs junit5, but not sure how to keep it generic for other 3rd party wrappers.

paparazzi-junit4/build.gradle
```diff
@@ -0,0 +1,129 @@
...
+  registerTransform(org.gradle.api.internal.artifacts.transform.UnzipTransform) {
+    from.attribute(AndroidArtifacts.ARTIFACT_TYPE, ArtifactTypeDefinition.JAR_TYPE)
+    to.attribute(AndroidArtifacts.ARTIFACT_TYPE, ArtifactTypeDefinition.DIRECTORY_TYPE)
+  }
+}
+
+dependencies {
+  testImplementationAarAsJar libs.androidx.compose.ui.android
+  compileOnlyAarAsJar libs.androidx.compose.ui.android
+  compileOnlyAarAsJar libs.androidx.compose.runtime
```
jrodbx, line 33: calling out this switcheroo; replacing `compileOnlyAarAsJar libs.androidx.activity` with this: https://github.com/cashapp/paparazzi/pull/1462/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df

geoff-powell: nice, much better dependency

BrianGardnerAtl (approved):
Looks good! Looking forward to the sdk builder pattern

paparazzi-junit4/src/test/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtensionTest.kt
TWiStErRob: Not K2 friendly: https://youtrack.jetbrains.com/issue/KT-67920

Maybe it is possible to befriend another module's sourceSet with this?
https://github.com/TWiStErRob/repros/blob/main/kotlin/test-fixtures-and-suites-friends/build.gradle.kts

paparazzi-junit4/src/test/java/app/cash/paparazzi/RenderExtensionTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("invisible_reference", "invisible_member")
```
TWiStErRob, line 1: Consider using all uppercase to highlight that you're suppressing compiler errors

See https://github.com/JetBrains/kotlin/blob/master/compiler/frontend/src/org/jetbrains/kotlin/diagnostics/rendering/DefaultErrorMessages.java

```suggestion
@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
```

paparazzi-junit4/gradle.properties
TWiStErRob, line 1: Did you intentionally keep the `...:paparazzi` coordinate here?

Consider publishing the -junit4 from the -junit4 module and reintroduce a `:paparazzi` module that `api`s both the sdk and the junit4.

paparazzi-junit4/build.gradle
```diff
@@ -0,0 +1,129 @@
...
+    osLabel = 'linux'
+  }
+  layoutlibRuntime variantOf(libs.tools.layoutlib.runtime) { classifier(osLabel) }
+  layoutlibResources libs.tools.layoutlib.resources
+
+  testImplementation libs.truth
+}
+
+tasks.named("dokkaGfm").configure {
+  outputDirectory = rootProject.file("docs/1.x")
```
TWiStErRob, line 59: Amazing that you did all this while keeping 1.x 👏🏻

paparazzi-sdk/src/main/java/app/cash/paparazzi/internal/PaparazziLogger.kt
```diff
@@ -28,7 +28,7 @@ import java.util.logging.Logger.getLogger
  * This logger delegates to java.util.Logging.
  */
 internal class PaparazziLogger : ILayoutLog, ILogger {
```
TWiStErRob, line 30: rename class to match module/logger?

PR #1468, opened 2024-06-03
Make paparazzi aware or unit test dependencies

alyssoncs:
I'm having a problem on a project where I have a module that contains composables, those composables depend on a `IconProvider` interface, this interface outputs `@DrawableRes Int`. We do not want the UI module to depend on the implementation of this interface. But we do want this implementation available on unit tests.

We currently add a `testRuntimeOnly`  (could also be a `testImplementation`) on this interface implementation and get running espresso tests. But we get weird errors with Paparazzi. Looking into the code, it looks like the dependencies added on the test source set are not added in the `Config` class by `PaparazziPlugin`.

This PR is my attempt of making Paparazzi aware of resources of local modules and external libraries that are only visible to the test source set.

I have a [proof of concept](https://github.com/alyssoncs/paparazzi/tree/poc) of a composable depending on an interface, and this interface being provided only on the unit tests (it is also provided on the debug variant to make `@Preview`s work). Running `./gradlew :poc:ui:testReleaseUnitTest` should fail.

I also have the same [proof of concept + the code in this PR](https://github.com/alyssoncs/paparazzi/tree/use-test-resources-poc), and the same task succeeds.

This is my first time contributing to this project, let me know what you think of this. Also, does this deserve a new unit test? I could use some help on this. What would be important testing here?

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -23,13 +23,11 @@ import com.android.build.api.variant.ApplicationAndroidComponentsExtension
 import com.android.build.api.variant.DynamicFeatureAndroidComponentsExtension
 import com.android.build.api.variant.HasUnitTest
 import com.android.build.api.variant.LibraryAndroidComponentsExtension
-import com.android.build.api.variant.UnitTest
 import com.android.build.gradle.BaseExtension
 import com.android.build.gradle.internal.publishing.AndroidArtifacts.ArtifactType
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition
 import org.gradle.api.artifacts.type.ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE
@@ -117,10 +115,7 @@ public class PaparazziPlugin : Plugin<Project> {
 
       // https://android.googlesource.com/platform/tools/base/+/96015063acd3455a76cdf1cc71b23b0828c0907f/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/MergeResources.kt#875
 
-      val runtimeConfiguration: Configuration = variant
-        .nestedComponents
-        .filterIsInstance<UnitTest>()
-        .firstOrNull()?.runtimeConfiguration ?: variant.runtimeConfiguration
+      val runtimeConfiguration = testVariant.runtimeConfiguration
 
       val moduleResourceDirs = runtimeConfiguration
         .artifactsFor(ArtifactType.ANDROID_RES.type) { it is ProjectComponentIdentifier }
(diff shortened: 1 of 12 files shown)


paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-android/gradle.properties
```diff
@@ -1 +1,2 @@
 kotlin.mpp.androidSourceSetLayoutVersion1.nowarn=true
+android.useAndroidX=true
```
alyssoncs, line 2: The multiplatform test started to fail because `TestRuntimeClasspath` contained androidx dependencies, and `useAndroidX` was not set to true

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -70,12 +70,12 @@ public abstract class PrepareResourcesTask : DefaultTask() {
 
     val mainPackage = packageName.get()
     val resourcePackageNames = if (nonTransitiveRClassEnabled.get()) {
-      buildList {
+      buildSet {
```
alyssoncs, line 73: after the change, the `mainPackage` was also present on `artifactFiles`. Decided to remove duplicates.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -115,27 +117,32 @@ public class PaparazziPlugin : Plugin<Project> {
 
       // https://android.googlesource.com/platform/tools/base/+/96015063acd3455a76cdf1cc71b23b0828c0907f/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/MergeResources.kt#875
 
-      val moduleResourceDirs = variant.runtimeConfiguration
+      val runtimeConfiguration: Configuration = variant
+        .nestedComponents
+        .filterIsInstance<UnitTest>()
+        .firstOrNull()?.runtimeConfiguration ?: variant.runtimeConfiguration
```
TWiStErRob, line 123: see testVariant?

TWiStErRob:
> What would be important testing here?

The issue you're fixing, you can write a failing unit test (gradle project fixture) that only passes if the Paparazzi Plugin has the right wiring.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
alyssoncs: Missed that. Thanks!

alyssoncs:
> > What would be important testing here?
> 
> The issue you're fixing, you can write a failing unit test (gradle project fixture) that only passes if the Paparazzi Plugin has the right wiring.

Added a module that adds both a `testImplementation` and `testRuntimeOnly` on two other modules and tested that the resources were included on the generated `Config` class.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt
```diff
@@ -70,12 +70,12 @@ public abstract class PrepareResourcesTask : DefaultTask() {
...
         add(mainPackage)
         artifactFiles.files.forEach { file ->
           add(file.useLines { lines -> lines.first() })
         }
-      }
+      }.toList()
```
geoff-powell, line 78: nit: maybe this `resourcePackageNames` can just be a set instead of a list to remove the extra type transform.

alyssoncs: The only thing is that I would have to change `Config` to accept a `Set` instead of a `List`. And maybe for consistency it would require all the lists in `Config` to be sets as well. Let me know if you folks are ok with that.

I could also just use `distinct()` here and avoid the creation of a `Set`. Let me know what you think it's best.

TWiStErRob: I think it should be a list, order matters, or not? probably should be `buildList { }.distinct()` because `buildSet {}.toList()` (semantically) can have arbitrary order.

alyssoncs: Changed to `distinct`

geoff-powell (approved):

Issue #1340, opened 2024-03-14
Multiple paparazzi rules in the same test run give NullPointerException


eboudrant:
**Description**
We'd like to create our own rule to wrap some boiler plate code and help adopting some good practice by our users. One of it is to build snapshot test that cover font scale on device. Since the `fontScale` can only be set on the `DeviceConfig` I ended with a custom test rule that delegate to 2 `Paparazzi` rules (see [code](https://github.com/eboudrant/paparazzi/blob/multiple_paparazzi/sample/src/test/java/app/cash/paparazzi/sample/MyPaparazziRule.kt)). When the test run we get a NPE because of initialization code.

```
app.cash.paparazzi.sample.ScaledTextTest > compositeItems FAILED
    java.lang.NullPointerException
        at app.cash.paparazzi.Paparazzi.takeSnapshots(Paparazzi.kt:285)
        at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:216)
        at app.cash.paparazzi.Paparazzi.snapshot$default(Paparazzi.kt:215)
        at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:211)
        at app.cash.paparazzi.sample.MyPaparazziRule.snapshot(MyPaparazziRule.kt:39)
        at app.cash.paparazzi.sample.MyPaparazziRule.snapshot$default(MyPaparazziRule.kt:34)
        at app.cash.paparazzi.sample.ScaledTextTest.compositeItems(ScaledTextTest.kt:14)
```

`isInitialized` is `true` for the second `Paparazzi` rule giving it an invalid state.

Also a side question I guess would be, how would you recommend to approach testing on variants that are at the `DeviceConfig` level. Ultimately we want our user to write the test once and have multiple snapshot per `DeviceConfig` property we care about. 

**Steps to Reproduce**
This branch https://github.com/eboudrant/paparazzi/tree/multiple_paparazzi contains a new test `app.cash.paparazzi.sample.ScaledTextTest`

**Expected behavior**
Not crash, the test output would be 2 images

**Additional information:**
- Paparazzi Version: 1.3.3 (branch created from 1.3.3 tag)

eboudrant:
Btw, I just realized to solve the `fontScale` (or any attribute from the `DeviceFactor`) you can use `@TestParameter` at the test class level.

So paparazzi might just need a better error message than the `NullPointerException` noted here.

jrodbx:
[...]

PR #1473, opened 2024-06-06
Add error message when multiple instances of Paparazzi is being used

NaromKuoch:
Address https://github.com/cashapp/paparazzi/issues/1340

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -141,8 +141,7 @@ public class Paparazzi @JvmOverloads constructor(
     try {
       val snapshot = Snapshot(name, testName!!, Date())
       return snapshotHandler.newFrameHandler(snapshot, frameCount, fps)
-    } catch (e: NullPointerException) {
-      throw IllegalStateException("Missing testName. Multiple instances of Paparazzi possibly being used.")
+    } catch (e: NullPointerException) {a kdearjhgkjhb   throw IllegalStateException("Missing testName. Multiple instances of Paparazzi possibly being used.")
     }
   }
 
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -141,7 +141,8 @@ public class Paparazzi @JvmOverloads constructor(
     try {
       val snapshot = Snapshot(name, testName!!, Date())
       return snapshotHandler.newFrameHandler(snapshot, frameCount, fps)
-    } catch (e: NullPointerException) {a kdearjhgkjhb   throw IllegalStateException("Missing testName. Multiple instances of Paparazzi possibly being used.")
+    } catch (e: NullPointerException) {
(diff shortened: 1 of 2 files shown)


geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -138,8 +138,12 @@ public class Paparazzi @JvmOverloads constructor(
...
 
   private fun createFrameHandler(name: String? = null, frameCount: Int = 1, fps: Int = -1): SnapshotHandler.FrameHandler {
-    val snapshot = Snapshot(name, testName!!, Date())
-    return snapshotHandler.newFrameHandler(snapshot, frameCount, fps)
+    try {
+      val snapshot = Snapshot(name, testName!!, Date())
+      return snapshotHandler.newFrameHandler(snapshot, frameCount, fps)
+    } catch (e: NullPointerException) {
+      throw IllegalStateException("Missing testName. Multiple instances of Paparazzi possibly being used.")
+    }
```
geoff-powell, line 146: nit: instead of the try/catch can we do this instead?

```suggestion
    val testName = requireNotNull(testName) {
      "Missing testName. Multiple instances of Paparazzi possibly being used."
    }

     val snapshot = Snapshot(name, testName, Date())
     return snapshotHandler.newFrameHandler(snapshot, frameCount, fps)
```

sample/src/test/java/app/cash/paparazzi/sample/MultiplePaparazziTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class MultiplePaparazziTest {
+
+  @get:Rule
+  val paparazzi = MultiplePaparazziRule()
+
+  @Test(expected = IllegalStateException::class)
```
geoff-powell, line 17: nit: update expected exception with one thrown be `requireNotNull` from suggestion above.

jrodbx:
Thanks for the PR, @NaromKuoch!

However, this is more of a band-aid solution, vs tackling the root cause outlined in https://github.com/cashapp/paparazzi/issues/1340, namely, figuring out why testName is null and resolving that.  Otherwise, this can easily become a whack-a-mole scenario where another invalid state results in an NPE and we continue to update the error message. 😄 

I suspect our guard to avoid reinitializing the entire Android framework per every test method being a static global variable is the issue, so ideas on how to improve that would be the preferred route to go.

NaromKuoch:
@jrodbx @geoff-powell 
Thanks for the feedback!  I agree that my original solution is a band-aid for the case.  Here is my [alternative approach](https://github.com/cashapp/paparazzi/compare/master...NaromKuoch:paparazzi:narom/multiple-paparazzi-instance-error-message-2?expand=1).  I can open another PR if that's easier for you to take a look. I believe the alternative approach is better because it ensures only a single instance of Paparazzi can be applied in a test class.

>I suspect our guard to avoid reinitializing the entire Android framework per every test method being a static global variable is the issue, so ideas on how to improve that would be the preferred route to go.

I think that this is a nice guard to have otherwise I think it could impact the performance of overall test executions. My alternative solution doesn't align with your suggestion, so I'm ok with dropping it as well.

PR #1481, opened 2024-06-15
Publish Gradle build scans from CI

jrodbx:
To help investigate Gradle caching issues

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -5,8 +5,8 @@ plugins {
 develocity {
   buildScan {
     termsOfUseUrl = 'https://gradle.com/help/legal-terms-of-use'
-    termsOfUseAgree = 'yes'
     if (System.getenv('CI') == 'true') {
+      termsOfUseAgree = 'yes'
       tag 'CI'
     } else {
       publishing.onlyIf { false }
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,19 @@
+plugins {
+  id 'com.gradle.develocity' version '3.17.5'
+}
+
+develocity {
+  buildScan {
+    termsOfUseUrl = 'https://gradle.com/terms-of-service'
+    termsOfUseAgree = 'yes'
+    if (System.getenv('CI') == 'true') {
+      tag 'CI'
+    } else {
+      publishing.onlyIf { false }
+    }
+  }
+}
+
 rootProject.name = 'paparazzi-root'
 
 include ':paparazzi'
--- a/settings.gradle
+++ b/settings.gradle
@@ -4,7 +4,7 @@ plugins {
 
 develocity {
   buildScan {
-    termsOfUseUrl = 'https://gradle.com/terms-of-service'
+    termsOfUseUrl = 'https://gradle.com/help/legal-terms-of-use'
     termsOfUseAgree = 'yes'
     if (System.getenv('CI') == 'true') {
       tag 'CI'


settings.gradle
```diff
@@ -1,3 +1,19 @@
+plugins {
+  id 'com.gradle.develocity' version '3.17.5'
+}
+
+develocity {
+  buildScan {
+    termsOfUseUrl = 'https://gradle.com/terms-of-service'
+    termsOfUseAgree = 'yes'
```
TWiStErRob, line 8: Accept only on CI?

settings.gradle
```diff
@@ -1,3 +1,19 @@
...
+    if (System.getenv('CI') == 'true') {
+      tag 'CI'
+    } else {
+      publishing.onlyIf { false }
```
TWiStErRob, line 12: Doesn't this setup disable local ad-hoc `--scan` usage?

jrodbx: Haven't tried it, copied from cashapp/molecule.  The intent is for CI build scans only though, if that's the root question.

settings.gradle
jrodbx: Won't matter, due to publish config below

settings.gradle
TWiStErRob: > The intent is for CI build scans only though, if that's the root question.

Yep, I get what the goal is. I'm just highlighting that the current behaviour of `gradlew --scan` is going to be disabled locally. It's a surprise for any contributor.

If anyone wants to debug they'll have to first find this setting why it's not working and the move on to debugging.

jrodbx: > I'm just highlighting that the current behaviour of `gradlew --scan` is going to be disabled locally. It's a surprise for any contributor.
> 
> If anyone wants to debug they'll have to first find this setting why it's not working and the move on to debugging.

Out of curiosity, have you verified this?  Because when I run `gw paparazzi:test --scan`, the output included a link to [this build scan](https://scans.gradle.com/s/2jyggefft4d6s), so I think your concerns are covered.

TWiStErRob: That's because you have CI set, try it without.

jrodbx:
```
$ git checkout jrod/2024-06-07/develocity
$ unset CI
$ echo $CI

$ gw paparazzi:test --scan
...
Publishing build scan...
https://gradle.com/s/bd32kvhffkbbq
```

TWiStErRob: Ok, that's my bad then it seems `--scan` is the winner whatever is in the build script:
https://docs.gradle.com/develocity/gradle-plugin/current/#using_build_scans

TWiStErRob: However https://github.com/cashapp/paparazzi/pull/1481#discussion_r1641305568 does matter now, because `--scan` will auto-publish even locally without asking for an acceptance. Committing this outside of the CI block means you accept the terms on behalf of all the contributors.

See the warning at https://docs.gradle.com/develocity/gradle-plugin/current/#connecting_to_scans_gradle_com

settings.gradle
```diff
@@ -1,3 +1,19 @@
+plugins {
+  id 'com.gradle.develocity' version '3.17.5'
+}
+
+develocity {
+  buildScan {
+    termsOfUseUrl = 'https://gradle.com/terms-of-service'
```
TWiStErRob, line 7:
```suggestion
    termsOfUseUrl = 'https://gradle.com/help/legal-terms-of-use'
```

URL recently changed:
https://docs.gradle.com/develocity/gradle-plugin/current/#connecting_to_scans_gradle_com

settings.gradle
jrodbx: Ok, I've read the documented Gradle warning.  Thanks for providing that.

But I'm still confused...what is the root concern here?  If a project accepts the terms and a contributor subsequently passes the `--scan` option, wouldn't that imply that the terms would be accepted?

That being said, I'm not a lawyer and don't have a strong opinion either, so I've updated the PR to reflect your concern.

I would also be curious to get @JakeWharton's input here, since a quick search of Square and CashApp OSS repos show that they're similarly configured, which is why I went with that:

* https://github.com/search?q=org%3Asquare+termsOfUseAgree&type=code
* https://github.com/search?q=org%3Acashapp+termsOfUseAgree&type=code

TWiStErRob: > If a project accepts the terms and a contributor subsequently passes the --scan option, wouldn't that imply that the terms would be accepted?

Yes that's true, but as far as I understand "the project" is not really a legal entity, so you as the maintainer cannot accept the terms on behalf a contributor. It's them that use the services (of Gradle), when they publish their personal computer's/environment's data, not "the project". Not a lawyer either, this is just my understanding. Individuals can accept the terms globally in their init scripts on each machine.

JakeWharton: Someone did it once (purportedly incorrectly) and it was copy/pasted a million times. You were just doing your duty by copy/pasting it further.

JakeWharton (approved):

PR #1484, opened 2024-06-17
Extract Android sources logic to simplify plugin

jrodbx:


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -47,7 +47,11 @@ import java.util.Locale
 
 @Suppress("unused")
 public class PaparazziPlugin : Plugin<Project> {
+  private lateinit var providerFactory: ProviderFactory
```
TWiStErRob, line 50: As far as I know you can make the plugin abstract and use field injection for standard grade types

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -152,27 +130,19 @@ public class PaparazziPlugin : Plugin<Project> {
...
-
-        val localResourcePaths = localResourceDirs
-          .map { layers -> layers.flatten() }
-          .map { dirs -> dirs.map { projectDirectory.relativize(it.asFile) }.asReversed() }
-
-        task.projectResourceDirs.set(localResourcePaths)
-        task.moduleResourceDirs.set(moduleResourceDirs.relativize(projectDirectory))
-        task.aarExplodedDirs.set(aarExplodedDirs.relativize(gradleHomeDir))
+        task.projectResourceDirs.set(
+          sources.localResourceDirs.relativize(projectDirectory).reversed()
```
TWiStErRob, line 137: Have you considered these in, OR an extension of `AndroidVarianSources`?

jrodbx: Ah, yea the reversed stuff is for resource and asset priority ordering, so makes sense to move it there.  Done.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Yea, was feeling lazy, but that's a good call out.  Updated to constructor injection.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -152,27 +129,17 @@ public class PaparazziPlugin : Plugin<Project> {
...
         task.projectAssetDirs.set(
-          run {
-            val localAssetPaths = localAssetDirs
-              .map { layers -> layers.flatten() }
-              .map { dirs -> dirs.map { projectDirectory.relativize(it.asFile) }.asReversed() }
-            val moduleAssetPaths = moduleAssetDirs.relativize(projectDirectory)
-            localAssetPaths.zip(moduleAssetPaths, List<String>::plus)
-          }
+          sources.localAssetDirs.relativize(projectDirectory)
+            .zip(sources.moduleAssetDirs.relativize(projectDirectory), List<String>::plus)
```
TWiStErRob, line 140: minor, but this feels like:

minor, but this might be cleaner (if compiles)

```suggestion
          sources.localAssetDirs.zip(sources.moduleAssetDirs, List<String>::plus)
            .relativize(projectDirectory)
```

jrodbx: yea, i thought of that, but didn't want to push the providerFactory into AndroidVariantSource in order to handle the nullability of the localAssetDirs and zip doesn't take a nullable receiver.  Might just be worth splitting this out into two separate task properties as we do for resources, but I think this is fine to land for now.

kevinzheng-ap (approved):

PR #1489, opened 2024-06-25
Fix aar2jar output file paths to resolve missing compose dependencies

colinmarsch:
Previously the `outputFile` path was something like the following for each input: `/Users/cmarsch/.gradle/caches/8.8/transforms/f148f675e43cbd2a21221ba4f5df0fa0-749cc3c4-82fb-4932-a240-115b2a884778/transformed/lifecycle-livedata-core-2.6.2`, now it's more simply: `lifecycle-livedata-core-2.6.2-classes.jar`.

This resolves an issue where all `androidx.compose` dependencies in the `paparazzi` module were appearing as `Unresolved references` in the IDE.

This change is based on [the previous logic](https://github.com/cashapp/paparazzi/blob/18e71964fad9de91002b91a5c100a7a49915b15f/paparazzi/build.gradle#L228) that was used in this transfromation.


geoff-powell (approved):
hmm, nice find!!! Must have missed this when the gradle script was converted to Kotlin.

Issue #1485, opened 2024-06-19
Support adding more accessibility information


VirtualParticle:
This is currently how the text to be shown on the accessibility pane is chosen:
```kotlin
private fun SemanticsNode.accessibilityText() =
  (
    config.getOrNull(SemanticsProperties.ContentDescription)?.joinToString(", ")
      ?: config.getOrNull(SemanticsProperties.Text)?.joinToString(", ")
      ?: config.getOrNull(SemanticsProperties.StateDescription)
      ?: config.getOrNull(SemanticsActions.OnClick)?.label
      ?: config.getOrNull(SemanticsProperties.Role)?.toString()
    ).let {
    // Escape newline characters to simplify accessibility text.
    it?.replaceLineBreaks()
  }
```
I think adding some extra info like `Heading` or `Selected` would be helpful, maybe even more if it fits the circumstance. Not sure if it would be best to just add it here or make `AccessibilityRenderExtension` take in some properties for that. The way that it currently only picks one also means that we can't see both the text and the role if they're both present, for example.

geoff-powell:
Closing this since it is now fixed with @colinmarsch updates. Kudos!

PR #1491, opened 2024-06-27
Support multiple accessibility fields at once on composables

colinmarsch:
Adds support for displaying multiple accessibility feature descriptions at once, in the order Talkback reads them. Adds a test case to the sample that showcases this new functionality.

Partially addresses this issue: https://github.com/cashapp/paparazzi/issues/1485

app cash paparazzi sample_ComposeButtonA11yTest_buttonStates


colinmarsch:
cc @geoff-powell @adamalyyan who are familiar with this functionality

colinmarsch:
Ah, I see the compose-a11y tests in the `paparazzi-gradle-plugin` module now. I'll update those and ensure test cases cover this new logic.

- [ ] Update paparazzi plugin compose a11y tests

sample/src/test/java/app/cash/paparazzi/sample/ComposeButtonA11yTest.kt
```diff
@@ -0,0 +1,98 @@
...
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.stateDescription
+import androidx.compose.ui.unit.dp
+import app.cash.paparazzi.DeviceConfig
+import app.cash.paparazzi.Paparazzi
+import app.cash.paparazzi.accessibility.AccessibilityRenderExtension
+import org.junit.Rule
+import org.junit.Test
+
+class ComposeButtonA11yTest {
```
jrodbx, line 21: let's merge this with ComposeA11yTest, keeping it as a single a11y test

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -143,17 +143,22 @@ private fun SemanticsNode.findAllUnmergedNodes(): List<SemanticsNode> {
...
+  val onClickLabel =
+    if (disabled != null) null else config.getOrNull(SemanticsActions.OnClick)?.label?.let { "Double tap to $it" }
+
+  return listOfNotNull(stateDescription, mainAccessibilityText, role, disabled, onClickLabel)
+    .joinToString(", ")
     // Escape newline characters to simplify accessibility text.
-    it?.replaceLineBreaks()
-  }
+    .replaceLineBreaks()
+    .takeUnless { it.isBlank() }
```
jrodbx, line 160: nit: prefer if (...) else null for readability

jrodbx (approved):

geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -143,17 +143,22 @@ private fun SemanticsNode.findAllUnmergedNodes(): List<SemanticsNode> {
...
     config.getOrNull(SemanticsProperties.ContentDescription)?.joinToString(", ")
       ?: config.getOrNull(SemanticsProperties.Text)?.joinToString(", ")
-      ?: config.getOrNull(SemanticsProperties.StateDescription)
-      ?: config.getOrNull(SemanticsActions.OnClick)?.label
-      ?: config.getOrNull(SemanticsProperties.Role)?.toString()
-    ).let {
+  val role = config.getOrNull(SemanticsProperties.Role)?.toString()
+  val disabled = if (config.getOrNull(SemanticsProperties.Disabled) != null) "Disabled" else null
+  val onClickLabel =
+    if (disabled != null) null else config.getOrNull(SemanticsActions.OnClick)?.label?.let { "Double tap to $it" }
```
geoff-powell, line 154: nit, this assumes all accessibility is in US English. We might want to think how we want to support this for other languages/Locales if we are providing our own text.

colinmarsch: Good call out. I think removing any provided text here for now might be the safer option, however I think adding this text in with translations in the future would be a good improvement. The on click label content on it's own often relies on the rest of the screen reader description to provide relevant context.

TWiStErRob: Meet halfway with something like `<double-tap>: $it`?

colinmarsch: I updated this to `<on-click>: $it`. Since that is the field name in the `SemanticsActions` maybe this is reasonable even without translations?

TWiStErRob: Does the same apply for Disabled? Are there any other related strings?

colinmarsch: Good catch! `Disabled` is the only other supplied text here, just updated it to be `<disabled>` to match the `<on-click>` styling.

Issue #1465, opened 2024-05-30
Rendering differences between platforms seems to have increased with 1.3.4


VirtualParticle:
**Description**
In previous versions, there have been slight variations between the Paparazzi screenshots generated on Linux CI runs and local MacOS runs. This required setting a maximum percent difference above zero, although I was able to set it very low (0.00005%) and still have it work. The primary difference seems to be around text anti-aliasing and color blending more generally.

In the latest version, however, it seems like the difference has increased. It might be something with the new version of layoutlib. Seems like this is happening on the Paparazzi repo itself too: https://github.com/cashapp/paparazzi/pull/1314/files#r1544924645

**Steps to Reproduce**

Here's an example of a test with a screenshot originally generated on MacOS and failing on Linux with Paparazzi 1.3.4 (this has a difference of 0.007134%):

```kotlin
@Test
fun example() {
    paparazzi.snapshot {
        Box(
            modifier = Modifier.background(Color(0xFF000033))
        ) {
            Text("ExampleText", color = Color.White)
        }
    }
}
```

![image](https://github.com/cashapp/paparazzi/assets/20135416/21c5b45c-9f64-4c7d-89d1-052e0271145d)

With Paparazzi 1.3.3, the same test passes on Linux CI with the screenshot generated on MacOS, even with the `maxPercentDifference` set to 0.

When the accessibility overlay is enabled, the error increases to 0.083399%.

![image](https://github.com/cashapp/paparazzi/assets/20135416/e89b046d-54c3-460c-8181-d50abe7311ed)

**Expected behavior**
I'd expect the tests to be identical on all platforms but that wasn't the old behavior either. Acceptable behavior would just be to have the discrepancy be small enough to ignore. To deal with the higher difference scenarios, we'd have to set the `maxPercentDifference` to be high enough where actual bugs may be missed.

**Additional information:**
- Paparazzi Version: 1.3.4
- OS: MacOS (local) and Linux (CI)
- Compile SDK: 34
- Gradle Version: 8.7
- Android Gradle Plugin Version: 8.4.1

addonsdetector:
We are having the same issue. PR to bump to 1.3.4 fails on CI, while on our developer Macbooks  no `verify` step fails and also `record` step does not generate updated screenshot. So this blocks us from going to 1.3.4, cause we need this in our CI.

peterdk:
[...]

PR #1496, opened 2024-07-08
Replace delta text labels with pre-rendered bitmaps to sidestep cross-platform text rendering issues

jrodbx:
Detailed here: https://github.com/cashapp/paparazzi/issues/1465#issuecomment-2214868862

Some of our cross-platform verification errors stem from how the "Expected" and "Actual" labels get rasterized to images on different platforms.

TL;DR: Graphics2D.drawString() is not pixel-perfect due to things like:
* MacOSX 14+ disallows disabling antialiasing
  * https://stackoverflow.com/q/76362308
  * https://github.com/adoptium/jdk11u/blob/jdk-11.0.19+7/src/java.desktop/share/classes/sun/java2d/SunGraphics2D.java#L774
* Graphics2D defers to the native rendering engines so perfect alignment is near impossible due to platform-engine-specific implementations for antialiasing, dithering, shaders, etc.

Long term, we should replace this 3-pane image with something akin to HTML tables, but for now, pre-rendering the text labels into placeholder image labels that have been generated on MacOSX 14 should work, since image composing DOES appear to pixel-perfect across platforms.


geoff-powell (approved):

jrodbx:
Heads up: [c6e44a8](https://github.com/cashapp/paparazzi/pull/1496/commits/c6e44a85cd3cc0e5de08fde424b58babfc90a8a9) also changes when the "expected" and "actual" labels are composed onto the delta diff.  Now, it will only be added when writing the failure file to disk, since it adds an unnecessary extra consideration during snapshot verification and is more of an "output formatter" implementation detail.

PR #1521, opened 2024-08-02
Fix tests dependent on locale

ribafish:
Fix tests dependant on locale that are expecting the decimal output in a specific (US) locale.

### DIFF
--- a/paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngVerifierTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngVerifierTest.kt
@@ -8,12 +8,14 @@ import okio.Path
 import okio.Path.Companion.toOkioPath
 import okio.Path.Companion.toPath
 import org.junit.Assert.fail
+import org.junit.Before
 import org.junit.Rule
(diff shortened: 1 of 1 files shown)


paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngVerifierTest.kt
```diff
@@ -23,6 +25,12 @@ class ApngVerifierTest {
   private val secondFrame = createImage(squareOffset = Point(25, 25))
   private val thirdFrame = createImage(squareOffset = Point(45, 45))
 
+  @Before
+  fun setup() {
+    // set locale to US to avoid issues with different locales
+    Locale.setDefault(Locale.US)
```
jrodbx, line 31: reset on test teardown via @After?

jrodbx (approved):

paparazzi/src/test/java/app/cash/paparazzi/internal/apng/ApngVerifierTest.kt
ribafish: Done

PR #1517, opened 2024-07-30
Move reporting to separate task and cleanup inputs/outputs

geoff-powell:
Fixes gradle task caching issue in #1334

PR #1522, opened 2024-08-02
Fix caching of test tasks, broken by the snapshotOutputDir

ribafish:
~~Moved configuring test tasks for paparazzi to taskGraph.whenReady so we're sure in which mode it should run. Configuring is moved to `taskGraph.whenReady ` to counter the race condition, while task inputs and outputs are configured separately for verify and record tasks. This builds upon PR https://github.com/cashapp/paparazzi/pull/852 and solves the same issue.~~

Applied the fix suggested by @geoff-powell in his [PR](https://github.com/cashapp/paparazzi/pull/1517/files#diff-197fe329c4084290675667dd6b6194a6db2af00d5410525ad2d64bf4591d9b82R225-R241).

The snapshot dir inputs/outputs are set as optional and set to null if not in verify (input) or record (output) mode. This fixes the overlapping outputs caching problem.


geoff-powell (approved):
Nice!

AlexanderGH:
Friendly ping to see what the plans are to merge this. Really excited for this fix.

geoff-powell:
@~jrodbx can take a look since he has been working with the gradle folks on this.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -166,62 +166,67 @@ public class PaparazziPlugin @Inject constructor(
...
+          test.outputs.dir(reportOutputDir)
+
+          if (isVerifyRun) {
+            test.inputs.dir(snapshotOutputDir)
+              .withPropertyName("paparazzi.snapshot.output.dir")
+              .withPathSensitivity(PathSensitivity.RELATIVE)
+          }
+          if (isRecordRun) {
+            test.outputs.dir(snapshotOutputDir)
+          }
```
jrodbx, line 213: discussed offline on Slack, but sharing here as well:

Given that the Gradle build tool team raised some potential concerns about configuring the entire Test task in the  `TaskExecutionGraph.whenReady` block (i.e., missing any inputs that might affect the graph, like task dependencies), maybe a better alternative would be deferring the resolution of `isVerifyRun`/`isRecordRun` as suggested here: https://github.com/cashapp/paparazzi/pull/852#discussion_r1194683002, i.e., 

```
        val dummyUnusedDir = project.layout.projectDirectory

        test.inputs.dir(isVerifyRun.map { if (it) snapshotOutputDir else dummyUnusedDir })

        test.outputs.dir(isRecordRun.map { if (it) snapshotOutputDir else dummyUnusedDir })
          .withPropertyName("paparazzi.snapshot.output.dir")
```

(my choice of `dummyUnusedDir` might be wrong, feel free to suggest an alternative, but you get the gist...)

geoff-powell: another suggestion would be to take what you provided @~jrod and make these input/output properties optional since the else case uses a dummyDir.

ribafish: @jrodbx I'm also not really a fan of a dummy unused dir, as those inputs/outputs are still being tracked and I believe it only adds to the confusion, I believe not defining them when not needed is a cleaner solution. We could also do task configuration in two steps: the "static" one, where most of task is configured, except for stuff that relies on taskGraph, and the "dynamic" part, where the mode is decided and all the configuration tied to it is applied.

jrodbx: > We could also do task configuration in two steps: the "static" one, where most of task is configured, except for stuff that relies on taskGraph, and the "dynamic" part, where the mode is decided and all the configuration tied to it is applied.

I find dummyUnusedDir to be more clear than dividing into "static" and "dynamic" configuration, which imo is harder for maintainers to contribute/grok the intent of the configuration.

>  I believe not defining them when not needed is a cleaner solution. 

If there is a way to take @geoff-powell's proposal and make those properties optional, then i'm all for that! However, I struggle to see how this can be accomplished conditionally, while still deferring the isVerifyRun/isRecordRun resolution...in other words,

```
     if (isVerifyRun.get()) { 
        test.inputs.dir(snapshotOutputDir)
          .optional()
     }
     
     ...
```

still won't work, because of the race condition between whenReady and test task configuration!

This is likely why `@TWiStErRob` suggested (and I agree with) the dummy unused dir approach.  The point of the exercise here to defer resolution of isVerifyRun/isRecordRun.  Non-ideal?  Perhaps, but let's fix the current issue and brainstorm a better design later.  Based on our previous discussions, a more ideal solution involves a huge redesign/refactoring in order to move away from whenReady, but we really just want to solve the caching issue now for our consumers.

ribafish: Applied the fix suggested, this works much better than dummy dir 👍

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -205,8 +205,26 @@ public class PaparazziPlugin @Inject constructor(
           .withPropertyName("paparazzi.nativeRuntime")
           .withPathSensitivity(PathSensitivity.NONE)
 
-        test.outputs.dir(reportOutputDir)
-        test.outputs.dir(snapshotOutputDir)
+        test.inputs.dir(
+          isVerifyRun.flatMap {
+            project.objects.directoryProperty().apply {
+              set(if (it) snapshotOutputDir else null)
+            }
```
jrodbx, line 212: nit: is the directoryProperty step necessary, vs:

```suggestion
            if (it) snapshotOutputDir else null)
```
?

Otherwise, agreed, I like this better!

geoff-powell: so the output of `Property.map {}` cant be null so this is a workaround for that.

geoff-powell (approved):

jrodbx (approved):

Issue #1446, opened 2024-05-23
Add screenshots to the JUnit report


fcduarte:
When generating the HTML report for each test at JUnit, add the screenshots of the reference, actual and diff images.

![image](https://github.com/cashapp/paparazzi/assets/1507064/567bdef1-7658-4db2-8048-d6712afa6a29)

SofiaReilly3788:
✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 

✅ 👉👉 Click Here to Free CASH APP MONEY Generator 👈👈 ✅
 https://telegra.ph/Cash-App-Money-Generator-Link-12-20 


[...]

PR #1524, opened 2024-08-02
Add simple html snapshots to reports

geoff-powell:
This adds our delta images we generate to the junit html reports. #1446 

Screenshot 2024-08-02 at 2 26 01 PM


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -205,6 +205,37 @@ class PaparazziPluginTest {
     assertThat(snapshotsDir.exists()).isTrue()
   }
 
+  @Test
+  fun junitReportSnapshot() {
+    val fixtureRoot = File("src/test/projects/junit-report-snapshots")
+    val testFile = File(fixtureRoot, "src/test/java/app/cash/paparazzi/plugin/test/LaunchViewTest.kt")
+    val failingTestFile = File(fixtureRoot, "src/test/java/app/cash/paparazzi/plugin/test/LaunchViewTest.kt.tmp")
```
gamepro65, line 212: Instead of having a separate tmp file, should the test itself just be wrong to the golden snapshot?

geoff-powell: ya, that might make things simpler

gamepro65 (approved):
Looks good, excited to use this 🙌

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -227,6 +244,41 @@ public class PaparazziPlugin @Inject constructor(
...
+          .withPropertyName("paparazzi.snapshot.failures")
+          .withPathSensitivity(PathSensitivity.RELATIVE)
+        it.inputs.dir(testReportOutputLocation)
+          .withPropertyName("paparazzi.html.report.dir")
+          .withPathSensitivity(PathSensitivity.RELATIVE)
+
+        it.mustRunAfter(testTaskProvider)
+
+        it.doLast {
+          val reportRootDir = testReportOutputLocation.get().asFile
```
jrodbx, line 258: is there a way to customize junit reports outside of the plugin?  how does the android studio team do it?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -227,6 +244,41 @@ public class PaparazziPlugin @Inject constructor(
...
+          generatedSnapshots.asSequence().filter { snapshot ->
+            snapshot.name.startsWith("delta")
+          }.forEach { snapshot ->
+            val nameSegments = snapshot.name.split("_")
+            if (nameSegments.size >= 3) {
+              val testClassPackage = nameSegments[0].replace("delta-", "")
+              val testClass = "$testClassPackage.${nameSegments[1]}"
+              val testMethod = nameSegments[2].split(".")[0]
+
+              val testHtmlFile = File(reportRootDir, "classes${separator}$testClass.html")
```
jrodbx, line 271: Ah, I see what you did here 😺

Screenshot 2024-08-26 at 1 16 37 AM

 ....but copying/manipulating outputs from other Gradle plugins/tasks is inviting maintenance burden 😢 

Ideally, there exists a way to configure the Test Reporting plugin to customize reporting.  If not, then I'd go with @TWiStErRob 's [original suggestion](https://squaremafia.slack.com/archives/C06HF0MLUL8/p1716322051974129?thread_ts=1716321664.541989&cid=C06HF0MLUL8) and working with [PageRenderer](https://cs.android.com/search?q=PageRenderer&sq=&ss=android-studio%2Fplatform%2Ftools%2Fbase), most likely via [TestReport](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:preview/screenshot/screenshot-test-gradle-plugin/src/main/java/com/android/compose/screenshot/report/TestReport.kt) and [ScreenshotClassPageRenderer](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:preview/screenshot/screenshot-test-gradle-plugin/src/main/java/com/android/compose/screenshot/report/ScreenshotClassPageRenderer.kt?q=PageRenderer&ss=android-studio%2Fplatform%2Ftools%2Fbase).

geoff-powell: yep, just took a look at those classes. Unfortunately they are reading the junit xml and creating the entire report themselves which is unfortunate

Issue #1527, opened 2024-08-05
Different solid color rectangles of the same area have the same screenshots


And42:
**Description**
If you have multiple different components that are solid rectangles of the same color and they have the same pixel area, library will produce only one screenshot for all of them

**Steps to Reproduce**
Project: https://github.com/And42/PaparazziSameContentIssue/tree/main
3 different components: https://github.com/And42/PaparazziSameContentIssue/blob/main/app/src/main/java/com/example/paparazzisamecontentissue/MyNiceComponent.kt
Recorded screenshots: https://github.com/And42/PaparazziSameContentIssue/tree/main/app/src/test/snapshots/images

**Expected behavior**
Library should produce different screenshots for the components

**Additional information:**
- Paparazzi Version: 1.3.4
- OS: Windows
- Compile SDK: 34
- Gradle Version: 8.7
- Android Gradle Plugin Version: 8.5.1

**Screenshots**
Expected:
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceVerticalLineComponent](https://github.com/user-attachments/assets/c1756bd7-593c-4518-afde-87c3105bfea1)
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceHorizontalLineComponent](https://github.com/user-attachments/assets/96c23b1b-a840-4707-be4c-e660e12cf4bb)
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceSquareComponent](https://github.com/user-attachments/assets/83ac78d3-f635-48ca-b652-f85550febf90)

Actual:
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceVerticalLineComponent](https://github.com/user-attachments/assets/009cea5a-8d05-473e-8277-efe8189878da)
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceHorizontalLineComponent](https://github.com/user-attachments/assets/c0eaac5f-666e-4b24-96ab-dd5d6bc70074)
![com example paparazzisamecontentissue_PaparazziTest_testMyNiceSquareComponent](https://github.com/user-attachments/assets/79d50a17-7d86-4ea4-bc1b-ff271c246ba0)

**Possible fix:**
I've already found that this is most probably cause by the function that calculates the hash of the image: https://github.com/cashapp/paparazzi/blob/8912b77cbefcb422b2697d96566b5e7c50150d35/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt#L121
[...]

PR #1531, opened 2024-08-07
Add test project to verify image hashing

BrianGardnerAtl:
* Based on issue #1527 where different images resulted in the same hash
* Could not repro on Mac but potentially a windows issue?


geoff-powell (approved):

paparazzi-gradle-plugin/src/test/projects/image-hash/src/test/java/app/cash/paparazzi/plugin/test/ImageHashTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+@Composable
+fun SquareComponent() {
+  Spacer(
+    modifier = Modifier
+      .width(2.dp)
+      .height(2.dp)
+      .background(Color.Red)
+  )
+}
```
jrodbx, line 75: plugin tests are great for testing gradle things (e.g., caching), android things (e.g., svgs, resource stuff) and high level paparazzi stuff (e.g, record, verify).  i'd consider this more internal implementation details (since hashing) so can we remove this plugin test and instead add an additional test to HtmlReportWriterTest that takes these 3 image resources and asserts the json outputs to confirm different hashes, once we land on the approved fix?

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -121,6 +121,10 @@ public class HtmlReportWriter @JvmOverloads constructor(
   private fun hash(image: BufferedImage): String {
     val hashingSink = HashingSink.sha1(blackholeSink())
     hashingSink.buffer().use { sink ->
+      // write image dimensions to avoid views with the same contents but different dimensions
+      // causing a hash collision.
+      sink.writeInt(image.width)
+      sink.writeInt(image.height)
```
jrodbx, line 127: As mentioned in: https://github.com/cashapp/paparazzi/issues/1527#issuecomment-2278496452, I think this will only solve a subset of the root cause problem.  How about instead writing the (x,y) values prior to the pixel rgb data?

jrodbx (changes requested):

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
And42: Small note: https://github.com/cashapp/paparazzi/issues/1527#issuecomment-2279741497

Base on the upper: writing (x,y) prior the pixel rgb data will solve the collision problem but will not give extra protection (unless there are other edge cases) while increasing the hashing time a bit (insignificantly though)

BrianGardnerAtl:
Closing this PR since it attempts to solve two issues: one issue moving files during the recording task, and one issue related to different images producing the same hash.
I'm close to solving the file movement problem so I'll have a PR with tests up for that shortly. Then I'll get another test together for the hashing collisions.
Separating them out will make it easier to see and test each individual fix

PR #1533, opened 2024-08-12
Fix file movement when taking snapshots

BrianGardnerAtl:
This PR fixes new snapshot recording on Windows devices when two images with the same hash value are snapshotted together. Related to #1527 

The `renameTo` function behaves differently depending on the platform. In rare cases, such as when there's a hash collision, the rename would not work properly so the new snapshot file would contain the contents from the first snapshot. This results in all of snapshots with the same hash value to contain the same image data as the first one.

A follow up PR will contain modifications to the `hash()` function in an attempt to avoid further collisions.


paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -266,9 +267,106 @@ class HtmlReportWriterTest {
...
+  // that ApngWriter can consume
+  private fun BufferedImage.convertImage() = BufferedImage(
+    width,
+    height,
+    BufferedImage.TYPE_INT_ARGB_PRE // Something ApngWriter can process
+  ).apply {
+    graphics.drawImage(this, 0, 0, null)
+  }
+
+  private fun BufferedImage.pixelEqual(other: BufferedImage): Boolean {
```
BrianGardnerAtl, line 358: Is there an existing comparator for this? I looked briefly but didn't want to block review.

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -266,9 +267,106 @@ class HtmlReportWriterTest {
...
+
   private fun Instant.toDate() = Date(toEpochMilli())
 
   private fun File.lastModifiedTime(): FileTime {
     return Files.readAttributes(this.toPath(), BasicFileAttributes::class.java).lastModifiedTime()
   }
+
+  // Function that takes in an existing Buffered image and converts its type to one
+  // that ApngWriter can consume
+  private fun BufferedImage.convertImage() = BufferedImage(
```
BrianGardnerAtl, line 350: This is needed because the test images do not have a color type that works with `ApngWriter`. I find it odd that I need to do this since I took the images from a test project that the `ApngWriter` created. Do we need to expand the list of valid image types for the writer?

Issue #1422, opened 2024-05-13
How to wait for recomposition


karsie:
**Description**
Releasenotes for 1.3.0 say: Recomposition does not happen unless lifecycle is RESUMED
However, when we record previews (in combination with Showkase), only the first composition is snapshotted.
This doesn't allow us to show dialogs for example, or components that rely on onTextLayout or onGloballyPositioned.
Whatever I tried to keep the state as resumed, it didn't work.
Any tips?

**Steps to Reproduce**
I tried adding something like this to our preview functions:
LaunchedEffect(true) {
  launch {
    delay(1000)
  }.join()
}

**Expected behavior**

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: maxos
- Compile SDK: 34
- Gradle Version: 8.4
- Android Gradle Plugin Version: 8.3.2

**Screenshots**

geoff-powell:
Seems to be related to https://github.com/cashapp/paparazzi/issues/1414

PR #1537, opened 2024-08-15
Fix pending recompositions to render similarly to Android Studio

### LINKED ISSUE #1414: Popup Content Missing in Paparazzi Rendered Snapshots for Composable Functions
**Description**
When attempting to record a snapshot using Paparazzi for a composable function with a Popup, the content of the Popup disappears in the recorded snapshot.

**Steps to Reproduce**

1. Create a composable function that includes a Popup to display information.
Sample Code:
```kotlin
@Composable
fun TestPopup() {
    MaterialTheme {
        Column(modifier = Modifier.fillMaxSize()) {
            Text(text = "Sample text")
            Popup(alignment = Alignment.BottomCenter) {
                Text(
                    text = "Hello Snapshot!",
                    modifier = Modifier
                        .fillMaxWidth()
                        .heightIn(min = 60.dp),
                    textAlign = TextAlign.Center,
                    color = Color.Black
                )
            }
        }
    }
}
```
2. Attempt to record a snapshot using Paparazzi.

**Expected behavior**
The recorded snapshot should accurately capture the content of the Popup along with the rest of the composable function.

**Actual behavior**
In the recorded snapshot, the Popup is drawn, but the content within it is missing.

**Additional information:**
- Paparazzi Version: 1.3.1
- OS: Mac - Sonoma 14.2.1
- Compile SDK: 34
- Gradle Version:
- Android Gradle Plugin Version:

**Screenshots**
Preview Screenshot:
Screenshot 2024-05-07 at 11 17 07
Recorded Snapshot:
Screenshot 2024-05-07 at 10 38 52

geoff-powell:
Fixes #1414 and #1422

AS Preview calls a second render in LayoutLib after the initial one is fired. 
https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:preview-designer/src/com/android/tools/idea/preview/representation/CommonPreviewRepresentation.kt;l=680-684?q=onAfterRender

This mimics this approach while only doing so if Compose is in the class path and the recomposer is not idle.


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -285,21 +285,29 @@ public class PaparazziSdk @JvmOverloads constructor(
       viewGroup.addView(modifiedView)
       for (frame in 0 until frameCount) {
         val nowNanos = (startNanos + (frame * 1_000_000_000.0 / fps)).toLong()
-        withTime(nowNanos) {
-          val result = renderSession.render(true)
-          if (result.status == ERROR_UNKNOWN) {
-            throw result.exception
+        var result = withTime(nowNanos) {
```
BrianGardnerAtl, line 288: I like isolating just the render call in the `withTime {}` block, it makes it feel less cluttered

paparazzi/src/main/java/app/cash/paparazzi/internal/compose/RecomposerWatcher.kt
```diff
@@ -0,0 +1,86 @@
...
+    result: Result,
+    onAdditionalRender: () -> Result
+  ): Result {
+    if (hasComposeRuntime && result.status != Result.Status.ERROR_UNKNOWN) {
+      // If there is a recomposition that needs to happen, we need to trigger it within the context of the first frame.
+      val recomposerInstance = recomposer as? Recomposer
+      val hasPendingWork = recomposerInstance?.hasPendingWork
+      if (frame < frameLimit && hasPendingWork == true) {
+        if (additionalRenderCount >= additionalRenderLimit) {
+          paparazziLogger.info(
```
BrianGardnerAtl, line 52: What are your thoughts on making this an error? If we're taking a single snapshot and the view never stabilizes an error would be a nice debugging tool. With the info log it could be ignored / missed in all the output.

Maybe we could provide a flag that makes it an error or not?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -611,7 +620,7 @@ public class PaparazziSdk @JvmOverloads constructor(
         |              android:layout_height="${if (renderingMode.vertAction == RenderingMode.SizeAction.SHRINK) "wrap_content" else "match_parent"}"/>
     """.trimMargin()
 
-    private fun isPresentInClasspath(vararg classNames: String): Boolean {
+    internal fun isPresentInClasspath(vararg classNames: String): Boolean {
```
BrianGardnerAtl, line 623: We can do it later but can we move this out of the sdk into a separate place?

paparazzi/src/main/java/app/cash/paparazzi/internal/compose/RecomposerWatcher.kt
```diff
@@ -0,0 +1,86 @@
...
+        }
+        val newResult = onAdditionalRender()
+        if (newResult.status == ERROR_UNKNOWN) {
+          throw newResult.exception
+        }
+
+        return renderForIdleRecomposer(frame, additionalRenderCount + 1, newResult, onAdditionalRender)
+      }
+
+      if (hasPendingWork == true) {
```
BrianGardnerAtl, line 65: Ditto on the error comment above

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -285,21 +285,29 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+        }
+
+        // If we have pendingTasks run recomposer to ensure we get the correct frame.
+        result = recomposerWatcher.onRenderResult(frame, result) {
+          withTime(nowNanos) {
+            renderSession.render(true)
           }
+        }
+
+        if (result.status == ERROR_UNKNOWN) {
```
BrianGardnerAtl, line 299: We can move this above the `onRenderResult` call above since that function will throw the exception on subsequent renders. This could help bail earlier if the first render is an error

BrianGardnerAtl (commented):
I think it looks pretty good! I left a few comments but I really like the logic in the `RecomposerWatcher`

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
geoff-powell: I might move this into `RecomposerWatcher.onRenderResult()` to do the same for any renders that fail.

paparazzi/src/main/java/app/cash/paparazzi/internal/compose/RecomposerWatcher.kt
geoff-powell: hmm, maybe I can up it to a warning. I wasn't able to hit this via a repro case but its more of insight if this does occur.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
geoff-powell: Yep, I can do this as a follow up

paparazzi/src/main/java/app/cash/paparazzi/internal/compose/RecomposerWatcher.kt
geoff-powell: Ya I felt like warning made sense because it could be an expected result of the state you are trying to test. Not married to `warning` and can up it to error.

jrodbx:
(oops, sorry about that, meant to write a comment 😅)

jrodbx:
> AS Preview calls a second render in LayoutLib after the initial one is fired.
https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:preview-designer/src/com/android/tools/idea/preview/representation/CommonPreviewRepresentation.kt;l=680-684?q=onAfterRender

Nice work!  Out of curiosity: were you able to attach a debugger and step through this code path in any way?  When I tried, I set a breakpoint in `RenderSessionImpl#renderAndBuildResult` and could only see:

Screenshot 2024-09-03 at 12 29 28 AM


However, working with the source link you provided, I think the execution path below makes sense.  Can you confirm?:

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:preview-designer/src/com/android/tools/idea/preview/representation/CommonPreviewRepresentation.kt;l=680-684?q=onAfterRender
```
# CommonPreviewRepresentation.kt: 679

  private fun onAfterRender() {
    // We need to run any callbacks that have been registered during the rendering of the preview
    surface.sceneManagers.forEach {
      onAfterRender(it)
      it.executeCallbacksAndRequestRender()
    }
    previewViewModel.afterPreviewsRefreshed()
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneManager.java;l=638?q=executeCallbacksAndRequestRender
```
# LayoutlibSceneManager.java: 638

  public @NotNull CompletableFuture<Void> executeCallbacksAndRequestRender() {
    return executeCallbacksAsync().thenCompose(b -> requestRenderAsync());
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneManager.java;l=442?q=executeCallbacksAndRequestRender
```
# LayoutlibSceneManager.java: 442  

  public CompletableFuture<Void> requestRenderAsync() {
    return requestRenderAsync(getTriggerFromChangeType(getModel().getLastChangeType()));
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneManager.java;l=411?q=executeCallbacksAndRequestRender
```
# LayoutlibSceneManager.java: 411  

  protected CompletableFuture<Void> requestRenderAsync(@Nullable LayoutEditorRenderResult.Trigger trigger) {
    ...

    return myLayoutlibSceneRenderer.renderAsync(trigger).thenCompose((unit) -> CompletableFuture.completedFuture(null));
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneRenderer.kt;l=239
```
# LayoutlibSceneRenderer.kt: 239

  fun renderAsync(trigger: LayoutEditorRenderResult.Trigger?): CompletableFuture<Unit> {
    return scope.launch { requestRenderAndWait(trigger) }.asCompletableFuture()
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneRenderer.kt;l=263
```
# LayoutlibSceneRenderer.kt: 263

 suspend fun requestRenderAndWait(trigger: LayoutEditorRenderResult.Trigger?) {
    val request = enqueueRenderRequest(trigger)
    // Suspends until this or any newer request is processed
    lastProcessedRenderRequestTime.first { it >= request.requestTime }
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneRenderer.kt;l=270
```
# LayoutlibSceneRenderer.kt: 270

  private suspend fun enqueueRenderRequest(
    trigger: LayoutEditorRenderResult.Trigger?
  ): RenderRequest {
    // Mutex is needed to guarantee that requests are sent to the channel in order of their
    // requestTime
    return requestsLock.withLock { RenderRequest(trigger).also { requestsChannel.send(it) } }
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneRenderer.kt;l=208
```
# LayoutlibSceneRenderer.kt: 208

  init {
    scope.launch {
      requestsChannel.receiveAsFlow().collect {
        renderIsRunning.set(true)
        if (isActive.get()) {
          val reverseUpdate = AtomicBoolean(false)
          val rerenderIfNeeded = sceneRenderConfiguration.doubleRenderIfNeeded.getAndSet(false)
          doRender(it, reverseUpdate)
          ...
        } 
		...
        renderIsRunning.set(false)
      }
    }
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:designer/src/com/android/tools/idea/uibuilder/scene/LayoutlibSceneRenderer.kt;l=290
```
# LayoutlibSceneRenderer.kt: 290
  
  private suspend fun doRender(
    request: RenderRequest,
    reverseUpdate: AtomicBoolean,
  ): RenderResult? {
    ...
      renderTask?.let {
        ...
        result = it.render().await()
        ...
      }
    ...
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/RenderTask.java;l=1198?q=RenderTask.render()
```
RenderTask.java: 1198

  public CompletableFuture<RenderResult> render() {
    return render(myCachingImageFactory);
  }  
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/RenderTask.java;l=1182?q=RenderTask.render()
```
RenderTask.java: 1182
  
  public CompletableFuture<RenderResult> render(@NotNull IImageFactory factory) {
    myImageFactoryDelegate = factory;
    // If a re-render is happening after changing the quality up, then we need to re-measure
    // to avoid the rendered image to show things out of place.
    boolean forceMeasure = myTargetQuality > myCurrentQuality;
    myCurrentQuality = myTargetQuality;
    return renderInner(forceMeasure);
  }
```

https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/RenderTask.java;l=1084?q=RenderTask.render()
```
RenderTask.java: 1084  
  
  private CompletableFuture<RenderResult> renderInner(boolean forceMeasure) {
    ...
    return inflateCompletableResult.thenCompose(inflateResult -> {
      try {
        long startRenderTimeMs = System.currentTimeMillis();
        return runAsyncRenderAction(() -> {
          myTestEventListener.onBeforeRender();
          myRenderSession.render(forceMeasure);  <===== finally, the entry point we can see in the stack trace!
```

paparazzi-gradle-plugin/src/test/projects/compose-recomposition/src/test/java/app/cash/paparazzi/plugin/test/RecomposeTest.kt
```diff
@@ -39,4 +47,46 @@ class RecomposeTest {
...
+              .background(Color.Green)
+              .align(Alignment.CenterHorizontally)
+              .onGloballyPositioned {
+                globalPosition = it.localToRoot(Offset.Zero)
+              }
+          )
+        }
+      }
+    }
+  }
```
jrodbx, line 91: confirming: these tests fail before and pass after the change, correct?

paparazzi/src/main/java/app/cash/paparazzi/internal/compose/RecomposerWatcher.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.compose.ui.platform.WindowRecomposerPolicy
+import androidx.compose.ui.platform.createLifecycleAwareWindowRecomposer
+import app.cash.paparazzi.PaparazziSdk.Companion.isPresentInClasspath
+import app.cash.paparazzi.internal.PaparazziLogger
+import com.android.ide.common.rendering.api.Result
+import com.android.ide.common.rendering.api.Result.Status.ERROR_UNKNOWN
+import kotlin.coroutines.CoroutineContext
+
+@InternalComposeUiApi
+internal class RecomposerWatcher(
```
jrodbx, line 17: given the limited use case, extracting this to its own class is a bit overkill, imo.  I really liked where https://github.com/cashapp/paparazzi/commit/505720f93df6685f44d40870ee0637065c0614f2 was heading, let's stick with that!

jrodbx (changes requested):

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -309,16 +326,54 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+    if (result.status == ERROR_UNKNOWN) {
+      throw result.exception
+    }
+
+    if (hasComposeRuntime) {
+      // If there is a recomposition that needs to happen, we need to trigger it within the context of the first frame.
+      val recomposerInstance = recomposer as? Recomposer
+      val hasPendingWork = recomposerInstance?.hasPendingWork
+
+      if (frame < frameLimit && hasPendingWork == true) {
```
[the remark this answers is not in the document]
geoff-powell: Yep, this does the same logic but allows for customization of what frames you want to retry.
Happy to remove this for simplification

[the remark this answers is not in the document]
jrodbx: > allows for customization of what frames you want to retry.

yea, this is indeed what I'd like to avoid.  To be clear, this pending recomposition trick is already a hack that Android Studio has decided to allow given how the Compose framework (and more specifically, in this example, Popup) requires another layout pass.  We could take a stance of not supporting it, but I imagine that would hurt adoption.

That being said, customization would allow developers to work around improperly-designed UI elements that are skipping multiple frames before they finally draw to the Canvas.  I am fairly Android Studio would not support as that would be a developer error, and neither should we.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -317,13 +323,47 @@ public class PaparazziSdk @JvmOverloads constructor(
...
       if (hasComposeRuntime) {
         forceReleaseComposeReferenceLeaks()
       }
     }
   }
 
+  private fun renderForIdleRecomposer(
+    recomposer: Any?,
+    frame: Int,
+    additionalRenderCount: Int,
```
BrianGardnerAtl, line 335: Are we planning on adding a future api to allow users to configure this and the frame parameter?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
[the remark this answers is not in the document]
jrodbx: > Happy to remove this for simplification

Yes, let's remove/inline renderForIdleRecomposer since its more handling of an edge case vs a general-purpose solution.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
geoff-powell: That was the initial idea but I can strip this down for now

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -285,21 +294,41 @@ public class PaparazziSdk @JvmOverloads constructor(
...
-          if (result.status == ERROR_UNKNOWN) {
-            throw result.exception
-          }
 
-          val image = bridgeRenderSession.image
-          if (validateAccessibility) {
-            require(renderExtensions.isEmpty()) {
-              "Running accessibility validation and render extensions simultaneously is not supported."
+        // If we have pendingTasks run recomposer to ensure we get the correct frame.
+        var result = withTime(nowNanos) {
```
jrodbx, line 299:
```suggestion
        var hasPendingWork = false
        withTime(nowNanos) {
          val result = renderSession.render(true)
          if (result.status == ERROR_UNKNOWN) {
            throw result.exception
          }

          if (hasComposeRuntime && recomposer != null) {
            // If we have pending tasks, we need to trigger it within the context of the first frame.
            if (frame == 0 && (recomposer as Recomposer).hasPendingWork) {
              hasPendingWork = true
            }
          }
        }
```

then within the same loop iteration of `frame`:

```
        if (hasPendingWork) {
          withTime(nowNanos) {
            val result = renderSession.render(true)
            if (result.status == ERROR_UNKNOWN) {
              throw result.exception
            }
          }

          val recomposerInstance = recomposer as Recomposer
          if (recomposerInstance.hasPendingWork) {
            logger.warning(
              "Pending work detected. This may cause unexpected results in your generated snapshots. ${recomposerInstance.changeCount}"
            )
          }
        }
```

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -285,21 +294,41 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+        }
+
+        val image = bridgeRenderSession.image
+        if (validateAccessibility) {
+          require(renderExtensions.isEmpty()) {
+            "Running accessibility validation and render extensions simultaneously is not supported."
+          }
+          validateLayoutAccessibility(modifiedView, image)
+        }
+        onNewFrame(scaleImage(frameImage(image)))
```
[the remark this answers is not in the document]
geoff-powell: actually its in the correct spot. Now its within the for loop for each frame instead of `withTime { }`

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -255,17 +256,24 @@ public class PaparazziSdk @JvmOverloads constructor(
         // Initialize the choreographer at time=0.
       }
 
+      // The consumer may not have compose runtime on the classpath, so we don't reference the type.
+      var recomposer: Any? = null
+
+      // By default, Compose UI uses its own implementation of CoroutineDispatcher, `AndroidUiDispatcher`.
+      // Since this dispatcher does not provide its own implementation of Delay, it will default to using DefaultDelay which runs
+      // async to our test Handler. By initializing Recomposer with Dispatchers.Main, Delay will now be backed by our test Handler,
+      // synchronizing expected behavior.
```
jrodbx, line 265: nit: this comment should move back to where it was on line 272

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -315,10 +344,10 @@ public class PaparazziSdk @JvmOverloads constructor(
     }
   }
 
-  private fun withTime(
+  private fun <T> withTime(
     timeNanos: Long,
-    block: () -> Unit
-  ) {
+    block: () -> T
+  ): T {
```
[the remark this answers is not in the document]
jrodbx: result no longer being used, so can be reverted

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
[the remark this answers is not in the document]
jrodbx: yep, that's what I meant; in other words, it's nice to move it out of the `withTime` block.

jrodbx (approved):
nice work, some small final cleanup

jisungbin:
Thank you for your hard work to bring us this long-awaited feature!

PR #1541, opened 2024-08-18
Replace all accessing of gradle properties

ansman:
The old usage (`findProperty` or `properties`) is not compatible with project isolation. The new `providers.gradleProperty` is, however.


jrodbx (approved):
thanks!

PR #1544, opened 2024-08-21
Add version ref for related compose dependencies

geoff-powell:


gradle/libs.versions.toml
```diff
@@ -16,13 +17,14 @@ compileSdk = "34"
...
+androidx-compose-ui-android = { module = "androidx.compose.ui:ui-android", version.ref = "compose" }
 androidx-recyclerview = { module = "androidx.recyclerview:recyclerview", version = "1.3.2" }
 
 bytebuddy-agent = { module = "net.bytebuddy:byte-buddy-agent", version.ref = "bytebuddy" }
 bytebuddy-core = { module = "net.bytebuddy:byte-buddy", version.ref = "bytebuddy" }
 
-composeUi-material = { module = "androidx.compose.material:material", version = "1.6.8" }
+compose-runtime = { module = "androidx.compose.runtime:runtime", version.ref = "compose" }
+composeUi-material = { module = "androidx.compose.material:material", version.ref = "compose" }
 composeUi-uiTooling = { module = "androidx.compose.ui:ui-tooling" }
```
BrianGardnerAtl, line 28: should this also be anchored to the compose version?

paparazzi-annotations/src/main/java/app/cash/paparazzi/annotations/PaparazziPreviewData.kt
```diff
@@ -0,0 +1,24 @@
+package app.cash.paparazzi.annotations
+
+import androidx.compose.runtime.Composable
+
+public sealed interface PaparazziPreviewData {
```
BrianGardnerAtl, line 5: I have a guess what this interface is for but I'm not positive. A comment would help explain its purpose.

gradle/libs.versions.toml
geoff-powell: ya, I think we are using BOM so we shouldn't need the version refs. Let me double check

paparazzi-annotations/src/main/java/app/cash/paparazzi/annotations/PaparazziPreviewData.kt
jrodbx: I'd go a bit further and say whatever is using this should also be part of this PR; otherwise, it's challenging to understand the intent behind this.

would you mind splitting up the PRs in a way that makes incremental development and review easier?

paparazzi-annotations/api/paparazzi-annotations.api
```diff
@@ -1,3 +1,39 @@
 public abstract interface annotation class app/cash/paparazzi/annotations/Paparazzi : java/lang/annotation/Annotation {
 }
 
+public abstract interface class app/cash/paparazzi/annotations/PaparazziPreviewData {
+}
+
+public final class app/cash/paparazzi/annotations/PaparazziPreviewData$Default : app/cash/paparazzi/annotations/PaparazziPreviewData {
```
jrodbx, line 7: this is implementation detail right?  so this shouldn't be exposed as public api

jrodbx (changes requested):

paparazzi-annotations/src/main/java/app/cash/paparazzi/annotations/PaparazziPreviewData.kt
geoff-powell: yep, this should be in the a pr down the stack. Cleaning it up

paparazzi-annotations/api/paparazzi-annotations.api
jrodbx: thanks.  now this PR is a bunch of dependency version bumps, can this be moved (maybe even split if multiple concerns are being bundled) to its/their respective PR(s)?

geoff-powell: no dep bump is in this PR, just defined the version as a variable

jrodbx: oh gotcha, missed that, my bad.  in that case, let's change the commit title and PR title appropriately, i.e., "Align compose dependencies to same version"?

geoff-powell: I see, yes this pr is just those updates, so i changed them appropriately

jrodbx: I see you updated the PR title but I don't think that will update the commit title when merging into history.  Also, nit: cleanup is a bit vague.

jrodbx (approved):

PR #1554, opened 2024-08-24
Add preview param support for @Paparazzi

### LINKED ISSUE #1687: How to pass Composable Preview parameters to the Snapshot tests
Hi Team, I have been using Paparazzi for snapshot tests of our views built using Jetpack compose. We are using Reflections API to scan the code to find the method annotated with previews. These methods are then passed to Paparazzi for snapshot tests. Can we have capability of passing parameters to previews using ParameterPreviewProvider class from previews to snapshot tests. Is it possible ?. What about different previews that got generated with annotations like @PreviewScreenSizes, @PreviewFontScales, @PreviewLightDark, and @PreviewDynamicColors .

Here is my paparazzi snapshot test class

`class PaparazziTest {

    @get:Rule
    val paparazzi: Paparazzi = Paparazzi()

    @Test
    fun generateSnapshots() {
        val reflections = Reflections("xxx.zpreview", Scanners.MethodsAnnotated)
        val methods = reflections.getMethodsAnnotatedWith(Preview::class.java)
            .filterNot { java.lang.reflect.Modifier.isPrivate(it.modifiers) }

        val composeProxy = ComposeProxy()
        for (method in methods) {
            val name = "${method.declaringClass.simpleName}_${method.name}"

            try {
                snapshot(name = name) {
                    composeProxy.wrappedInstance(method).GenerateSnapshot()
                }
            } catch (e: Throwable) {
                throw RuntimeException("Error generating snapshot $name", e)
            }
        }
    }

    private fun snapshot(name: String, composable: @Composable () -> Unit) {
        paparazzi.snapshot(name) {
            Box {
                composable()
            }
        }
    }

    private interface ComposeProxyInterface {

        @Composable
        fun GenerateSnapshot()
    }

    @Suppress("EmptyFunctionBlock")
    private class ComposeProxy : ComposeProxyInterface {
        @Composable
        override fun GenerateSnapshot() {
        }

        fun wrappedInstance(composableFunction: Method): ComposeProxyInterface {
            val proxyClass = Proxy.getProxyClass(
                ComposeProxy::class.java.classLoader,
                ComposeProxyInterface::class.java
            )

            val invocationHandler = InvocationHandler { _, _, args: Array<out Any>? ->
                composableFunction.invoke(null, *args!!)
            }

            return proxyClass
                .getConstructor(InvocationHandler::class.java)
                .newInstance(invocationHandler) as ComposeProxyInterface
        }
    }
}`

geoff-powell:
Adds the ability to read `@PreviewParameters` for @Composable's with the @Paparazzi annotation

fixes #1687


paparazzi-annotations/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Run `:recordPaparazziDebug` in your module to generate preview snapshots (and optionally verify them using `:verifyPaparazziDebug`) as you normally would.
+
+A test class to generate snapshots for annotated previews will automatically be generated.
+If you prefer to define a custom snapshot test, you mey disable test generation by adding the following to your `build.gradle` file.
+
+```groovy
+paparazziPreview {
+  generateTestClass = false
```
jrodbx, line 32: What's the proposed use case where someone would mark a `@Preview` with `@Paparazzi`, but not want a test class to be generated as a result of the latter?

geoff-powell: you would do this if you wanted to create your own implementation of the generated test class instead of using the codegen version.

jrodbx: I understand that from the property naming convention, but can you elaborate more on a use case where this might be desirable?  And how would that differ from today where one manually crafts the test class?

Unless I'm missing something, a nice layered design decision would be to either write the test manually or let the annotation generate it for you.  Otherwise, integrating with customizations and keeping up with Google's @Preview and @PreviewParameter exposes us to maintenance burden.

geoff-powell: I ended up making this a gradle property instead of gradle extension dsl.

jrodbx: I see that, however, that still bleeds the implementation detail of `PaparazziPreviewData`, per

> Especially, given the plan to replace this test source codegen with direct SDK snapshotting anyway, I'd be even more wary of exposing customization hooks from the get-go.

nak5ive (approved):

paparazzi-annotations/build.gradle
```diff
@@ -3,5 +3,6 @@ apply plugin: 'org.jetbrains.kotlin.plugin.compose'
 apply plugin: 'com.vanniktech.maven.publish'
 
 dependencies {
-  implementation libs.compose.runtime
+  compileOnly libs.compose.runtime
+  compileOnly libs.tools.layoutlib
```
geoff-powell, line 7: nit: I don't love having to depend on this but we need access to `android.content.res.Configuration` used in preview. Maybe we could hardcode these values instead?

PR #1556, opened 2024-08-26
Hash collision fix

BrianGardnerAtl:
Add the pixel coordinates to the hashing sink. Last PR needed to address #1527

### DIFF
--- a/paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
+++ b/paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
@@ -368,8 +368,8 @@ class HtmlReportWriterTest {
           squareFrameHandler.handle(convertedSquare)
         }
       }
-      // Verify that there are 3 files in the reports/paparazzi/images file
-      // This should confirm that the snapshot images used in the test report are correct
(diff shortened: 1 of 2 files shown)


geoff-powell (approved):

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
```diff
@@ -385,6 +385,71 @@ class HtmlReportWriterTest {
     }
   }
 
+  @Test
```
jrodbx, line 388: doesn't this test already exist above from the previous PR that was merged in?

jrodbx (approved):
let's merge this test with the previous, it's 90% similar

paparazzi/src/test/java/app/cash/paparazzi/HtmlReportWriterTest.kt
BrianGardnerAtl: That test verified that the snapshots are saved correctly, this one verifies that the files in the report folder are also correct. This fixes the bug where recording snapshots with the same hash would display the same image in the paparazzi html report because they generated the same file name

PR #1559, opened 2024-08-28
Update macOS 14 to latest

geoff-powell:
macOS-latest is [now](https://github.com/actions/runner-images?tab=readme-ov-file#available-images) macOS 14 arm64 so we no longer need to target a direct version.

Add support for macOS 14 x64 to cover more end user scenarios.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [
           # Check https://github.com/actions/runner-images?tab=readme-ov-file#available-images for actual versions for latest
-          macOS-latest
+          macOS-latest,
           windows-latest,
           ubuntu-latest
         ]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [
           # Check https://github.com/actions/runner-images?tab=readme-ov-file#available-images for actual versions for latest
-          macOS-latest, # arm64
+          macOS-latest
           windows-latest,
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -18,9 +18,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          # Migrate back to macOS-latest once 'latest' workflow label is updated (April – June 2024)
-          # https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
-          macOS-14,
+          # Check https://github.com/actions/runner-images?tab=readme-ov-file#available-images for actual versions for latest
+          macOS-latest, # arm64
+          macOS-14-latest-large, # intel
```
jrodbx, line 23:
```suggestion
```

Afaik, Intel Macs are no longer supported?  If correct, it'd be best not increasing the test matrix and build times.

jrodbx (approved):

.github/workflows/build.yml
```diff
@@ -18,9 +18,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          # Migrate back to macOS-latest once 'latest' workflow label is updated (April – June 2024)
-          # https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
-          macOS-14,
+          # Check https://github.com/actions/runner-images?tab=readme-ov-file#available-images for actual versions for latest
+          macOS-latest
```
jrodbx, line 22:
```suggestion
          macOS-latest,
```

PR #1563, opened 2024-08-29
Generate separate report for including snapshots

geoff-powell:
https://github.com/cashapp/paparazzi/issues/1446

This borrows a bunch of classes from Gradle reporting to generate a junit html report with diff images included. It will add  snapshots to the existing junit report.
This was mainly in inspiration of the Google team's [compose preview snapshot gradle plugin](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:preview/screenshot/screenshot-test-gradle-plugin/). Kudos!

Sample html report:
[sample-html-report.zip](https://github.com/user-attachments/files/16801575/sample-html-report.zip)

Screenshot 2024-08-29 at 11 34 59 AM


paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -92,7 +92,7 @@ public class HtmlReportWriter @JvmOverloads constructor(
       val snapshotDir = if (fps == -1) imagesDirectory else videosDirectory
       val goldenDir = if (fps == -1) goldenImagesDirectory else goldenVideosDirectory
       val hashes = mutableListOf<String>()
-      val snapshotTmpFile = File(snapshotDir, snapshot.toFileName(extension = "temp.png"))
```
geoff-powell, line 95: needed a better delimeter to denote method names from custom names appended at the end

jrodbx: what's the difference with/without this delimiter change?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/AllTestResults.kt
```diff
@@ -0,0 +1,63 @@
+package app.cash.paparazzi.gradle.reporting
+
+import java.util.TreeMap
+
+/**
+ *
+ * Custom test results based on Gradle's AllTestResults
+ */
+public class AllTestResults : CompositeTestResults(null) {
```
jrodbx, line 9: internal?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ClassTestResults.kt
```diff
@@ -0,0 +1,48 @@
+package app.cash.paparazzi.gradle.reporting
+
+import java.util.TreeSet
+
+/**
+ * Custom ClassTestResults based on Gradle's ClassTestResults
+ */
+public class ClassTestResults(
```
jrodbx, line 8: internal?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ScreenshotClassPageRenderer.kt
```diff
@@ -0,0 +1,147 @@
+package app.cash.paparazzi.gradle.reporting
+
+import com.android.build.gradle.internal.test.report.ErroringAction
+import com.android.build.gradle.internal.test.report.SimpleHtmlWriter
```
jrodbx, line 4: We should probably avoid relying on internal classes, any alternatives?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -169,13 +173,53 @@ public class PaparazziPlugin @Inject constructor(
...
+          val resultsDir = xmlTestOutputDir.get().asFile
+
+          val report = app.cash.paparazzi.gradle.reporting.TestReport(
+            resultDir = resultsDir,
+            reportDir = reportDir,
+            failureSnapshotDir = failureSnapshotsDir,
+            applicationId = variant.namespace.get(),
+            variantKey = variant.name
+          )
+          report.generateScreenshotTestReport()
```
jrodbx, line 219: move into separate test reporting task?

paparazzi-gradle-plugin/api/paparazzi-gradle-plugin.api
```diff
@@ -24,3 +24,219 @@ public abstract class app/cash/paparazzi/gradle/PrepareResourcesTask : org/gradl
...
+	public fun <init> ()V
+	protected abstract fun getContentRenderer ()Lorg/gradle/reporting/ReportRenderer;
+	protected abstract fun getHeaderRenderer ()Lorg/gradle/reporting/ReportRenderer;
+	protected final fun getModel ()Ljava/lang/Object;
+	protected abstract fun getTitle ()Ljava/lang/String;
+	public fun render (Ljava/lang/Object;Lcom/android/build/gradle/internal/test/report/SimpleHtmlWriter;)V
+	public synthetic fun render (Ljava/lang/Object;Ljava/lang/Object;)V
+}
+
+public final class app/cash/paparazzi/gradle/reporting/TabsRenderer : org/gradle/reporting/ReportRenderer {
```
jrodbx, line 182: i suspect most of this can be made internal and not part of the public plugin api

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ScreenshotClassPageRenderer.kt
geoff-powell: what about `org.gradle.internal.html.SimpleHtmlWriter` instead?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: its already in a separate task 🤔 ? `reportPaparazzi{Variant}UnitTest`

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
geoff-powell: before: `delta-app.cash.paparazzi.sample_LaunchViewTest_pixel3_differentThemes_light.png`

after: `delta-app.cash.paparazzi.sample_LaunchViewTest_pixel3_differentThemes-light.png`

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ScreenshotClassPageRenderer.kt
geoff-powell: I updated the pr with a couple more internal classes to remove these usages

paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
```diff
@@ -71,6 +72,7 @@ public class SnapshotVerifier @JvmOverloads constructor(
           """.trimIndent()
         )
         ImageUtils.assertImageSimilar(
+          relativePathFailure = expectedFailure.path,
```
jrodbx, line 75: as a byproduct of the work in https://github.com/cashapp/paparazzi/issues/1465#issuecomment-2214901484, I'd like to avoid passing more state into this function.

paparazzi/src/test/java/app/cash/paparazzi/helper/TestSnapshotVerifier.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import app.cash.paparazzi.Snapshot
+import app.cash.paparazzi.SnapshotHandler
+import app.cash.paparazzi.internal.ImageUtils
+import org.junit.rules.TemporaryFolder
+import java.awt.image.BufferedImage
+import java.io.File
+import javax.imageio.ImageIO
+
+internal class TestSnapshotVerifier(val tempDir: TemporaryFolder) : SnapshotHandler {
```
jrodbx, line 11:
```suggestion
internal class TestSnapshotVerifier(val failureDir: File) : SnapshotHandler {
```

nit: this class doesn't have to know about TemporaryFolder from an API perspective

paparazzi/src/test/java/app/cash/paparazzi/helper/TestSnapshotVerifier.kt
```diff
@@ -0,0 +1,36 @@
...
+      override fun handle(image: BufferedImage) {
+        val expected = File("src/test/resources/${snapshot.name}.png")
+        val failed = File("src/test/resources/${snapshot.name}/failed.png")
+        ImageUtils.assertImageSimilar(
+          relativePath = expected.path,
+          image = image,
+          goldenImage = ImageIO.read(expected),
+          maxPercentDifferent = 0.1,
+          relativePathFailure = failed.path,
+          failureDir = tempDir.newFolder()
```
jrodbx, line 27:
```suggestion
          failureDir = failureDir
```

per the comment in this same source file; however, I'd really prefer not to pass this into assertImageSimilar

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -169,13 +173,70 @@ public class PaparazziPlugin @Inject constructor(
...
+      val htmlTestReportOutputDir = project.objects.directoryProperty()
+
       project.gradle.taskGraph.whenReady { graph ->
         isRecordRun.set(recordTaskProvider.map { graph.hasTask(it) })
         isVerifyRun.set(verifyTaskProvider.map { graph.hasTask(it) })
+
+        // Unfortunately, we have to wait till the task graph is ready. When using a custom reporting base directory
+        // the actual report directory gets overwritten after task configuration. Seems like AGP is not respecting this extensions config.
+        // So we wait to set the testReportOutputDir until the task graph is ready.
+        htmlTestReportOutputDir.set(testTaskProvider.firstOrNull()?.reports?.html?.outputLocation?.get())
```
jrodbx, line 186: please don't add more to this callback, instead let's use a lambda/property to defer resolution.

jrodbx: Also, can you elaborate more on "Seems like AGP is not respecting this extensions config".  In other words, what is AGP actually doing here?  I would guess they're not using `taskGraph.whenReady`.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: sorry, i wasn't clear, and actually I was combining two questions:

1) Most important: Is there a way to customize the currently-existing Junit test reporting task?  A quick glance leads me to think that might be possible by configuring a custom [TestReport](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.TestReport.html) appropriately.  Otherwise, I am very reluctant to copying the entire test reporting task internals just to show the failure image diff.

jrodbx (changes requested):

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
jrodbx: let's keep this as is, although I do think output filename formatting is definitely an improvement to make down the road.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: ya unfortunately `TestReport` task doesn't have the configuration we need. I tried this initially and opted for some string appending.

Yep totally down with creating an explicit task class

jrodbx: > Otherwise, I am very reluctant to copying the entire test reporting task internals just to show the failure image diff.
> ya unfortunately TestReport task doesn't have the configuration we need.

What's your take on filing a feature request with Gradle to make this happen?  And is the value prop of copying (and then supporting) this extra task/configuration to expose one failure diff image worthwhile?

jrodbx: > Is there a way to customize the currently-existing Junit test reporting task?

I looked into this some more, and it looks like there is a way!

Here's how I coming to this conclusion:

If I add the following to `sample/build.gradle`:

```
gradle.taskGraph.whenReady {
  gradle.taskGraph.allTasks.each { task ->
    println "task: " + task.name + ", class: " + task.class.name
  }
}

gradle.taskGraph.afterTask { task ->
  println "  inputs: "
  task.inputs.files.each { println "    " + it }

  println "  outputs: "
  task.outputs.files.each { println "    " + it }
}
```

and run:
```
$ gw sample:rPD --rerun
```

I get:
```
task: testDebugUnitTest, class: com.android.build.gradle.tasks.factory.AndroidUnitTest_Decorated

> Task :sample:testDebugUnitTest

ComposeA11yTest > compositeItems STANDARD_ERROR
    Sep 04, 2024 12:53:54 AM app.cash.paparazzi.internal.PaparazziLogger logAndroidFramework
    INFO: Settings [3]: Can't get key animator_duration_scale from content://settings/global

  ...

  inputs:
    /Users/jrod/Development/paparazzi/sample/build/tmp/kotlin-classes/debugUnitTest/app/cash/paparazzi/sample/WidgetTest.class
    /Users/jrod/Development/paparazzi/sample/build/tmp/kotlin-classes/debugUnitTest/app/cash/paparazzi/sample/ResourcesTest$Locale.class
	...
  outputs:
    ...
    /Users/jrod/Development/paparazzi/sample/build/reports/tests/testDebugUnitTest <==== the directory where HTML reports are written
    ...
	
```

From there, setting breakpoints in AGP's `AndroidUnitTest` and Gradle's `Test`, `AbstractTestTask` seem to show where reporting happens at the [end of test execution](https://github.com/gradle/gradle/blob/master/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java#L533), in particular, how a custom `TestReporter` can [be provided](https://github.com/gradle/gradle/blob/master/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java#L598) to generate a [custom report](https://github.com/gradle/gradle/blob/master/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java#L624).

While `AbstractTestTask#setTestReporter` is [package-private](https://github.com/gradle/gradle/blob/master/platforms/software/testing-base/src/main/java/org/gradle/api/tasks/testing/AbstractTestTask.java#L247), it's fairly reasonable to work around that:

```
# PaparazziPlugin: 177

      val testTaskProvider =
        project.tasks.withType(Test::class.java).named { it == "test$testVariantSlug" }
      testTaskProvider.configureEach { test ->
        test.setTestReporter { _, reportDir -> println(reportDir) }

    ....
}

private fun <T : AbstractTestTask> T.setTestReporter(testReporter: TestReporter) {
  AbstractTestTask::class.java
    .getDeclaredMethod("setTestReporter", TestReporter::class.java).apply {
      isAccessible = true
      invoke(this@setTestReporter, testReporter)
    }
}
```

Unless I'm missing something, this should require minimal change to the AGP-copied rendering classes, while removing (most of?) "reportPaparazzi$testVariantSlug", which was duplicating all that glue code anyway.

jrodbx: > Otherwise, I am very reluctant to copying the entire test reporting task internals just to show the failure image diff.

To clarify this a bit, I'm fine with copying custom report rendering bits, but duplicating the test task is less ideal.

geoff-powell: nice find, i was digging through gradle source and the api docs and they didn't expose a public way to do this. I will try this route and defince a custom test reporter and see if that works.

I might still have to use the custom TestReport class potentially but we will see. 🤞🏽

geoff-powell: `To clarify this a bit, I'm fine with copying custom report rendering bits, but duplicating the test task is less ideal.` 

Makes sense, I was able to use the reflection method to hook into the test reporting. Once issue that could come up is that this could effect any consumers doing their own test manipulation. To prevent this, I plan on only overriding the report when `verifyPaparazziDebug` is being run on the task graph.

paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
geoff-powell: done, reverted the change. Turned out I didn't need this change anyway since snapshots are aggregated by test class and method name.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -169,13 +174,30 @@ public class PaparazziPlugin @Inject constructor(
...
+          set(if (it.asFile.exists()) it else null)
+        }
+      }
+
       project.gradle.taskGraph.whenReady { graph ->
         isRecordRun.set(recordTaskProvider.map { graph.hasTask(it) })
         isVerifyRun.set(verifyTaskProvider.map { graph.hasTask(it) })
+
+        // Override test reporter with custom one to include diff screenshots in the report.
+        if (isVerifyRun.get()) {
```
jrodbx, line 188: reminder: we def don't want to add anything else in this callback.

Let's use the same lambda approach as we did in the Gradle caching PR?  If that doesn't work, let's spike some design strategies in this thread!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -115,7 +118,10 @@ public class PaparazziPlugin @Inject constructor(
       val projectDirectory = project.layout.projectDirectory
       val buildDirectory = project.layout.buildDirectory
       val gradleUserHomeDir = project.gradle.gradleUserHomeDir
-      val reportOutputDir = project.extensions.getByType(ReportingExtension::class.java).baseDirectory.dir("paparazzi/${variant.name}")
+
+      val reportingBaseDirectory = project.extensions.getByType(ReportingExtension::class.java).baseDirectory
```
jrodbx, line 122: unused, let's inline back to the following line?

jrodbx (dismissed):

paparazzi/src/test/java/app/cash/paparazzi/helper/TestSnapshotVerifier.kt
```diff
@@ -0,0 +1,33 @@
+package app.cash.paparazzi.helper
+
+import app.cash.paparazzi.Snapshot
+import app.cash.paparazzi.SnapshotHandler
+import app.cash.paparazzi.internal.ImageUtils
+import java.awt.image.BufferedImage
+import java.io.File
+import javax.imageio.ImageIO
+
+internal class TestSnapshotVerifier(val failureDirProvider: () -> File) : SnapshotHandler {
```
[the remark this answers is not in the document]
geoff-powell: there is a reason. Since this is instantiated as a dependency on the Paparazzi test rule, it invokes `temporaryDirectory.newFolder()`. You can't invoke that until `TemporaryDirectory` test rule has started.

I decided to defer the failureDir since it is not needed till later.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -226,6 +236,20 @@ public class PaparazziPlugin @Inject constructor(
 
         test.outputs.dir(reportOutputDir).withPropertyName("paparazzi.report.dir")
 
+        val isVerifying = isVerifyRun.map {
```
BrianGardnerAtl, line 239: It doesn't seem like we need this new field since it's always equal to `isVerifyRun`. Can we just extract the `if` check to set the test reporter, then revert the change on line 261?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/AllTestResults.kt
```diff
@@ -0,0 +1,63 @@
...
+    addVariant(project, flavor, testResult)
+    return testResult
+  }
+
+  fun addTestClass(className: String): ClassTestResults {
+    return addPackageForClass(className).addClass(className)
+  }
+
+  private fun addPackageForClass(className: String): PackageTestResults {
+    val packageName: String
```
BrianGardnerAtl, line 42: Can you just combine the val declaration with setting the value on line 44?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ImagePanelRenderer.kt
```diff
@@ -0,0 +1,61 @@
+package app.cash.paparazzi.gradle.reporting
+
+import org.gradle.reporting.ReportRenderer
+
+internal class ImagePanelRenderer : ReportRenderer<List<ScreenshotDiffImage>, SimpleHtmlWriter>() {
+  override fun render(image: List<ScreenshotDiffImage>, htmlWriter: SimpleHtmlWriter) {
+    // Wrap in a <span>, to work around CSS problem in IE
```
BrianGardnerAtl, line 7: side question, isn't IE dead now? Does Edge have the same rendering issues that need the workaround?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/PackagePageRenderer.kt
```diff
@@ -0,0 +1,71 @@
...
+  override fun renderBreadcrumbs(htmlWriter: SimpleHtmlWriter) {
+    htmlWriter.startElement("div").attribute("class", "breadcrumbs")
+    htmlWriter.startElement("a").attribute("href", "index.html").characters("all").endElement()
+    htmlWriter.characters(String.format(" > %s", results.name))
+    htmlWriter.endElement()
+  }
+
+  @Throws(IOException::class)
+  private fun renderClasses(htmlWriter: SimpleHtmlWriter) {
+    htmlWriter.startElement("table")
```
BrianGardnerAtl, line 20: Much of this writer logic is similar in this file and the `OverviewPageRenderer`. Would it make sense to extract the logic to extension functions on the html writer so it can be reused in multiple places?

This would also help spot differences and make sure they're intended. One case I can see is in the `OverviewPageRenderer`, after the `thead` element has been ended, it adds a `tbody` element but this renderer does not. It is difficult to determine if this is an intended difference or not.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: unfortunately you can't set the test reporter until isVerify is available. Doing it that way breaks config cache. I can add a comment about that

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ImagePanelRenderer.kt
geoff-powell: this is copied over from the actualy impl from gradle. I tend to agree with you but for now think we can keep this in.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/PackagePageRenderer.kt
geoff-powell: ya again i might keep this as is to not differentiate to much from the gradle source

gamepro65 (approved):
Awesome, looks good to me 🙌

eygraber:
I'n not seeing the `TestReport` reporter being used with 1.3.5-SNAPSHOT. There are delta images being generated under `build/paparazzi/failures/` but the HTML under `build/reports/tests/test<variant>UnitTest` doesn't reference them in any way.

I used reflection to read the reporter field and it is null.

geoff-powell:
It will only be used if `verifyPaparazzi[Debug]` task is run. Otherwise it uses the default TestReporter

eygraber:
I ran the verify task.

geoff-powell:
hmm, interesting let me take a look

eygraber:
If it helps the following is printed when running `verifyPaparazziDevDebug` (I have a dev/prod build flavor):

> See the Paparazzi report at: file:///home/eli/workspace/<my project>/build/reports/paparazzi/devDebug/index.html

There is a `build/reports/paparazzi/devDebug` directory, but it is empty. There is typical test result content under `build/reports/tests/testDevDebugUnitTest`.

jrodbx:
Thanks for flagging @eygraber.  Reminder to please file a new issue for tracking, vs commenting on closed PRs and issues.

Tracking in https://github.com/cashapp/paparazzi/issues/1651 and assigning to myself, as this is (hopefully) the last release blocker for 1.3.5.

eygraber:
I don't have a sample project, but I'll try making one.

PR #679, opened 2023-01-09
Fix AndroidX font-family loading with ASM transform on ResourcesCompat

tevjef:
Adapt bytecode transformation done in [Android Studio](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:android/src/com/android/tools/idea/rendering/classloading/ResourcesCompatTransform.kt) to be performed during unit test compilation.

Resolves #488.

PR #1586, opened 2024-09-17
Fix ResourcesCompat.getFont() resource checks to load app compat fonts

### LINKED ISSUE #488: Tests fails when using AndroidX font-family
**Description**
It seems like `ResourcesInterceptor` overrides to use the platform Resources.getFont instead of the AndroidX `ResourcesCompat`. This causes a problem when using a font-family that only specifies `app:` attributes, instead of `android:` attributes.

**Steps to Reproduce**
Use an AppTheme that sets `textAppearance` with a font-family like:
```xml
<font-family xmlns:app="http://schemas.android.com/apk/res-auto">
    <font
        app:font="@font/fk_grotesk_neue_regular"
        app:fontStyle="normal"
        app:fontWeight="400" />

    <font
        app:font="@font/fk_grotesk_neue_medium"
        app:fontStyle="normal"
        app:fontWeight="500" />

    <font
        app:font="@font/fk_grotesk_neue_bold"
        app:fontStyle="normal"
        app:fontWeight="700" />

</font-family>
```

By adding `android:` prefix the tests can inflate everything again.
```xml
<font-family xmlns:app="http://schemas.android.com/apk/res-auto"
xmlns:android="http://schemas.android.com/apk/res/android">
<font
    android:font="@font/fk_grotesk_neue_regular"
    android:fontStyle="normal"
    android:fontWeight="400"
    app:font="@font/fk_grotesk_neue_regular"
    app:fontStyle="normal"
    app:fontWeight="400" />

<font
    android:font="@font/fk_grotesk_neue_medium"
    android:fontStyle="normal"
    android:fontWeight="500"
    app:font="@font/fk_grotesk_neue_medium"
    app:fontStyle="normal"
    app:fontWeight="500" />

<font
    android:font="@font/fk_grotesk_neue_bold"
    android:fontStyle="normal"
    android:fontWeight="700"
    app:font="@font/fk_grotesk_neue_bold"
    app:fontStyle="normal"
    app:fontWeight="700" />

</font-family>
```

**Expected behavior**
It seems like the behaviour was added because of this bug https://issuetracker.google.com/issues/156065472 which now is supposed to be fixed. Could the Interception be removed?

**Additional information:**
- Paparazzi Version: 1.0.0
- OS: MacOS 12.3.1 
- Compile SDK: 32
- Gradle Version: 7.4.1
- Android Gradle Plugin Version: 7.2.1

**Screenshots**
-

geoff-powell:
Fixes #488.  Closes #679.


paparazzi/src/main/java/app/cash/paparazzi/internal/resources/ResourceFolderRepository.kt
```diff
@@ -40,7 +40,7 @@ internal class ResourceFolderRepository(
    * Common prefix of paths of all file resources.  Used to compose resource paths returned by
    * the [BasicFileResourceItem.getSource] method.
    */
-  private val resourcePathPrefix: String = "${resourceDir.path}/"
+  private val resourcePathPrefix: String = "${resourceDir.path}${File.separator}"
```
geoff-powell, line 43: this, was causing a `FileNotFoundException` on Windows. Needed to respect the correct file separator on the platform

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatTransform.kt
```diff
@@ -0,0 +1,75 @@
...
+      isInterface: Boolean
+    ) {
+      return if ("startsWith" == name) {
+        super.visitMethodInsn(opcode, owner, "contains", "(Ljava/lang/CharSequence;)Z", isInterface)
+      } else {
+        super.visitMethodInsn(opcode, owner, name, descriptor, isInterface)
+      }
+    }
+
+    override fun visitLdcInsn(value: Any?) {
```
BrianGardnerAtl, line 60: I noticed this function isn't coming from the ASM fix. What is this solving that wasn't covered by `visitMethodInsn`?

BrianGardnerAtl (commented):
One question about the `LoadFontVisitor` implementation but I think it looks good. I love seeing an interceptor go

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatTransform.kt
geoff-powell: This is needed due to platform differences in how ResourcesCompat loads resources. For Windows the path used in source code should be `res\`. Since android is Linux based, the path is `res/`

BrianGardnerAtl: Ok, I was assuming the ASM fix would run on both windows and linux but it sounds like it's just running internally on Android which is why it only uses the linux based path?

geoff-powell: Yes sorry for the confusion. ASM does run on Windows and Linux. `ResourcesCompat` comes from app compat dependency that doesn't support multi-plat and is only for Android since its an AAR. Hence the folder check looks for `res/` and needs to be changed to `res\` in the bytecode.

BrianGardnerAtl (approved):

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/ResourceFolderRepository.kt
jrodbx: Actually, Android Studio does fine with "/": https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:android/src/com/android/tools/idea/res/ResourceFolderRepository.java;l=275?q=ResourceFolderRepository

The ResourceRepository mechanism is separate from the ASM compat stuff they load.  Let's investigate where that change is being made.

paparazzi/src/test/java/app/cash/paparazzi/internal/resources/PseudolocaleGeneratorTest.kt
```diff
@@ -163,7 +164,7 @@ class PseudolocaleGeneratorTest {
         assertThat(size).isEqualTo(3)
         assertThat(getValue("test_color").value).isEqualTo("#ffffffff")
         assertThat(getValue("test_color_2").value).isEqualTo("#00000000")
-        assertThat(getValue("color_selector").value).isEqualTo("$resFolderRoot/color/color_selector.xml")
+        assertThat(getValue("color_selector").value).isEqualTo("$resFolderRoot${File.separator}color/color_selector.xml")
```
jrodbx, line 167: similar to above, let's revert this and figure out why this is failing

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatTransform.kt
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * [ClassVisitor] that fixes a hardcoded path in ResourcesCompat.loadFont.
+ * In this method, there is a check that font files have a path that starts with res/.
+ * This is not the case in Studio. This replaces the check with one that verifies that the path contains res/.
+ *
+ * Inspired by ASM fix:
+ * https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/classloading/ResourcesCompatTransform.kt;drc=5bb41b6d5e519c891a4cd6149234138faa28e1af
+ */
+internal class ResourcesCompatTransform(api: Int, delegate: ClassVisitor) : ClassVisitor(api, delegate) {
```
jrodbx, line 15: what's different between this and https://github.com/cashapp/paparazzi/pull/679?  Would be nice to see if that version can be merged in.

jrodbx (changes requested):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatTransform.kt
geoff-powell: Its the same just without the platform plumbing that was implemented. I didn't want to push to that branch so I created this one.

paparazzi/src/main/java/app/cash/paparazzi/internal/resources/ResourceFolderRepository.kt
geoff-powell: Ya I think the issue stems from our internal resource code we pulled from Android Studio code. I can dig into why folder structure changes for windows

geoff-powell: So, once I set up Parallels on my Mac, I was able to debug this and it is now fixed! We don't need this change after all!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatTransform.kt
```diff
@@ -0,0 +1,66 @@
...
+  private class LoadFontVisitor(api: Int, delegate: MethodVisitor) : MethodVisitor(api, delegate) {
+
+    override fun visitMethodInsn(
+      opcode: Int,
+      owner: String,
+      name: String,
+      descriptor: String,
+      isInterface: Boolean
+    ) {
+      return if ("startsWith" == name) {
```
jrodbx, line 52: nit: expression function

jrodbx: actually, this method returns Unit, so an alternative might be to remove the explicit return?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatVisitorFactory.kt
```diff
@@ -0,0 +1,18 @@
...
+internal abstract class ResourcesCompatVisitorFactory : AsmClassVisitorFactory<InstrumentationParameters.None> {
+  override fun createClassVisitor(
+    classContext: ClassContext,
+    nextClassVisitor: ClassVisitor
+  ): ClassVisitor {
+    return ResourcesCompatTransform(instrumentationContext.apiVersion.get(), nextClassVisitor)
+  }
+
+  override fun isInstrumentable(classData: ClassData): Boolean = true
+}
```
jrodbx, line 18: nit: I rather like nesting the Transform class inside the factory class as done here: https://github.com/cashapp/paparazzi/pull/679/files#diff-b388a8009e690deb676684efd348e96131578944fae01677f60b93e921698a39

Shall we follow that pattern?  (keeping the name `ResourcesCompatTransform`)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -121,6 +124,15 @@ public class PaparazziPlugin @Inject constructor(
...
 
+      val testInstrumentation = testVariant.instrumentation
+      testInstrumentation.transformClassesWith(
+        ResourcesCompatVisitorFactory::class.java,
+        InstrumentationScope.ALL
+      ) { }
+      testInstrumentation.setAsmFramesComputationMode(
+        FramesComputationMode.COMPUTE_FRAMES_FOR_INSTRUMENTED_METHODS
+      )
+
```
jrodbx, line 135: nit: this is OK in the short term to support AndroidX fonts, but this adds more layers to difficulty to:
* any hopes of a Bazel plugin
* SDKifying

I'm less concerned about the first bullet point (although I try my best to be mindful of it in design decisions like this).  The second one makes me more sad, cause we know we have to tackle this soon-ish anyway.

In a follow-up, we should follow in Android Studio's footsteps and move this asm stuff into some classloader shim.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatVisitorFactory.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.android.build.api.instrumentation.ClassData
+import com.android.build.api.instrumentation.InstrumentationParameters
+import org.objectweb.asm.ClassVisitor
+
+internal abstract class ResourcesCompatVisitorFactory : AsmClassVisitorFactory<InstrumentationParameters.None> {
+  override fun createClassVisitor(
+    classContext: ClassContext,
+    nextClassVisitor: ClassVisitor
+  ): ClassVisitor {
+    return ResourcesCompatTransform(instrumentationContext.apiVersion.get(), nextClassVisitor)
```
jrodbx, line 14: question: why use the apiVersion vs Opcodes.ASM9?  I kinda like what you did here, but could upgrades to the underlying libs break us in the future?

geoff-powell: ya, I noticed that forcing the version caused issues. I saw this usage in the other pr and thought that might be a better way to go.

Looks like `Opcodes.ASM9` works

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: > I'm less concerned about the first bullet point (although I try my best to be mindful of it in design decisions like this). The second one is more top of mind for me, cause we know we'll have to tackle this soon-ish anyway.
> 
> In a follow-up, we should follow in Android Studio's footsteps and move this asm stuff into some classloader shim.

Yep 100% that relying on gradle more isn't the final way to do this. I think we can look into the ASM class loading that way we can remove the reliance on ByteBuddy shims and implement these asm transforms at runtime.

jrodbx: Yep, that's precisely been the goal!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -126,7 +126,7 @@ public class PaparazziPlugin @Inject constructor(
 
       val testInstrumentation = testVariant.instrumentation
       testInstrumentation.transformClassesWith(
-        ResourcesCompatVisitorFactory::class.java,
+        ResourcesCompatTransform::class.java,
```
jrodbx, line 129: oh, sorry for the confusion.  This should remain `ResourcesCompatVisitorFactory` and the transform should remain `ResourcesCompatTransform`.

DSteve595:
This didn't fix it for our project and I found out that it's because our font resources are in a dir called `resources/`, not `res/`, which was just a casual decision due to the project being multiplatform and not Android-only. Renaming our directory to `res/` fixes the issue.. but it feels really opaque that that would be required.

Is there any chance this whole `startsWith`/`contains` check could be skipped entirely?

PR #1606, opened 2024-09-26
Add Compose support for InstantAnimationsRule

### LINKED ISSUE #901: `InstantAnimationsRule` doesn't work with Compose
`InstantAnimationsRule` doesn't work with compose.

I have some components that have animations and it would be great to be able to take a screenshot of the last frame.

This use case could be fullfilled with #627 or #752. Anyway it would be great than instead of hardcode an `offset` you could just say "take a screenshot of the last frame".

geoff-powell:
Fixes #901

Animation without rule:
app cash paparazzi plugin test_AnimationRuleTest_test


paparazzi/src/main/java/app/cash/paparazzi/InstantAnimationsRule.kt
```diff
@@ -45,6 +46,7 @@ public class InstantAnimationsRule : TestRule {
     }
     set(value) {
       setDurationScale.invoke(null, value)
+      PaparazziMotionDurationScale.animationScaleFactor = value
```
jrodbx, line 49: If I'm understanding this correctly, a non-Compose project applying this rule would result in a ClassNotFoundException, right?

paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziMotionDurationScale.kt
```diff
@@ -0,0 +1,10 @@
+package app.cash.paparazzi.internal
+
+import androidx.compose.ui.MotionDurationScale
+
+internal object PaparazziMotionDurationScale : MotionDurationScale {
```
jrodbx, line 5: nit: 

```suggestion
internal object InstantDurationScale : MotionDurationScale {
```

jrodbx (commented):
Btw, this is another example of allowing to skip frames. 😄 

For context, the [pending recompositions PR](https://github.com/cashapp/paparazzi/pull/1537) allows for an edge case where Composables require a second measure/layout pass.  Now, in an ideal world, neither Android Studio nor we would allow that, but since some androidx Composables require this, for things like `Text.onTextLayout` and `Modifier.onGloballyPositioned`, we decided to support it if for no other reason than to avoid the "why does Android Studio do this, but not Paparazzi"?  

As a result, after some iterations, we landed on only allowing a second frame and not more.  Why?  Imagine an implementation of a Composable that takes several frames to render.  Paparazzi today will snapshot a blank canvas, which is a signal that the UI was ill-designed!  Allowing the ability to skip to the end would circumvent that superpower. 😢 

See https://github.com/cashapp/paparazzi/pull/1205#discussion_r1496432901 for a similar discussion.

Also, is it possible that the InstantAnimationsRule we have today for legacy views, was a workaround that can now be deleted, given that we since have gif support + pending recompositions support?

In summary:

- If frames = 1, snapshot static image
- If frames = 2, grrr, but ok, we allow via pending recompositions, cuz androidx Composable kinda ~misbehave~ require another pass?
- if frames > 2, either use gif support OR you have a bug, fix yo' UI widget.

Is this a reasonable stance?  I'm open to pivoting, but my current thinking remains: how can we allow for such a use case without compromising the framework's current superpower to signal when a UI is taking too many frames to render?

cc: @queencodemonkey who shared thoughts re: this @ Droidcon NYC last week.

queencodemonkey:
Because I was summoned… ;)

> In summary:
> If frames = 1, snapshot static image
> If frames = 2, grrr, but ok, we allow via pending recompositions cuz androidx Composable kinda misbehave require another pass?
> if frames > 2, either use gif support OR you have a bug, fix yo' UI widget.

IMO, this is reasonable and allows for the leniency in the API and in Studio via the `onSizeChanged`/`onGloballyPositioned` type APIs, as John mentioned above. Honestly, I would prefer not to have the 2-frame allowance because these APIs in the context of Compose rendering tend to be misuses rather than a "no better option" in situations where one could use `BoxWithConstraints` OR a custom layout instead.

In fact, there is a section about this in the docs: [Recomposition Loop (cyclic phase dependency)](https://developer.android.com/develop/ui/compose/phases#recomp-loop). It's inefficient, generally unnecessary, and breaks principles around letting a Compose parent layout be the single source of truth regarding the layout of children.

> The general principle here is to have a single source of truth for multiple UI elements that should be measured and placed with regards to one another. Using a proper layout primitive or creating a custom layout means that the minimal shared parent serves as the source of truth that can coordinate the relation between multiple elements. Introducing a dynamic state breaks this principle.

Because GIF support is available for the case of animations and because I think the aforementioned superpower of signaling inefficient/suboptimal UI implementation legitimately brings specific and concrete awareness to the above practices, I am all for the approach/philosophy John outlined.

paparazzi/src/main/java/app/cash/paparazzi/InstantAnimationsRule.kt
geoff-powell: possibly, should add tests to ensure this.

geoff-powell:
Yep, I agree on this stance. There maybe room from an api that allows us to await frames until the recomposer is idle. Effectively what we have in Cash where we can snapshot the last frame generated.

This api has been in discussion for support for a while so that was my thinking to unblock this for paparazzi before our 2.0 rewrite happens. Will close this for now and think on it a bit more.

Thanks for the feedback @jrodbx @queencodemonkey!

PR #1608, opened 2024-09-27
Update AllTestResults.kt

GeorgCantor:
**Use of `substringBeforeLast`**: This function simplifies the extraction of the package name by directly getting the substring before the last occurrence of `"."`. If there is no `"."`, it returns an empty string, which eliminates the need for a conditional check.


geoff-powell (approved):
Thanks, I konverted this from java and was like there should be a better way to do this lol

PR #1620, opened 2024-10-07
Hash dimensions instead of pixel coordinates

jrodbx:
Revisited https://github.com/cashapp/paparazzi/issues/1527 where the OP kindly reminded me ordering matters when adding bytes to a sha1 computation.

- https://github.com/cashapp/paparazzi/issues/1527#issuecomment-2279741497
- https://github.com/cashapp/paparazzi/issues/1527#issuecomment-2311356219

A quick recipe highlights this:

```
fun main() {
  println("before: " + hash(listOf(0, 16777215)))
  println("after: " + hash(listOf(16777215, 0)))
}

fun hash(input: List<Int>): String {
  val hashingSink = HashingSink.sha1(blackholeSink())
  hashingSink.buffer().use { sink ->
    input.forEach(sink::writeInt)
  }
  return hashingSink.hash.hex()
}
```
```
before: b553455ddc0907c93bca95b9c2127ed14b0a4549
after: ff51f8c7a18c33064eb6a52217df7466705f0492
```
As a result, we can speed up the O(3N) algorithm a bit into an O(N+1) by hashing the image dimensions instead.  My bad for not understanding this earlier!

In addition, I've opportunistically converted the unit test into a plugin test since pixel comparison is already part of the snapshot verification and I anticipate potential refactoring of HtmlReportWriter so this potentially sidesteps any future pain in doing so.


geoff-powell:
> In addition, I've opportunistically converted the unit test into a plugin test since pixel comparison is already part of the snapshot verification and I anticipate potential refactoring of HtmlReportWriter so this potentially sidesteps any future pain in doing so.

Nice, was thinking this would be good when working on the snapshot gradle task inputs/outputs. Great work!!!

geoff-powell (approved):

PR #1627, opened 2024-10-09
Set delta image's similar pixels to Blue

### LINKED ISSUE #1612: Differentiate similar and different pixels in delta image
The `OffByTwo` differ is currently drawing [similar pixels](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/internal/OffByTwo.kt#L67) with the same rgb value as [different pixels](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/internal/OffByTwo.kt#L72). This makes it hard to know which differences lead to a failure and which don't and can lead to developer confusion.

benwicks:
...to differentiate them from the different pixels

Resolves #1612

I added a plugin test (`verifySimilar`) demonstrating that similar pixels are expected to be blue 🟦.

| Before                                                | After                                                   |
|-----------------------------------|------------------------------------|
|  |  |


sample/src/test/java/app/cash/paparazzi/sample/SimilarDeltaTest.kt
```diff
@@ -0,0 +1,62 @@
...
+) {
+  Column(
+    Modifier
+      .background(Color.White)
+      .fillMaxSize()
+      .wrapContentSize()
+  ) {
+    Box(
+      Modifier
+        .background(if (shouldUseOriginalBackgroundColor) Color.Yellow else Color(0xFFFFFD00))
```
benwicks, line 46: `0xFFFFFD00` triggers the similar pixel check as its green value is 2 less than yellow.

jrodbx: rather than a test in sample, i'd move this to a plugin test.  I'd suggest:

* copying `verify-mode-failure` to something like `verify-similar`
* copy/paste `PaparazziPluginTest.verifyFailure` to something like `PaparazziPluginTest.verifySimilar`
  * update the fixtureRoot accordingly
  * temporarily, update the argument from `verifyPaparazziDebug` to `recordPaparazziDebug`
  * you might have to make other changes, depending on whether you want to keep the test class and method names as is, or replace with yours.
* replace VerifyTest with your test here
  * might have to add compose to build.gradle, see other projects for examples
* remove the conditional logic from this test, making it the "before" case
* run the test project and copy the image from `build/reports/paparazzi/debug/images` to `src/test/snapshots/images`.  
  * This is your "golden" file.
* update the argument back to `verifyPaparazziDebug`
* modify this test to exhibit the "after" case.
* run the test project again (it should fail) and copy the delta file in `build/paparazzi/failures` to `src/test/resources`.
* run the test project again (it should pass)

paparazzi/src/main/java/app/cash/paparazzi/internal/OffByTwo.kt
```diff
@@ -64,7 +64,7 @@ internal object OffByTwo : Differ {
 
         if (abs(deltaR) <= 2 && abs(deltaG) <= 2 && abs(deltaB) <= 2) {
           similarPixels++
-          deltaImage.setRGB(expectedWidth + x, y, newRGB)
+          deltaImage.setRGB(expectedWidth + x, y, actualRgb)
```
jrodbx, line 67: let's go with blue?  something that stand outs but doesn't signal red/error.

jrodbx (changes requested):

paparazzi/src/main/java/app/cash/paparazzi/internal/OffByTwo.kt
```diff
@@ -64,7 +64,7 @@ internal object OffByTwo : Differ {
 
         if (abs(deltaR) <= 2 && abs(deltaG) <= 2 && abs(deltaB) <= 2) {
           similarPixels++
-          deltaImage.setRGB(expectedWidth + x, y, newRGB)
+          deltaImage.setRGB(expectedWidth + x, y, 0xFF0000FF.toInt())
```
benwicks, line 67: Similar pixels will be blue 🟦 in the delta

jrodbx (approved):

PR #1636, opened 2024-10-15
Upgrade AGP to 8.7.1

geoff-powell:
- [x] Wait till 1.3.5 is released


jrodbx:
Remember that AGP, sdk-common, layoutlib-api, etc. are correlated.  Also, this is quite a large bump.  Is there a reason to bump this now, i.e., is this blocking anything?  Otherwise, I think 1.3.5 should stay on 8.4.2.

geoff-powell:
> Remember that AGP, sdk-common, layoutlib-api, etc. are correlated. Also, this is quite a large bump. Is there a reason to bump this now, i.e., is this blocking anything? Otherwise, I think 1.3.5 should stay on 8.4.2.

Yep had the same thought. I was thinking this could land after 1.3.5. I will label it as such

jrodbx (approved):

Issue #1577, opened 2024-09-10
Material 3 `ModalBottomSheet` does not display Accessibility


Jonas-I:
**Description**
When using the [`ModalBottomSheet`](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#ModalBottomSheet(kotlin.Function0,androidx.compose.ui.Modifier,androidx.compose.material3.SheetState,androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Shape,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Color,kotlin.Function0,kotlin.Function0,androidx.compose.material3.ModalBottomSheetProperties,kotlin.Function1)) from Material3, the accessibility for the components inside the Bottom Sheet is not rendered. The Bottom Sheet also takes up the width of the entire image rather than width of the device.

It is possible that this issue is due to the component being a modal which overlays the main UI instead of being a part of the screen layout. Using a regular [`BottomSheetScaffold`](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#BottomSheetScaffold(kotlin.Function1,androidx.compose.ui.Modifier,androidx.compose.material3.BottomSheetScaffoldState,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Shape,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,kotlin.Function0,kotlin.Boolean,kotlin.Function0,kotlin.Function1,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,kotlin.Function1)) would output the expected result.

**Steps to Reproduce**
To display the BottomSheets, we have to set the `sheetState` to have an `initialValue` of `Expanded`.

For the `ModalBottomSheet`, we can use this test:
```kotlin
class ModalBottomSheetTest {

    @get:Rule
    val paparazzi = Paparazzi(renderExtensions = setOf(AccessibilityRenderExtension()))

    @Test
    @OptIn(ExperimentalMaterial3Api::class)
    fun modalBottomSheetTest() {
        paparazzi.snapshot {
            ModalBottomSheet(
                onDismissRequest = {},
                sheetState = SheetState(
                    skipPartiallyExpanded = true,
[...]

PR #1580, opened 2024-09-12
Add support for dialogs and sheets for accessibility render extension

geoff-powell:
#1577

PR #1637, opened 2024-10-15
Fix background color covering content and bottom sheet / dialogs rendering for accessibility

### LINKED ISSUE #1634: Fix window manager view rendering Material3 Compose Bottom Sheet over AccessibilityRenderExtension info.
**Description**
When bumping LayoutLib dependency to 14.0.11+, Material3 bottom sheets render overtop the render extension metadata. This issue is to investigate a possible fix. Ignoring the test `modalBottomSheetMaterial3` in this [LayoutLib bump PR](https://github.com/cashapp/paparazzi/pull/1578) till this issue is resolved.
Image

**Steps to Reproduce**
- Remove @Ignore in `modalBottomSheetMaterial3()` in compose-a11y test project test foile.
- Run `composeA11y` test in PaparazziGradlePluginTest.kt.

**Expected behavior**
Snapshot is unchanged.

**Additional information:**
- Paparazzi Version: 1.3.5 ([renovate/layoutlib](https://github.com/cashapp/paparazzi/pull/1633/commits/728a77a653da5ae9f33062b8d23b3b7455d7c5e7))
- OS: Mac OS 
- Compile SDK: 34
- Gradle Version: 8.7.1
- Android Gradle Plugin Version: 8.4.2

**Screenshots**
See above

geoff-powell:
Closes #1634.  Related to #1577 and #1414.


paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
```diff
@@ -13,4 +14,12 @@ import android.widget.FrameLayout
...
   context: Context,
   attrs: AttributeSet
-) : FrameLayout(context, attrs)
+) : FrameLayout(context, attrs) {
+  init {
+    /**
+     * Needed as [android.view.WindowManagerImpl] uses the view root background color to set the WindowManagerImpl view's background color.
+     * If we set this as transparent, the WindowManagerImpl view will be transparent as well and correctly renders the window above content.
+     */
+    setBackgroundColor(Color.TRANSPARENT)
```
geoff-powell, line 23: Change is needed due to a view ordering change in WindowManagerImpl in 14.0.11.

(line 105) https://www.diffchecker.com/fHBDkOTq/
AS diff: https://android.googlesource.com/platform/frameworks/layoutlib/+/f86a9059e4e0fbc6908c32aa88bddc640de86831%5E%21/bridge/src/android/view/WindowManagerImpl.java

BrianGardnerAtl (commented):
Nice! I noticed in the updated snapshot it seems like the whole screen is tinted pink for the "close sheet" accessibility label, and the highlights around the drag handle and text 2 seem to be missing. Either that or it's just being colored over by the close sheet highlight. Is this something we're fine with for now and can fix later or should we try to get that resolved in this PR?

geoff-powell:
> Nice! I noticed in the updated snapshot it seems like the whole screen is tinted pink for the "close sheet" accessibility label, and the highlights around the drag handle and text 2 seem to be missing. Either that or it's just being colored over by the close sheet highlight. Is this something we're fine with for now and can fix later or should we try to get that resolved in this PR?

Nice find!! this is most likely due to the same change of WindowManager index in the parent (boxes in overlay are rendered behind content)

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -39,11 +41,22 @@ public class AccessibilityRenderExtension : RenderExtension {
   override fun renderView(
     contentView: View
   ): View {
+    // Window Manager needed to access accessibility elements for views that draw to other windows.
```
jrodbx, line 44:
```suggestion
    // WindowManager needed to access accessibility elements for views that draw to other windows.
```

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -39,11 +41,22 @@ public class AccessibilityRenderExtension : RenderExtension {
...
+    val windowManager = contentView.context.getSystemService(WindowManager::class.java)
+
     return LinearLayout(contentView.context).apply {
       orientation = LinearLayout.HORIZONTAL
       weightSum = 2f
       layoutParams = ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)
 
+      viewTreeObserver.addOnGlobalLayoutListener {
+        // Window manager is rendered at the root of the view hierarchy, rendering it full width.
```
jrodbx, line 53:
```suggestion
        // The root of the view hierarchy is rendered at full width.
```

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -39,11 +41,22 @@ public class AccessibilityRenderExtension : RenderExtension {
...
+        // We need to restrict it when taking accessibility snapshots.
+        (windowManager as WindowManagerImpl).currentRootView?.let {
+          layoutParams = FrameLayout.LayoutParams(contentView.measuredWidth, MATCH_PARENT, Gravity.START)
```
jrodbx, line 56: interesting, is this how layoutlib does it?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -70,6 +81,7 @@ public class AccessibilityRenderExtension : RenderExtension {
     }
   }
 
+  @SuppressLint("VisibleForTests")
```
jrodbx, line 84: nit: revert

You may recall this is because we should be moving away from [ViewRootForTest](https://github.com/cashapp/paparazzi/pull/1580#discussion_r1759282744)

paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
jrodbx: The timestamp of that change is dated 2021, so I am skeptic of this being the root cause since we have been on later layoutlib versions.  What am I missing?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
jrodbx: question: will this layout change affect subsequent non-accessibility tests, i.e., do we need to revert back to full width in some sort of shutdown?

paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
jrodbx: update: I had only clicked on the AS diff per my comment above.  I just viewed the diffcheck link and that makes more sense.  Is it possible you posted the wrong AS diff?

geoff-powell:
> The timestamp of that change is dated 2021, so I am skeptic of this being the root cause since we have been on later layoutlib versions. What am I missing?

Unfortunately, the git history doesn't add up with the change they made. I don't see the WindowManagerImpl change in the file git history but I am looking at LL `14.0.9` `WindowManagerImpl` I can see at line 105 the add view passed a specific index.
Screenshot 2024-10-16 at 2 04 10 PM

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
geoff-powell: > interesting, is this how layoutlib does it?
No this is specifically to get AccessibilityRenderExtension to render as expected.

> question: will this layout change affect subsequent non-accessibility tests, i.e., do we need to revert back to full width in some sort of shutdown?

Hmm, we can I guess. I hadn't thought about that. Whats the use case where this is needed?

paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
geoff-powell: yes the one I posted isn't correct but according to the cs.android.com that is the last revision with that line change.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
jrodbx: Oh, I don't know if this is *needed* per se, thus the question.  

I'm thinking of scenarios where you run two tests, an accessibility test and a non-accessibility test, in that order, and whether the second test may render incorrectly.

Actually, since this is mutating the currentRootView, which is presumably detached/re-attached to between tests, then this shouldn't be a problem?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityOverlayDrawable.kt
```diff
@@ -0,0 +1,41 @@
+package app.cash.paparazzi.accessibility
+
+import android.graphics.Canvas
+import android.graphics.ColorFilter
+import android.graphics.Paint
+import android.graphics.drawable.Drawable
+import app.cash.paparazzi.accessibility.RenderSettings.toColorInt
+
+internal class AccessibilityOverlayDrawable : Drawable() {
```
geoff-powell, line 9: Ended up making this a foreground drawable instead of a view so it can be drawn overtop WindowManagerImpl

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityOverlayDrawable.kt
```diff
@@ -0,0 +1,41 @@
...
+      strokePaint.color = it.color.toColorInt()
+      strokePaint.alpha = RenderSettings.DEFAULT_RENDER_ALPHA * 2
+      canvas.drawRect(it.displayBounds, strokePaint)
+    }
+  }
+  override fun setAlpha(alpha: Int) = Unit
+  override fun setColorFilter(p0: ColorFilter) = Unit
+
+  @Deprecated("Not used", ReplaceWith("255"))
+  override fun getOpacity(): Int = 255
```
jrodbx, line 40: nit: PixelFormat.OPAQUE?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityOverlayDrawable.kt
```diff
@@ -0,0 +1,41 @@
...
+      paint.color = it.color.toColorInt()
+
+      canvas.drawRect(it.displayBounds, paint)
+
+      strokePaint.color = it.color.toColorInt()
+      strokePaint.alpha = RenderSettings.DEFAULT_RENDER_ALPHA * 2
+      canvas.drawRect(it.displayBounds, strokePaint)
+    }
+  }
+  override fun setAlpha(alpha: Int) = Unit
```
jrodbx, line 36:
```suggestion

  override fun setAlpha(alpha: Int) = Unit
```

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityOverlayDetailsView.kt
```diff
@@ -51,8 +51,8 @@ internal class AccessibilityOverlayDetailsView(context: Context) : FrameLayout(c
     setBackgroundColor(RenderSettings.DEFAULT_DESCRIPTION_BACKGROUND_COLOR.toColorInt())
   }
 
-  fun addElements(elements: Collection<AccessibilityElement>) {
-    accessibilityElements.addAll(elements)
+  fun updateElements(elements: Collection<AccessibilityElement>) {
+    accessibilityElements = elements
```
jrodbx, line 55: keep `addAll` and clear on each add, if you're looking to avoid leaking memory?  this allows the list to remain a val.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityOverlayDrawable.kt
jrodbx: cool

paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/test/java/app/cash/paparazzi/plugin/test/ComposeA11yTest.kt
```diff
@@ -52,7 +51,7 @@ class ComposeA11yTest {
       ) {
         Text(text = "Text 2")
       }
-      Text(text = "Text 1")
+      Text(modifier = Modifier.wrapContentSize(), text = "Text 1")
```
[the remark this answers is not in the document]
jrodbx: nice

paparazzi/src/main/java/app/cash/paparazzi/internal/ComposeViewAdapter.kt
jrodbx: Here's the history for WindowManagerImpl: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/layoutlib/bridge/src/android/view/WindowManagerImpl.java;bpv=1

in case there is something we can learn from their changes!

paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/test/snapshots/images/app.cash.paparazzi.plugin.test_ComposeA11yTest_modalBottomSheetMaterial3.png
jrodbx, line 1: is it expected that the text centering is now different in the background view?

geoff-powell: Ya I moved that to better capture the text bounds and the `close` tap area of the dialog. Before they were overlapping completely

jrodbx (approved):

PR #1645, opened 2024-10-18
Proposed fix to better time Compose UI animations

geoff-powell:
So animations with `offsetMillis` didn't work properly in paparazzi. This was due to the Transition apis tracking the `startTimeNanos` when a transition starts. Since this is set ahead of the transition, the delta to render the frame was always 0. For `gif()` methods that used `start = 0` they render correctly because the starting frame is at `0` nanos.

The best way I could combat this was for cases in Compose where the caller was manually setting the `startNanos` for `Paparazzi.takeSnapshots()` I call multiple render calls with time 0 to ensure Transition apis have `startTimeNanoes = 0`


BrianGardnerAtl (commented):
I think this looks good. Having to render twice with a 0 start time is a little quirky but the comment explains it well

BrianGardnerAtl (approved):

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -318,16 +318,30 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+        withTime(0) {
+          renderSession.render(false)
+        }
+      }
+
       for (frame in 0 until frameCount) {
         val nowNanos = (startNanos + (frame * 1_000_000_000.0 / fps)).toLong()
 
         // If we have pendingTasks run recomposer to ensure we get the correct frame.
         var hasPendingWork = false
```
harrytmthy, line 385: Hi @geoff-powell, spotted a subtle bug here. `hasPendingWork` is a computed property backed by synchronized state in Recomposer, so storing its value in a local var like this might introduce an issue if the underlying state changes between reads (which is likely during animation rendering).

Would it make sense to unwrap `recomposer` once, then query `recomposerInstance.hasPendingWork` at the moment it's needed, instead of snapshotting into a local var?

That way the logic stays aligned with Recomposer's live state, like:

```kotlin
if (frame == 0 && recomposerInstance.hasPendingWork) {
    withTime(nowNanos) {
        renderForResult()
    }
    if (recomposerInstance.hasPendingWork) {
        logger.warning("...")
    }
}
```

PR #1650, opened 2024-10-22
Improve accessibility render extension property support for legacy views

colinmarsch:
Adds support for more accessibility properties on legacy Android views to more closely match the support that Compose has in the `AccessibilityRenderExtension`.


geoff-powell (approved):
NIce!

PR #1686, opened 2024-11-11
Bump gradle version to 8.11

geoff-powell:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.2-bin.zip
```
jrodbx, line 3: this is already on master

gradle.properties
```diff
@@ -28,3 +28,6 @@ android.suppressUnsupportedCompileSdk=UpsideDownCakePrivacySandbox
 
 # Signals to our own plugin that we are building within the repo.
 app.cash.paparazzi.internal=true
+
+# Gradle
+org.gradle.configuration-cache.parallel=true
```
jrodbx, line 33: doesn't apply to CI clean builds

TWiStErRob: But it would do for any contributor cloning the repository? There are a lot of modules, because of test projects. Wouldn't it be beneficial to share the win with everyone and configure the CI to be the exception?

(I'm curious in general, about the approach you take wrt sharing config that helps developers)

geoff-powell: Yep my thought was it would be an improvement for contributors. I think your suggestion of adding a gradle flag to not use config cache makes sense if we enable this.

Issue #1693, opened 2024-11-15
DropdownMenu not appearing in snapshots


corentin-stamper:
**Description**
`DropdownMenu` not appearing in snapshots

**Steps to Reproduce**
```kotlin
package paparazzi.dropdown_menu

import androidx.compose.foundation.background
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.Column
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.layout.padding
import androidx.compose.material3.DropdownMenu
import androidx.compose.material3.DropdownMenuItem
import androidx.compose.material3.MaterialTheme
import androidx.compose.material3.Text
import androidx.compose.ui.Modifier
import androidx.compose.ui.graphics.Color
import androidx.compose.ui.unit.dp
import app.cash.paparazzi.Paparazzi
import com.google.testing.junit.testparameterinjector.TestParameterInjector
import org.junit.Rule
import org.junit.Test
import org.junit.runner.RunWith

@RunWith(TestParameterInjector::class)
class DropdownMenuTest {

    @get:Rule
    val paparazzi = Paparazzi()

    @Test
    fun test() {
        paparazzi.snapshot {
            MaterialTheme {
                Column(
                    modifier = Modifier
                        .fillMaxSize()
                        .background(Color.White)
                        .padding(8.dp)
                ) {
                    Box {
                        Text(
                            text = "Label 1",
                        )

                        DropdownMenu(
                            expanded = true,
                            onDismissRequest = {},
                        ) {
                            DropdownMenuItem(
                                text = {
                                    Text(
                                        text = "Label 1",
                                    )
                                },
                                onClick = {}
                            )
                            DropdownMenuItem(
                                text = {
                                    Text(
                                        text = "Label 2",
                                    )
                                },
                                onClick = {}
                            )
                            DropdownMenuItem(
                                text = {
                                    Text(
                                        text = "Label 3",
                                    )
                                },
                                onClick = {}
                            )
                        }
                    }
                }
            }
        }
    }
}
```

**Expected behavior**
The `DropdownMenu` appears

**Additional information:**
- Paparazzi Version: 1.3.5
- OS:
- Compile SDK: 35
- Gradle Version: 8.9
- Android Gradle Plugin Version: 8.7.2

**Screenshots**

![Image](https://github.com/user-attachments/assets/65be7cfa-f409-47ff-8fbd-d5b98e0fc4b9)

[...]

Issue #678, opened 2023-01-09
Compose delay for animation effect


ButovAndersen:
Hi. 
I have compose widget with animation effect.
This animation has a duration about 500ms.
How can I add delay 500ms before the paparazzi takes a screenshot?

ButovAndersen:
@jisungbin I expected this to work but it doesn't :(

leo-beliakov:
@jrodbx Any updates on this?

PR #1694, opened 2024-11-15
Fix accessibility overlay drawing in root view instead of window manager

geoff-powell:
When debugging #1693 issue I noticed the WindowManagerImpl view can have different sizes based on the window rendered. This caused issues with `AccessibilityRenderExtension` where overlay foreground drawable wasn't rendering in the correct place. 

The fix is to render the content overtop the `ComposeViewAdapter` which is the root view that gets rendered by paparazzi instead.


jrodbx (approved):

paparazzi-gradle-plugin/src/test/projects/compose-a11y/src/test/snapshots/images/app.cash.paparazzi.plugin.test_ComposeA11yTest_dropDownMaterial3.png
geoff-powell, line 1: yes this looks like a strange snapshot but this is due to an animation used with DropDown. 

Should be fix once animation timing can be controlled in Compose #678 #1693

colinmarsch (approved):
Just curious why this changed the accessibility snapshot images? It looks like there are some slight pixel differences

geoff-powell:
> Just curious why this changed the accessibility snapshot images? It looks like there are some slight pixel differences

ya, the pixel diff is most likely due to the window manager change.

PR #1695, opened 2024-11-15
Fix non-existent snapshot directory when running verifyPaparazziDebug

geoff-powell:


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -237,7 +237,7 @@ public class PaparazziPlugin @Inject constructor(
         test.inputs.dir(
           isVerifyRun.flatMap {
             project.objects.directoryProperty().apply {
-              set(if (it) snapshotOutputDir else null)
+              set(if (it && snapshotOutputDir.asFile.exists()) snapshotOutputDir else null)
```
ansman, line 240: If you use a file collection instead you won't have this issue. This works but creating that directory will cause a configuration cache miss when the directory is later created I believe.

geoff-powell: yep, I noticed that in the gradle docs but didn't go that way. Let me check that

ansman: Switching to this should work as a drop in replacement:
```kotlin
inputs.files(isVerifyRun.map { if (it) listOf(snapshotOutputDir) else emptyList() })
    .optional()
```

ansman: It works because calling `input.files` with an empty directory simply adds no input files

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -234,10 +234,10 @@ public class PaparazziPlugin @Inject constructor(
           .withPropertyName("paparazzi.nativeRuntime")
           .withPathSensitivity(PathSensitivity.NONE)
 
-        test.inputs.dir(
+        test.inputs.files(
           isVerifyRun.flatMap {
-            project.objects.directoryProperty().apply {
-              set(if (it) snapshotOutputDir else null)
+            project.objects.fileProperty().apply {
+              set(if (it) snapshotOutputDir.asFile else null)
```
ansman, line 240: I think you either want to ditch the file property or change keep it as a directory property. I'm not sure what happens if you pass a file property that points to a directory.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
ansman: To make caching work (and predictive test selection) the files themselves need to be included in the cache key so using a file property is not a good idea.

geoff-powell: yep makes sense, I opted to instead keep the directory property I originally had.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -258,12 +260,24 @@ public class PaparazziPlugin @Inject constructor(
...
+          test.systemProperties["paparazzi.test.verify"] = isVerificationTask
+
+          if (isVerificationTask && snapshotOutputDir.asFile.exists().not()) {
+            test.logger.log(
+              WARN,
+              """
+                Snapshot directory not found: ${snapshotOutputDir.asFile.path}
+                Please run the `recordPaparazzi$variantSlug` task to generate snapshots.
+              """.trimIndent()
+            )
```
geoff-powell, line 279: cc @jrodbx I added a warning log for cases where the verify task runs without a snapshot directory created.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
ansman: The main issue with this approach is that the outcome can depend on when the task is configured. I also want to make clear that a directory property will still include the contents of the directory in the cache key for the task so you're still going to be checksumming all the snapshots (which is a good thing)

PR #1699, opened 2024-11-18
Fail faster on ksp codegen, add custom lint checks for user experience

jrodbx:
Currently, the ksp plugin propagates error and empty state to the generated code, to be surfaced as test failures.  This PR proposes failing faster on ksp processing and adding custom lint checks as signals to the test author on how to properly use the annotation.


geoff-powell (approved):

geoff-powell:
needs a rebase

PR #1712, opened 2024-11-21
Add build support for JDK 21+

geoff-powell:


.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           windows-latest,
           ubuntu-latest
         ]
-        java-version: [17, 18]
+        java-version: [17, 18, 22]
```
TWiStErRob, line 26: Consider:
```suggestion
        java-version: [17, 21, 22]
```

So that it definitely works on LTS 21. Testing only on 22 could result in the library using 22-only features that are not in 21.

build.gradle
```diff
@@ -50,6 +50,9 @@ subprojects {
       showStackTraces true
       showCauses true
     }
+
+    // Enabled dynamic agent loading (ByteBuddy) for JDK 21+ https://openjdk.org/jeps/451
+    jvmArgs '-XX:+EnableDynamicAgentLoading'
```
TWiStErRob, line 55: Do consumers need to add this too?

(i.e. should this be part of the plugin, behind a version check: [`JavaVersion.current()`](https://github.com/TWiStErRob/net.twisterrob.gradle/blob/cff3808e335220f2ca9c3aa6803c8243b19a8123/docs/examples/local/settings.gradle.kts#L40-L42) or [`Test.javaVersion`](https://github.com/TWiStErRob/net.twisterrob.inventory/blob/d9c3f9dc2e80ae08fb83f606e436071fad8552e9/gradle/plugins-inventory/src/main/kotlin/net/twisterrob/inventory/build/allprojects.gradle.kts#L56-L70))

geoff-powell: good point let me update this 👍🏽

geoff-powell:
verified gradle java sdk and compile java sdk to 22 builds successfully.
https://github.com/cashapp/paparazzi/actions/runs/11959970311/job/33342949222

build.gradle
```diff
@@ -50,6 +50,11 @@ subprojects {
       showStackTraces true
       showCauses true
     }
+
+    if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_21)) {
```
TWiStErRob, line 54:
```suggestion
    if (javaVersion.isCompatibleWith(JavaVersion.VERSION_21)) {
```
I think you want this variant:
 * Test (receiver here).javaVersion is the resolved tool chain for this task
 * JavaVersion.current() is the Gradle wrapper daemon JVM. Those two could differ.

build.gradle
TWiStErRob: What do you think about making it part of the Paparazzi Gradle Plugin? If I added Paparazzi to a project running CI on 22, would I need to add this line too?

build.gradle
geoff-powell: For multiplatform, doing `Test.javaVersion` throws this error

```
tion: The value for task ':test-projects:multiplatform-plugin-with-android:testDebugUnitTest' property 'javaLauncher' is final and cannot be changed any further.
```

build.gradle
geoff-powell: yep, I added this in my latest commit

TWiStErRob: note: if you have it in plugin, then is it still necessary in root build.gradle?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -255,6 +256,13 @@ public class PaparazziPlugin @Inject constructor(
 
         test.outputs.dir(reportOutputDir).withPropertyName("paparazzi.report.dir")
 
+        if (test.javaVersion.isCompatibleWith(JavaVersion.VERSION_21)) {
+          // Enabled dynamic agent loading (ByteBuddy) for JDK 21+ https://openjdk.org/jeps/451
+          test.jvmArgs = (test.jvmArgs?.toMutableList() ?: mutableListOf()).apply {
+            add("-XX:+EnableDynamicAgentLoading")
+          }
```
TWiStErRob, line 263: consider:
```suggestion
          test.jvmArgs = test.jvmArgs.orEmpty() + "-XX:+EnableDynamicAgentLoading"
```
does the result **have to** be mutable?

jrodbx (approved):

PR #1737, opened 2024-12-10
Add lfsconfig and new field to improve git checkout speed with snapshots

### LINKED ISSUE #1729: Suggest using setLockableReadOnly in documentation
We have seen great improvement to our git checkout speed if we set `lfs.setLockableReadOnly=false`.

See:

- https://github.com/git-lfs/git-lfs/issues/5897
- https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-track.adoc
- https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-config.adoc

geoff-powell:
Closes #1729.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ git config core.hooksPath  # optional, confirm where your git hooks will be inst
 git lfs install --local
 git lfs track "**/snapshots/**/*.png"
 git add .gitattributes
+# Optional to improve git checkout performance
+git config lfs.setlockablereadonly false
 ```
 
 On CI, you might set up something like:


jrodbx (approved):
TIL, nice!  Any potential drawbacks?

geoff-powell:
> TIL, nice! Any potential drawbacks?

I don't think so, at least not that I saw in testing it in Cash. This really is for the default case for lockable files.
From lfs config docs:

```
These settings, the first an environment variable and the second a gitconfig setting, control whether files marked as 'lockable' in git lfs track are made read-only in the working copy when not locked by the current user. The default is true; you can disable this behaviour and have all files writeable by setting either variable to 0, 'no' or 'false'.
```

PR #1740, opened 2024-12-11
Add support for editable text on a11y snapshots

darshanparajuli:
I noticed editable text was missing in the a11y snapshots. This PR simply includes it if both context description and text semantic properties are missing.

I tried updating existing tests, not sure if that's the right thing to do here over adding specific test.


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * Kotlin 2.0.21
 * [Gradle Plugin] Gradle 8.10.2
 * [Gradle Plugin] Android Gradle Plugin 8.4.2
+* Support for editable text in accessibility snapshots
```
jrodbx, line 19: move to Unreleased.  this section is for an already-released version.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -169,6 +169,7 @@ public class AccessibilityRenderExtension : RenderExtension {
     val mainAccessibilityText =
       config.getOrNull(SemanticsProperties.ContentDescription)?.joinToString(", ")
         ?: config.getOrNull(SemanticsProperties.Text)?.joinToString(", ")
+        ?: config.getOrNull(SemanticsProperties.EditableText)?.text
```
jrodbx, line 172: does `joinToString` not make sense for this case?

jrodbx (approved):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
darshanparajuli: It's just a string, as opposed `List<String>` returned by content description and text.

CHANGELOG.md
darshanparajuli: oops!

colinmarsch:
Thanks for adding this!

Does this support already work for Views or do we need a similar addition in `View.accessibilityText()`?

darshanparajuli:
> Does this support already work for Views or do we need a similar addition in `View.accessibilityText()`?

@colinmarsch I just tested snapshotting `EditText` and it works as expected (it picks up on the actual text).

darshanparajuli:
Looks like `SemanticsProperties.Error` is also missing. I'm going to include it in this change as well if there are no objections.

jrodbx:
> Looks like `SemanticsProperties.Error` is also missing. I'm going to include it in this change as well if there are no objections.

Mind adding in a separate PR?

PR #1750, opened 2024-12-15
Couple Poko to Kotlin for Renovate PRs

drewhamilton:
Poko versions are often tightly coupled to Kotlin versions since the compiler plugin APIs are not stable. This makes it easier to update those together.

Also removed "Compose" mentions from the Kotlin group since Compose is no longer included in that group.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -11,11 +11,12 @@
       "automerge": true,
     },
     {
-      "description": "Compose compiler is tightly coupled to Kotlin version.",
-      "groupName": "Kotlin and Compose",
+      "description": "Compiler plugins are tightly coupled to Kotlin version.",
+      "groupName": "Kotlin",
       "matchPackagePrefixes": [
         "org.jetbrains.kotlin:",
-        "com.google.devtools.ksp"
+        "com.google.devtools.ksp",
+        "dev.drewhamilton.poko",
       ],
       "automerge": false,
     },


renovate.json5
```diff
@@ -11,11 +11,12 @@
       "automerge": true,
     },
     {
-      "description": "Compose compiler is tightly coupled to Kotlin version.",
-      "groupName": "Kotlin and Compose",
+      "description": "Compiler plugins are tightly coupled to Kotlin version.",
```
TWiStErRob, line 14: For reference Compose prefix was removed here: https://github.com/cashapp/paparazzi/commit/4c9a34bebef170637f35f7e04873a7a42ae021af#diff-31b1aa2b93afe11792d84c6cb5057a064b5a3294fa0caab92d7570ea604f759fR17

renovate.json5
```diff
@@ -11,11 +11,12 @@
...
+      "groupName": "Kotlin",
       "matchPackagePrefixes": [
         "org.jetbrains.kotlin:",
-        "com.google.devtools.ksp"
+        "com.google.devtools.ksp",
+        "dev.drewhamilton.poko",
```
jrodbx, line 19: Makes sense, my only concern here is the implied (and perhaps unrealistic) expectation that releases of Kotlin and Poko will be in relative lockstep with each other.  Thoughts?

drewhamilton: This doesn't inherently require that: Renovate will open a PR if either has a new version. This just means that if both have a new version, they'll land in the same PR—which you'll often need to get a successful compilation.

In any case, I think I'm pretty good about timely releases. I do long-running branches for Kotlin betas/RCs to try to catch potential issues early, and usually have the Poko release a couple days after the Kotlin release, if one is required. It did take me about a week when 2.1.0 dropped on Thanksgiving though 😅

jrodbx (approved):

Issue #1660, opened 2024-10-28
Font scaling applies to the text in the legend when using `AccessibilityRenderExtension`


colinmarsch:
**Description**
When applying larger font scaling on a test using the `AccessibilityRenderExtension`, the text in the legend also grows in size.

**Steps to Reproduce**

```
@get:Rule
  val paparazzi = Paparazzi(
    theme = "Theme.AppCompat.Light.NoActionBar",
    deviceConfig = DeviceConfig.PIXEL.copy(fontScale = 2.0f),
    renderExtensions = setOf(AccessibilityRenderExtension())
  )
```

**Expected behavior**
I would expect only the text in the UI of the application to increase in size and the legend text would remain the same size.

**Screenshots**

Default scale

![Image](https://github.com/user-attachments/assets/9cead7c5-81cf-4836-8a6c-ce1eb31571f5)

Large scale

![Image](https://github.com/user-attachments/assets/1c26c04c-5463-48e5-87b3-2470f64853e4)

PR #1760, opened 2024-12-18
Don't scale accessibility legend text when font scale is non-default

colinmarsch:
Resolves https://github.com/cashapp/paparazzi/issues/1660


colinmarsch:
- [ ] Need to add a test case that runs this new test class before merging

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1388,6 +1388,14 @@ class PaparazziPluginTest {
       .runFixture(fixtureRoot) { build() }
   }
 
+  @Test
+  fun scaledA11y() {
+    val fixtureRoot = File("src/test/projects/scaled-a11y")
+    gradleRunner
+      .withArguments("verifyPaparazziDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
```
jrodbx, line 1396: nit: can we merge this into ComposeA11yTest + use unsafeUpdateConfig?  imo, it's more obvious/readable to keep all A11y tests in the same project.

jrodbx (approved):

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: That suggestion sounds good to me, just updated this PR

PR #1798, opened 2025-01-15
Bypass font path prefix check in `ResourcesCompat::loadFont`

### LINKED ISSUE #1794: Font resources fail to load if in a directory that doesn't contain "res/"
**Description**
`ResourcesCompat::loadFont` checks for font paths to begin with `"res/"`, and returns null if they don't. This broke font loading in Paparazzi, so https://github.com/cashapp/paparazzi/pull/1586 worked around it by changing the `startsWith("res/")` to `contains("res/")`. This works if the font is indeed in a directory called "res", but this won't always be the case. In our multiplatform project, our font files are in a dir called "resources" which is added to our Android source set, and that works totally fine at runtime and in the IDE. But in Paparazzi tests, we get `java.lang.IllegalStateException: Could not load font` due to the `contains("res"/)` check failing.
Could this check be skipped entirely in Paparazzi?

**Steps to Reproduce**
1. Create a dir called "resources/font" and put a .ttf font in it
2. Add that dir to Android res sources with `android.sourceSets.named("main").res.srcDirs("resources")`
3. In a Paparazzi test, try to load this Font (e.g. using a Compose `Font` pointing to the font resource

Observe `java.lang.IllegalStateException: Could not load font` is thrown.

**Expected behavior**
Fonts should load as long as they are valid Android resources.

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: macOS 15
- Compile SDK: 33
- Gradle Version: 8.10.2
- Android Gradle Plugin Version: 8.8.0

DSteve595:
This changes the way in which `ResourcesCompatVisitorFactory` works around the annoying font prefix check in `ResourcesCompat::loadFont`, which checks that font paths begin with `"res/"`. The workaround introduced in https://github.com/cashapp/paparazzi/pull/1586 works when the font's file path in the project _contains_ `"res/"`, but that's not guaranteed (e.g. alternate source directories, generated sources).
Given that I don't think this resource path check gives any value to Paparazzi tests, this fixes those edge cases by bypassing the check completely, assuming `startsWith` returns true. Still hacky, but it already was.

Fixes #1794


paparazzi-gradle-plugin/src/test/projects/custom-fonts/build.gradle
```diff
@@ -17,6 +17,9 @@ android {
   kotlinOptions {
     jvmTarget = libs.versions.javaTarget.get()
   }
+
+  // For testing loading font resources from directories other than 'res/'
+  sourceSets.main.res.srcDirs += 'src/main/other_resources_dir'
```
geoff-powell, line 22: nit: this looks good but might be better in a separate test project. Not a blocker for this PR.

geoff-powell (approved):
this change makes sense instead of an arbitrary `contains` method, skipping the whole `startsWith` method eval. 
Thanks for the contribution!!

jrodbx:
> The workaround introduced in https://github.com/cashapp/paparazzi/pull/1586

FWIW, that workaround was adopted from Android Studio's own implementation here:
https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/classloading/ResourcesCompatTransform.kt;l=64?q=ResourcesCompat

>, but that's not guaranteed (e.g. alternate source directories, generated sources).

generated sources are handled in a different manner.  I'm curious: what's the use case for having alternate resource directories @DSteve595?  And more importantly, does that mean that Android Studio's Compose Previews are broken for you given its reliance on the above linked implementation?

geoff-powell:
> generated sources are handled in a different manner. I'm curious: what's the use case for having alternate resource directories @DSteve595? And more importantly, does that mean that Android Studio's Compose Previews are broken for you given its reliance on the above linked implementation?

yes I had the same thoughts. As this unblocks your issue, I have the same hesitancy to mimic AS vs our own differing solution.

DSteve595:
Our use case is basically just multiplatform, where we're using Android's directory structure but happened to called the directory "resources" rather than "res". It's something that's easily changed on our part (or easily worked around by copying the resources into a generated dir named "res"). But my thinking with this was: if it's a valid resource that an Android app can read, then Paparazzi should be able to read it too.
Previews are broken in Android Studio too, and I didn't realize it was due to this. I consider it an Android Studio bug (well, really a ResourcesCompat bug, but one that affects Studio), for the same reason.
Feel free to revert if matching Studio's behavior seems preferable.

Issue #1716, opened 2024-11-22
Golden images in `/build/paparazzi/failures` are not stored in some error cases


SimonMarquis:
I'm not sure if this should be classified as a bug of a feature request:

**Description**
Golden images are stored in `/build/paparazzi/failures` when there is a regression (next to the `delta-*.png` files) but, they are not when:
- The reference image (in `/test/snapshots/images`) is not present:
  ```
      java.lang.AssertionError: File module/ScreenshotTesting/src/test/snapshots/images/...ScreenshotTest_test.png does not exist
          at app.cash.paparazzi.SnapshotVerifier$newFrameHandler$1.handle(SnapshotVerifier.kt:60)
          at app.cash.paparazzi.Paparazzi$apply$1.evaluate$lambda$0(Paparazzi.kt:73)
          at app.cash.paparazzi.PaparazziSdk.takeSnapshots(PaparazziSdk.kt:333)
          at app.cash.paparazzi.PaparazziSdk.snapshot(PaparazziSdk.kt:178)
          at app.cash.paparazzi.PaparazziSdk.snapshot$default(PaparazziSdk.kt:176)
          at app.cash.paparazzi.PaparazziSdk.snapshot(PaparazziSdk.kt:173)
          at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:103)
  ```

- The reference image has a different size:
  ```
      java.lang.IllegalStateException: Images are different sizes
          at app.cash.paparazzi.internal.OffByTwo.compare(OffByTwo.kt:11)
          at app.cash.paparazzi.internal.ImageUtils.compareImages(ImageUtils.kt:144)
          at app.cash.paparazzi.internal.ImageUtils.assertImageSimilar(ImageUtils.kt:55)
          at app.cash.paparazzi.SnapshotVerifier$newFrameHandler$1.handle(SnapshotVerifier.kt:73)
          at app.cash.paparazzi.Paparazzi$apply$1.evaluate$lambda$0(Paparazzi.kt:73)
          at app.cash.paparazzi.PaparazziSdk.takeSnapshots(PaparazziSdk.kt:333)
          at app.cash.paparazzi.PaparazziSdk.snapshot(PaparazziSdk.kt:178)
          at app.cash.paparazzi.PaparazziSdk.snapshot$default(PaparazziSdk.kt:176)
          at app.cash.paparazzi.PaparazziSdk.snapshot(PaparazziSdk.kt:173)
[...]

PR #1821, opened 2025-02-02
Remove early size check requirement in `OffByTwo.kt`

SimonMarquis:
This will make `OffByTwo.kt` behave identical to the original `PixelPerfect` differ.

But most importantly, it will restore the initial behavior of `ImageUtils.assertImageSimilar()` itself, checking for differences in size and correctly producing the expected delta (and actual) images in case of a size mismatch.

Relates to #1716


SimonMarquis:
Let me know if you want a non-regression test to be added for this behavior.

geoff-powell:
>Let me know if you want a non-regression test to be added for this behavior.

Yes, please can you add a regression test for this so this doesn't happen in the future! 🙏🏽

geoff-powell (approved):

PR #1844, opened 2025-02-24
Check the Gradle distribution checksum

TheMrMilchmann:
It is generally a good idea to ensure the integrity of the Gradle distribution to prevent MITM attacks when downloading. (See https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification)

Renovate automatically updates this.


geoff-powell (approved):
Thanks for the contribution!

PR #1853, opened 2025-02-26
Migrate to new sonatype and signing credentials

JakeWharton:


.github/workflows/build.yml
```diff
@@ -72,11 +72,12 @@ jobs:
           java-version: 17
 
       - name: Publish Artifacts
-        run: ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publishAllPublicationsToMavenCentralRepository --no-parallel
+        run: ./gradlew publishMavenPublicationToMavenCentralRepository paparazzi-gradle-plugin:publishAllPublicationsToMavenCentralRepository
```
JakeWharton, line 75: The parallel problem has been fixed for a few years now

jrodbx: We still saw problems with this as recently as a year ago, but if we're asking _that_ has been fixed since then, sgtm!

jrodbx (approved):

PR #1880, opened 2025-03-18
Modernize Paparazzi API documentation webpages

colinmarsch:
Updates the Paparazzi API docs to use a more modern styling provided through dokka that is also used in other libraries such as [SQLDelight](https://sqldelight.github.io/sqldelight/2.0.2/2.x/extensions/coroutines-extensions/app.cash.sqldelight.coroutines/index.html). Dark / light mode is a configuration option that is selectable on the webpages.

| Before |  After |
| --- | --- |
| image | Screenshot 2025-03-18 at 3 45 14 PM |

### DIFF
--- a/.github/workflows/prepare_mkdocs.sh
+++ b/.github/workflows/prepare_mkdocs.sh
@@ -9,27 +9,7 @@
 set -ex
 
 # Generate the API docs
-./gradlew dokkaGfm
-
-# Dokka filenames like `-http-url/index.md` don't work well with MkDocs <title> tags.
-# Assign metadata to the file's first Markdown heading.
-# https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data
-title_markdown_file() {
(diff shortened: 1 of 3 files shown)


geoff-powell (approved):

jrodbx:
Quick question: was the move from Gfm to Html what enabled the light to dark mode feature?  

Mainly asking because I'm reviewing https://github.com/cashapp/paparazzi/pull/1783/files and I don't recall why we used Gfm in the first place, but probably better that we aren't anymore?

geoff-powell:
good question, I need to update that pr with the latest master

colinmarsch:
> Quick question: was the move from Gfm to Html what enabled the light to dark mode feature?
> 
> Mainly asking because I'm reviewing https://github.com/cashapp/paparazzi/pull/1783/files and I don't recall why we used Gfm in the first place, but probably better that we aren't anymore?

The dark mode feature was added in this PR: https://github.com/cashapp/paparazzi/pull/1887. It's just an mkdocs yaml configuration change. I think it's independent of gfm vs html

PR #1928, opened 2025-04-23
Verify screenshots from the `:sample` module

SimonMarquis:


.github/workflows/build.yml
```diff
@@ -42,8 +42,8 @@ jobs:
         with:
           validate-wrappers: true
 
-      - name: Run All Tests
-        run: ./gradlew check
+      - name: Run All Tests and verify :sample module
+        run: ./gradlew check :sample:verifyPaparazziDebug
```
SimonMarquis, line 46: This can be splitted in 2 distinct jobs if you prefer.

TWiStErRob: Should this be left intact and instead hook :check to depend on :sample:verifyPaparazziDebug? That way local `gradlew build` would also verify them.

SimonMarquis: That is another alternative indeed.  
Although since Paparazzi does not do this by default, it might be misleading for readers to configure it this way.

sample/src/test/snapshots/.gitkeep
geoff-powell, line 1: nit: can you remove this?

.github/workflows/build.yml
geoff-powell: yes please add this dependency configuration to sample.gradle. Thanks for adding this!

geoff-powell (approved):

sample/src/test/snapshots/.gitkeep
SimonMarquis: yes, it was to ensure the CI would correctly fail.
(when the snapshot directory is not present, it fails in other ways)

PR #1932, opened 2025-04-24
Don't validate APIs for test projects

Goooler:
Fixes

```
> Task :test-projects:build-class:apiCheck FAILED
Execution failed for task ':test-projects:build-class:apiCheck'.
> Expected file with API declarations 'paparazzi-gradle-plugin/src/test/projects/build-class/api/build-class.api' does not exist.
  Please ensure that ':apiDump' was executed in order to get an API dump to compare the build against
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,8 +27,14 @@ buildscript {
 
 apply plugin: 'binary-compatibility-validator'
 
+def testProjectPath = 'paparazzi-gradle-plugin/src/test/projects'.replace('/', File.separator)
+// These projects are imported in IDEA but we don't want to run the API validation on them.
+def testProjectNames = allprojects.findAll {
+  it.projectDir.path.contains(testProjectPath)
+}.collect { it.name }
+
 apiValidation {
-  ignoredProjects += ['sample']
+  ignoredProjects += ['sample'] + testProjectNames
 }
 
 subprojects {


colinmarsch:
When does this error happen? Just trying to figure out why this fix might be needed

Goooler:
This happends due to running `./gradlew apiDump` in IDEA, as the test projects are enabled in IDEA env

https://github.com/cashapp/paparazzi/blob/d2da4fd1854b8c66b02afcfc8b6bb8d6b4be558a/settings.gradle#L37-L40

The test projects are disabled on CI, so `apiCheck` tasks never fail.

build.gradle
```diff
@@ -27,8 +27,14 @@ buildscript {
...
 
+def testProjectPath = 'paparazzi-gradle-plugin/src/test/projects'.replace('/', File.separator)
+// These projects are imported in IDEA but we don't want to run the API validation on them.
+def testProjectNames = allprojects.findAll {
+  it.projectDir.path.contains(testProjectPath)
+}.collect { it.name }
+
 apiValidation {
-  ignoredProjects += ['sample']
+  ignoredProjects += ['sample'] + testProjectNames
```
jrodbx, line 37: how about using 

`ignoredPackages += ["app.cash.paparazzi.plugin.test"]`

instead?  Helps us avoid another use of allprojects/subprojects to be tackled whenever isolated projects hits stable.

jrodbx (changes requested):

build.gradle
Goooler: Need to dump empty API files for these projects first.

SimonMarquis:
ℹ️ BCV will be phased out in favor of a built-in setting in the KGP and is already available since Kotlin 2.2.0.
It might be a good idea to also migrate this (the update is trivial) if we also proceed to update the Kotlin version.
More details on https://kotlinlang.org/docs/whatsnew22.html#binary-compatibility-validation-included-in-kotlin-gradle-plugin

geoff-powell:
@SimonMarquis made the same change here https://github.com/cashapp/paparazzi/pull/2088 and its merged

PR #1936, opened 2025-04-26
Add `Paparazzi(withExpectedActualLabels: Boolean = true)`

### LINKED ISSUE #1691: Allow removal of `actual` & `expected` labels in delta images
As mentioned in https://github.com/cashapp/paparazzi/issues/1465#issuecomment-2474104888, I was wondering if there is a way to remove the `actual` & `expected` labels on the generated delta image?
For us, this does not provide any useful information since we are used to have before/after images next to each other, and if the changed pixels were located below those label, they would hide it.
I feel like "left == before" & and "right == after" is a common convention, but that might be true only for us using LTR locales.

The boolean `ApngVerifier#withErrorText` seems to be be related to this, but it is not exposed as a public API (and always true in the default configuration).

SimonMarquis:
to allow removal of `actual` & `expected` labels in delta images. Keeping the previous behavior as default value.

Closes #1691


SimonMarquis:
Let me know if you want to change the naming, the default value, or even the order in the signatures.

PR #1937, opened 2025-04-26
Disable unnecessary `release` build type in `sample` module

SimonMarquis:
It will save around 2 minutes on each CI build (3 * OS + 3 * JDK).

Here is an example of a run on the master branch: https://scans.gradle.com/s/4oc5b3rznajmi/timeline?name=release&project=:sample&sort=longest

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -22,6 +22,13 @@ android {
   }
 }
 
+// Disable unnecessary release build type
+androidComponents {
+  beforeVariants(selector().all()) {
+    enabled = buildType == "debug"
+  }
+}
+
 dependencies {
   implementation libs.composeUi.material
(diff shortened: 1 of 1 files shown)


sample/build.gradle
```diff
@@ -22,6 +22,13 @@ android {
   }
 }
 
+// Disable unnecessary release build type
+androidComponents {
+  beforeVariants(selector().all()) {
```
TWiStErRob, line 27: Would it make sense to encode the comment in code?
```
beforeVariants(selector().withBuildType("release")) { enable = false }
```
```

SimonMarquis: I tested mayny things with the Groovy API and forgot to actually make it straightforward 🙈 
Thanks!

Issue #1904, opened 2025-04-01
Accessibility snapshots do not print "In list" for Lazy Containers


hrcronin:
**Description**
Talkback will announce every item in a Lazy Container (i.e. LazyGrid, LazyColumn, etc.) as "item in list" or "Item out of list", depending on whether the view is in or out of the lazy container. This is not reflected in the accessibility snapshot

**Steps to Reproduce**
In the test below the items in the `LazyColumn` are just `Text` with a number appended. As an example, the accessibility snapshot just shows "Item 1", instead of "Item 1, in list", which is what the TalkBack reader announces.

```
class ScaffoldTest {
    @get:Rule
    val paparazzi = Paparazzi(
        renderExtensions = setOf(AccessibilityRenderExtension()),
        maxPercentDifference = 0.0,
        deviceConfig = DeviceConfig.PIXEL_5
    )

    @Test
    fun default() {
        val viewToCapture = @Composable {
            // allow tests to identify as Previews
            // in Fractal, we use this value to selectively support dark theme and system Locale with its corresponding layout direction
            CompositionLocalProvider(
                LocalInspectionMode provides true
            ) {
                AppTheme {
                    Scaffold(
                        modifier = Modifier.fillMaxWidth(),
                        topBar = {
                            Box(Modifier
                                .fillMaxWidth()
                                .background(MaterialTheme.colorScheme.primaryContainer)
                                .padding(10.dp)
                            ) {
                                Text("Title", color = MaterialTheme.colorScheme.onPrimaryContainer)
                            }
                        },
                        content = { padding ->
                            LazyColumn(modifier = Modifier.padding(padding)) {
                                items(5) {
                                    Text(text ="Item = $it")
                                }
                            }
                        }
                    )
                }
            }
        }
        paparazzi.snapshot(
            name = "Scaffold",
            composable = viewToCapture
        )
    }
}
```

**Expected behavior**
Every item in a lazy container should print out "in list" or "out of list" in the accessibility snapshot

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: 15
- Compile SDK: 35
- Gradle Version: 8.11.1
- Android Gradle Plugin Version: 8.9.0

**Screenshots**
![Image](https://github.com/user-attachments/assets/44f6e74d-d5d5-46d6-a666-a62339865c45)

**Talkback Experience on device**

https://github.com/user-attachments/assets/1fe56546-5c4b-493a-95eb-b9b9a609b8f4

PR #1942, opened 2025-05-01
Denote elements in lists in the accessibility snapshot legend

colinmarsch:
This changes adds support for denoting elements that are in a list in the accessibility snapshot legend. This aligns with Talkback behaviour that announces every item in a list component as "item in list".

Resolves https://github.com/cashapp/paparazzi/issues/1904


adamalyyan (approved):
Very nice!

geoff-powell (approved):

PR #1978, opened 2025-05-29
Fully advance choreographer correctly for Compose

geoff-powell:
Kudos to @mattprecious for the find

Compose UI 1.8 moved ripples into native when running with layoutlib.
The native layer is unable to see the manual clock ticks that are
currently being done and throws an exception due to the frame time
being 0. Invoking doFrame on the choreographer propagates the time to
the native layer.

This commit also cleans up clock management logic that was creating
invalid uptimes.

https://android.googlesource.com/platform/frameworks/support/+/09bea4c5dea7cf1aacf1a49b2f15595d9c159e2d


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -347,7 +348,7 @@ public class PaparazziSdk @JvmOverloads constructor(
 
     // Execute the block at the requested time.
     System_Delegate.setNanosTime(0L)
-    System_Delegate.setNanosTime(frameNanos)
+    Choreographer_Delegate.sChoreographerTime = frameNanos
```
geoff-powell, line 351: Now that we are using `Choreographer` frame we need to make sure the time is correct.

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -108,12 +108,16 @@ class PaparazziTest {
       "onDraw time=1000 animationElapsed=0.0",
       "onAnimationStart time=2000 animationElapsed=0.0",
       "onAnimationUpdate time=2000 animationElapsed=0.0",
+      "onAnimationUpdate time=2000 animationElapsed=0.0",
```
geoff-powell, line 111: A side-effect of `Choreographer_Delegate.doFrame` and `choreographerCallbacks.execute(currentTimeNanos, Bridge.getLog())` executing multiple update callback.

Need to look further into this as removing `choreographerCallbacks.execute(currentTimeNanos, Bridge.getLog())` causes frame times for animations to be incorrect

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -108,12 +108,16 @@ class PaparazziTest {
...
       "onDraw time=2000 animationElapsed=0.0",
       "onAnimationUpdate time=2250 animationElapsed=0.25",
+      "onAnimationUpdate time=2250 animationElapsed=0.25",
```
jrodbx, line 114: this is a bug; let's fix before merging this

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -339,6 +340,12 @@ public class PaparazziSdk @JvmOverloads constructor(
       if (hasComposeRuntime) {
         forceReleaseComposeReferenceLeaks()
       }
+
+      // Reset the choreographer to its initial state for last for future test runs as it is a singleton.
+      val choreographer = Choreographer.getInstance()
+      val mLastFrameTimeNanos = choreographer::class.java.getDeclaredField("mLastFrameTimeNanos")
+      mLastFrameTimeNanos.isAccessible = true
+      mLastFrameTimeNanos.set(choreographer, 0L)
```
geoff-powell, line 348: @jrodbx was able to step through and see what was going on. Since Choreographer is a singleton, that last frame time carries over from previous tests, causing animation timing issues.

Thought it would be a good clean up step to reset it to 0 after snapshot. Could also do it before in the prepare step.

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
geoff-powell: Yep agreed, found the culprit to my choreographer testing and updated the pr

PR #1983, opened 2025-06-04
Add fix for boundsInWindow incorrectly calculating position on screen

geoff-powell:
SemanticsNode.boundsInWindow was incorrectly reporting bounds with compose 1.8.x update.

This uses a combination of boundsInRoot and locationOnScreen to correctly calculate displayBounds.

Example of bug:
472a194c6c73002fce0a4bd046522876087c52a1


colinmarsch (approved):
We didn't have any tests in the repo that failed due to this issue?

PR #1989, opened 2025-06-10
Run builds on Java 24

geoff-powell:


.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           windows-latest,
           ubuntu-latest
         ]
-        java-version: [17, 21, 23]
+        java-version: [17, 21, 23, 24]
```
jrodbx: nit: replace 23 with 24, instead of appending?

paparazzi-gradle-plugin/src/test/projects/lifecycle-usages/src/test/java/app/cash/paparazzi/plugin/test/LifecycleUsageTest.kt
```diff
@@ -25,7 +24,7 @@ class LifecycleUsageTest {
     var currentLifecycleState: Lifecycle.State? = null
 
     view.doOnAttach {
-      val lifecycleOwner = ViewTreeLifecycleOwner.get(view)!!
+      val lifecycleOwner = view.findViewTreeSavedStateRegistryOwner()!!
```
jrodbx, line 27: should this have been merged in this PR?  why was this needed for Java 24?

PR #1993, opened 2025-06-16
Add ability to avoid recording if difference threshold not exceeded

### LINKED ISSUE #1951: Record new screenshots only if changes exceed the `maxPercentDifference` option
Running `recordPaparazzi` on large modules after updating something tangential to the UI such as the AGP often results in a huge number of updated snapshots. These can be hard to review, especially in large projects where the number of snapshots can reach into the thousands. 

To improve this, should we add an option to record new snapshots only if the changes exceed `maxPercentDifference`?

svyatoslav-starkov:
Added a boolean flag to use `maxPercentDifference` for `recordPaparazzi` command.
Not sure if it should be a boolean flag or a separate double type to set a different `maxPercentDifference` for record only.

Also, uncertain about API compatibility with the changes I introduced.

The problem with different operating systems that I described [here](https://github.com/cashapp/paparazzi/issues/1951#issuecomment-2976978049) is no longer reproducible in either direction (Mac/Windows). For AGP updates or other things that could have caused problems, this fix should work as well.

Closes #1951

### DIFF
--- a/paparazzi/api/paparazzi.api
+++ b/paparazzi/api/paparazzi.api
@@ -95,7 +95,8 @@ public final class app/cash/paparazzi/HtmlReportWriter : app/cash/paparazzi/Snap
 	public fun <init> (Ljava/lang/String;)V
 	public fun <init> (Ljava/lang/String;Ljava/io/File;)V
 	public fun <init> (Ljava/lang/String;Ljava/io/File;Ljava/io/File;)V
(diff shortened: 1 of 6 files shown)


michal-chojnacki:
@geoff-powell Is there any chance this functionality will be added in the near future? It’s something that’s really missing when using Paparazzi.

svyatoslav-starkov:
Closing since the issue has been resolved: https://github.com/cashapp/paparazzi/pull/2067

Issue #1836, opened 2025-02-14
Add option to run the Compose "UI checks" that studio has


yogurtearl:
To better test Compose Accessibility, it would be good if we could run the "Compose UI Checks" that are available in Studio via Compose previews. 

Docs are here:
https://developer.android.com/guide/topics/ui/accessibility/testing#compose-ui-check

### Context

There is a `validateAccessibility` option, but seems to do something different? seems to use [ATF](https://github.com/google/Accessibility-Test-Framework-for-Android) to check Views (not Compose) ?

colinmarsch:
ATF supports Compose accessibility checks on API 34+ per [this issue thread](https://github.com/google/Accessibility-Test-Framework-for-Android/issues/40#issuecomment-2127976675). ATF is also what is used under the hood in Android Studio for the Compose UI checks (also mentioned in that linked comment thread). 

I will verify this is the case next week, but I believe with `layoutlib` version `15.x.x`+ that should include the Compose ATF checks and Paparazzi will have inherited that upgrade for free. For reference the `layoutlib` version at the time of the `1.3.5` Paparazzi release was just `14.0.11` ([code link](https://github.com/cashapp/paparazzi/blob/74ae7cfe42d7c0fb16358c3fe75e0d7a175718a8/gradle/libs.versions.toml#L12)).

A Google issue tracker thread that is relevant here for just some additional context is [this one](https://issuetracker.google.com/issues/198122348).

colinmarsch:
Just to provide an update here, I looked into this and it looks like it isn't just working out of the box unfortunately. Still have some more investigation to do (that I will continue next week) to figure out what might be missing here for us to support the Compose UI checks.

jamieherrera:
Any updates?

colinmarsch:
> Any updates?

Sorry for not providing an update on this earlier. The investigation that I mentioned above stalled and no more progress has been made on this. 

Beyond that however, I would like to propose deprecating (and eventually removing) the `validateAccessibility` functionality from Paparazzi. Paparazzi is, at it's core, a _visual_ snapshot library and this feature seems adjacent at best to that goal. Just because Paparazzi happens to be able to support this feature due to consuming `layoutlib` internally doesn't mean it should be part of the library in my view.

[...]

PR #1997, opened 2025-06-18
Deprecate validateAccessibility parameter

colinmarsch:
Paparazzi is, at it's core, a visual snapshot library and this feature seems adjacent at best to that goal. Just because Paparazzi happens to be able to support this feature due to consuming layoutlib internally doesn't mean it should be part of the library in my view.

I believe the future of accessibility testing within Paparazzi should live entirely with the AccessibilityRenderExtension and it's [visual based accessibility snapshots](https://cashapp.github.io/paparazzi/accessibility/). Removing the validateAccessibility feature would help provide clarity in Paparazzi's accessibility API and reduce some maintenance burden going forward.

Relates to https://github.com/cashapp/paparazzi/issues/1836 and in particular this comment https://github.com/cashapp/paparazzi/issues/1836#issuecomment-2907635271.


colinmarsch:
> this makes sense. what will this look like in practice (pointer to design doc/discussion maybe)?
> 
> is the concept that enabling accessibility tests will automatically enable accessibility validation (if so, makes sense!)?

My suggestion here is actually to remove the support for accessibility _validation_ from Paparazzi in favour of just focusing on the visual accessibility snapshots created by the `AccessibilityRenderExtension`. 

As I mentioned in the description, I think this validation feature sort of just came about due to the fact that it is available in `layoutlib` and therefore is relatively easy for Paparazzi to support. Looking at it with hindsight now, I don't think it really fits within the goals of Paparazzi as a snapshot library. The current state of the validation is more or less just a log of warnings generated and moving towards a more useful and strict model where we check against the past outputs of the tool feels to stray farther from Paparazzi's core use case.

jrodbx:
> My suggestion here is actually to remove the support for accessibility _validation_ from Paparazzi in favour of just focusing on the visual accessibility snapshots created by the `AccessibilityRenderExtension`.

Oh gotcha.

> As I mentioned in the description, I think this validation feature sort of just came about due to the fact that it is available in `layoutlib` and therefore is relatively easy for Paparazzi to support. Looking at it with hindsight now, I don't think it really fits within the goals of Paparazzi as a snapshot library. The current state of the validation is more or less just a log of warnings generated and moving towards a more useful and strict model where we check against the past outputs of the tool feels to stray farther from Paparazzi's core use case.

The 2.0 design work is about decomposing Paparazzi into useful parts.  One being the snapshot library (via the SDK), another the JUnit Rule(s) and yet another are the RenderExtensions (of which, Accessibility is the flagship example).

I could see a case for keeping validation as a feature and extending it to support Compose "UI Checks" as there remain potential gaps from relying solely on the image for information, e.g., in [this image from the developer docs](https://developer.android.com/static/studio/images/design/compose-ui-check.png): are "insufficient color constraint ratio, too many characters, field too wide, target size too small" etc. subjectively easy to miss?  would the additional logging be useful?  up to you!

colinmarsch:
>I could see a case for keeping validation as a feature and extending it to support Compose "UI Checks" as there remain potential gaps from relying solely on the image for information, e.g., in [this image from the developer docs](https://developer.android.com/static/studio/images/design/compose-ui-check.png): are "insufficient color constraint ratio, too many characters, field too wide, target size too small" etc. subjectively easy to miss? would the additional logging be useful? up to you!

Yeah I agree there are still accessibility coverage gaps from just the image based approach, but I'm not sure Paparazzi is the right library to fill those gaps in a text based way like how Android Studio does it for compose previews today. We could add some of these to the visual accessibility snapshots as well potentially. Touch target size could be determined from the bounding boxes that are already computed.

PR #2001, opened 2025-06-20
Allow differ to be configured

jrodbx:
Highlights
* Differ becomes a sealed interface
* Test rule checks for new gradle property ("app.cash.paparazzi.differ") defaulting to OffByTwo

### DIFF
--- a/paparazzi/api/paparazzi.api
+++ b/paparazzi/api/paparazzi.api
@@ -213,9 +213,9 @@ public final class app/cash/paparazzi/SnapshotHandlerKt {
 
 public final class app/cash/paparazzi/SnapshotVerifier : app/cash/paparazzi/SnapshotHandler {
 	public static final field $stable I
-	public fun <init> (D)V
-	public fun <init> (DLjava/io/File;)V
-	public synthetic fun <init> (DLjava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (DLapp/cash/paparazzi/internal/Differ;)V
+	public fun <init> (DLjava/io/File;Lapp/cash/paparazzi/internal/Differ;)V
+	public synthetic fun <init> (DLjava/io/File;Lapp/cash/paparazzi/internal/Differ;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun close ()V
 	public fun newFrameHandler (Lapp/cash/paparazzi/Snapshot;II)Lapp/cash/paparazzi/SnapshotHandler$FrameHandler;
 }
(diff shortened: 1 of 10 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/Differ.kt
```diff
@@ -2,23 +2,23 @@ package app.cash.paparazzi.internal
...
+public sealed interface Differ {
+  public fun compare(expected: BufferedImage, actual: BufferedImage): DiffResult
 
-  sealed interface DiffResult {
-    data class Identical(
-      val delta: BufferedImage
+  public sealed interface DiffResult {
+    public class Identical(
+      public val delta: BufferedImage
     ) : DiffResult
```
jrodbx, line 11: componentN, copy, hashcode seemed unnecessary and surfaced in the apiDump.

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -148,9 +151,19 @@ public class Paparazzi @JvmOverloads constructor(
     private val isVerifying: Boolean =
       System.getProperty("paparazzi.test.verify")?.toBoolean() == true
 
-    private fun determineHandler(maxPercentDifference: Double): SnapshotHandler =
+    private val differ: Differ =
+      System.getProperty("app.cash.paparazzi.differ")?.lowercase().let { differ ->
+        when (differ) {
```
geoff-powell, line 156: nit: Instead of arbitrary values, what if we opened this up to pass the FQCN instead?

geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
jrodbx: We could, but making the internal Differs public and invoking `compare` via reflection doesn't feel right.

Any benefit to that approach that i'm missing?

geoff-powell: My reasoning is trying to think of any scenario where people can provide their own diffing logic, if desired.

paparazzi/api/paparazzi.api
```diff
@@ -213,9 +213,9 @@ public final class app/cash/paparazzi/SnapshotHandlerKt {
 
 public final class app/cash/paparazzi/SnapshotVerifier : app/cash/paparazzi/SnapshotHandler {
 	public static final field $stable I
-	public fun <init> (D)V
-	public fun <init> (DLjava/io/File;)V
-	public synthetic fun <init> (DLjava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (DLapp/cash/paparazzi/internal/Differ;)V
```
JakeWharton, line 216: This puts an internal type into public API. The type should probably be moved somewhere public.

jrodbx: addressed in https://github.com/cashapp/paparazzi/pull/2009

PR #2010, opened 2025-06-23
Introduce snapshot browser

jrodbx:
This PR pivots from the "development" mode (to be removed in a follow-up) towards an aggregating root project feature -- a snapshot browser -- which pulls the snapshot metadata from dependent modules and displays them in a filterable interface.

The bulk of this change is getting Gradle to propagate each project's metadata up to the root project to aggregate.  Calling out that the plugin changes uses `allprojects` which will probably break isolated projects for the time being (is that ok?)

Note that this change also adds a new tags (Map<String, String>) attribute, but only to the test rule methods `snapshot` and `gif` and doesn't bleed into the SDK efforts.  The idea here is to allow feature teams to add custom tags denoting team ownership, or really anything else.

Still a work in progress, but the hopes are that the community can help drive the development of this feature.

https://github.com/user-attachments/assets/044bc1fe-a546-4c51-9752-79a4c7c2f195


geoff-powell (approved):
Nice this is great, seemed relatively quick to spin up!

eygraber:
> Calling out that the plugin changes uses allprojects which will probably break isolated projects for the time being (is that ok?)

Is there going to be a way to disable this?

geoff-powell:
> > Calling out that the plugin changes uses allprojects which will probably break isolated projects for the time being (is that ok?)
> 
> Is there going to be a way to disable this?

Good point, we can add a gradle property to disable this task registration and plumbing.

VladZawinski:
Hello @jrodbx , May I know it is ready to use?
Having Search feature is promising. Now I need to search the components out alphabetically.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -93,6 +106,22 @@ public class PaparazziPlugin @Inject constructor(
...
+  private fun Project.configureRootProject() {
+    val resolver = createResolver()
+    tasks.register("generateSnapshotBrowser", GenerateBrowserTask::class.java) { task ->
+      val paths = resolver.map { c -> c.incoming.artifactView { it.lenient(true) }.artifacts.artifactFiles }
+      task.projectRoot.set(providerFactory.provider { layout.projectDirectory.asFile.path })
+      task.snapshotMetadataReports.setFrom(paths)
+      task.output.set(project.layout.buildDirectory.dir("paparazzi"))
+    }
+
+    allprojects.forEach { p ->
```
eygraber, line 130: Should this be `subprojects`? I believe that `allprojects` is `subprojects` + the current project.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -93,6 +106,22 @@ public class PaparazziPlugin @Inject constructor(
...
+      dependencies.let { d ->
+        d.add(SNAPSHOT_METADATA.declarableName, d.project(mapOf("path" to p.path)))
```
eygraber: If the `path` is all that's needed, would `p.isolated.path` work? I think that should be project isolation compliant.

jrodbx: good point, will explore and follow-up

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: the root/current project also needs to be added

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: done/works, thanks for the suggestion

jrodbx:
FYI, unsure why but https://github.com/cashapp/paparazzi/pull/2088 breaks IDE sync, will continue to investigate (unless someone has ideas).  Would like to get that fixed and land this in alpha03.

Issue #1902, opened 2025-04-01
Accessibility Snapshots print Scaffold items out of order


hrcronin:
**Description**
When using a scaffold, the talkback reader will read the screen from top to bottom. I would expect the accessibility snapshot to print in the same order.

**Steps to Reproduce**
Sample test to reproduce:
```
class ScaffoldTest {
    @get:Rule
    val paparazzi = Paparazzi(
        renderExtensions = setOf(AccessibilityRenderExtension()),
        maxPercentDifference = 0.0,
        deviceConfig = DeviceConfig.PIXEL_5
    )

    @Test
    fun default() {
        val viewToCapture = @Composable {
            // allow tests to identify as Previews
            // in Fractal, we use this value to selectively support dark theme and system Locale with its corresponding layout direction
            CompositionLocalProvider(
                LocalInspectionMode provides true
            ) {
                AppTheme {
                    Scaffold(
                        modifier = Modifier.fillMaxWidth(),
                        topBar = {
                            Box(Modifier
                                .fillMaxWidth()
                                .background(MaterialTheme.colorScheme.primaryContainer)
                                .padding(10.dp)
                            ) {
                                Text("Title", color = MaterialTheme.colorScheme.onPrimaryContainer)
                            }
                        },
                        content = { padding ->
                            LazyColumn(modifier = Modifier.padding(padding)) {
                                items(5) {
                                    Text(text ="Item = $it")
                                }
                            }
                        }
                    )
                }
            }
        }
        paparazzi.snapshot(
            name = "Scaffold",
            composable = viewToCapture
        )
    }
}
```

**Expected behavior**
I would expect the accessibility snapshot to print in the same order that talkback announces it.

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: 15
- Compile SDK: 35
- Gradle Version: 8.11.1
- Android Gradle Plugin Version: 8.9.0

**Screenshots**
Screenshot of output:
![Image](https://github.com/user-attachments/assets/79588b27-c115-4b77-bb5d-bdc2e93d34bd)
"Title" is showing at the bottom, when I'd expect it to be added at the top.

jasonctoms:
We have also noticed this issue! We added accessibility screenshots and thought we had way more TalkBack bugs until we confirmed that TalkBack was reading things in the correct order.

PR #2019, opened 2025-06-27
Handle SemanticsNode traversalIndices to support Compose accessibility reordering

colinmarsch:
This pull request introduces enhancements to accessibility rendering in the Paparazzi library, focusing on improving the handling of traversal order in semantic nodes. The changes include adding a new method to sort semantic nodes based on traversal index, updating the rendering logic to use this ordering, and introducing tests to validate the traversal order functionality.

Similar reordering support will be needed for View's as well based on their `getAccessibilityTraversalAfter` and `getAccessibilityTraversalBefore` values.

Sadly even with this added functionality, it seems that https://github.com/cashapp/paparazzi/issues/1902 is still an occurring issue in my testing. I will continue to dig into what might be the root cause there.


geoff-powell:
In addition to this, we can add support for `isTraversalGroup` as well to provide better ordering.
Also `zIndex` affects ordering of components which would be another feature to support alongside this.

https://developer.android.com/develop/ui/compose/accessibility/traversal#api-considerations

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -381,6 +413,17 @@ public class AccessibilityRenderExtension : RenderExtension {
     private const val LIVE_REGION_POLITE_LABEL = "polite"
     private const val EDITABLE_LABEL = "<editable>"
     private const val IN_LIST_LABEL = "<in-list>"
+
+    data class SemanticsNodeTraversalEntry(
+      val traversalIndex: Float?,
```
geoff-powell: traversalIndex should default to 0 for all elements. We might want to consider this instead of having a nullable Float.
Cash uses `traversalIndex = -1` to force elements to be read first. We should ensure this works as expected in the test

colinmarsch:
> In addition to this, we can add support for `isTraversalGroup` as well to provide better ordering. Also `zIndex` affects ordering of components which would be another feature to support alongside this.
> 
> https://developer.android.com/develop/ui/compose/accessibility/traversal#api-considerations

`isTraversalGroup` is also supported in this PR at [this line](https://github.com/cashapp/paparazzi/pull/2019/files#diff-0bc98113821b7e2e7ff262cb8b941bf195070e83c737ae82451b717066e5cbffR134). I agree that `zIndex` support here would be good, however I think adding it in a separate PR would be my preference as it is not accessible from the `SemanticsNode` and might require a bit more wrangling to pull in to the ordering logic here.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
colinmarsch: I'll try this out and update the test case to make sure negative traversalIndices are supported properly!

colinmarsch: Just updated this PR to include the default 0 value. Thanks for pointing this out, it actually was a bug before. The order of the focusable elements in the modal bottom sheet changed as a result of implementing this behaviour correctly.

geoff-powell:
> > In addition to this, we can add support for `isTraversalGroup` as well to provide better ordering. Also `zIndex` affects ordering of components which would be another feature to support alongside this.
> > https://developer.android.com/develop/ui/compose/accessibility/traversal#api-considerations
> 
> `isTraversalGroup` is also supported in this PR at [this line](https://github.com/cashapp/paparazzi/pull/2019/files#diff-0bc98113821b7e2e7ff262cb8b941bf195070e83c737ae82451b717066e5cbffR134). I agree that `zIndex` support here would be good, however I think adding it in a separate PR would be my preference as it is not accessible from the `SemanticsNode` and might require a bit more wrangling to pull in to the ordering logic here.

Ahh yes I missed that change, ignore my previous comment on `isTraversalGroup`.

+1 to look into `zIndex` in separate PR. Its set via a modifier so we will have to do some more hands on compose work to pull that information during accessibility text gathering.

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/java/app/cash/paparazzi/plugin/test/AccessibilityRenderingTest.kt
```diff
@@ -368,6 +369,28 @@ class AccessibilityRenderingTest {
     paparazzi.snapshot(listView)
   }
 
+  @Test
+  fun `verify traversalIndex order`() {
+    paparazzi.snapshot {
+      Column {
```
geoff-powell, line 375: nit: does `Column` semantics naturally set `isTraversalGroup` to true?

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/snapshots/images/app.cash.paparazzi.plugin.test_AccessibilityRenderingTest_modalBottomSheetMaterial3.png
geoff-powell, line 1: nice! This is a great result of this change where bottom sheet accessibility node order is correctly reported 🎉

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/java/app/cash/paparazzi/plugin/test/AccessibilityRenderingTest.kt
colinmarsch: I don't think it has that set by default. For this small example test case I don't think setting it to `true` here would change any behaviour though. It might be worth adding a more complex test case here in the near future. I could try to do that when I implement the View order traversal support (after/before view id)

PR #2038, opened 2025-07-11
Do not cache paparazzi-gradle-plugin:test

colinmarsch:
I noticed that `paparazzi-gradle-plugin:test` was always being pulled `FROM_CACHE` and never actually run in CI, no matter what files were updated. This change causes it to always be run and never cached. If anyone has ideas for a better fix here to resolve the root cause that might have been resulting in that task not being cached I'm open to changing this!

### DIFF
--- a/paparazzi-gradle-plugin/build.gradle
+++ b/paparazzi-gradle-plugin/build.gradle
@@ -63,6 +63,7 @@ buildConfig {
 
 tasks.withType(Test).configureEach {
   dependsOn(':paparazzi:publishMavenPublicationToProjectLocalMavenRepository')
+  outputs.cacheIf { false }
 }
 
 // When cleaning this project, we also want to clean the test projects.


SimonMarquis:
I don't think this is true. 
For instance, in this PR updating a dependency: https://github.com/cashapp/paparazzi/actions/runs/16231231629/job/45834302667#step:5:766

```
> Task :paparazzi-gradle-plugin:test

PaparazziPluginTest > verifyFailureMultiModule PASSED

PaparazziPluginTest > prepareResourcesCaching PASSED
```

Imo Gradle is working as expected here. If the inputs to the test task did not change, Gradle will pull the results from the cache.

geoff-powell:
@colinmarsch Do you mind providing insights into what builds you saw this occur. If there is an issue, we should keep task cachability. It might be that the inputs aren't correctly set causing the caching issue as @SimonMarquis mentions

colinmarsch:
> @colinmarsch Do you mind providing insights into what builds you saw this occur. If there is an issue, we should keep task cachability. It might be that the inputs aren't correctly set causing the caching issue as @SimonMarquis mentions

For a few examples I see these three builds all report `paparazzi-gradle-plugin:test FROM-CACHE`: [1](https://github.com/cashapp/paparazzi/actions/runs/16227528337/job/45822811444), [2](https://github.com/cashapp/paparazzi/actions/runs/16027677142/job/45219743196), [3](https://github.com/cashapp/paparazzi/actions/runs/15978905631/job/45068410260). For each of these builds there were changes in the main `paparazzi` module (to `AccessibilityRenderExtension` and other files) that impacts the `paparazzi-gradle-plugin:test` run. I'm not too familiar with how gradle task inputs are setup but maybe something needs to change there to ensure the gradle plugin tests are re-run if there are changes in the `paparazzi` module?

SimonMarquis:
I don't think the `:paparazzi-gradle-plugin` project is supposed to directly depends on the `:paparazzi` project.

colinmarsch:
> I don't think the `:paparazzi-gradle-plugin` project is supposed to directly depends on the `:paparazzi` project.

That's fair and I think that is the right choice here. Maybe a better fix is to move more of the test cases that involve the actual Paparazzi snapshotting logic from the `paparazzi-gradle-plugin` module to the `sample` module, which already has it's tests run on CI for `paparazzi` changes. cc @geoff-powell @jrodbx ?

PR #2048, opened 2025-07-18
Move AccessibilityRenderingTest cases into sample module

colinmarsch:
Moves all of the `AccessibilityRenderingTest` test cases from the `paparazzi-gradle-plugin` module into the `sample` module. This is more relevant as they are test cases verifying the output of the snapshot images and is another approach to get around the issue that I mentioned in [this PR](https://github.com/cashapp/paparazzi/pull/2038).


colinmarsch:
Hmm it looks like I'm running into [this issue](https://github.com/cashapp/paparazzi/issues/955) on this PR. Commenting out `apply plugin: 'jacoco'` in the sample/build.gradle seems to resolve this locally for me.

`java.lang.UnsupportedOperationException: class redefinition failed: attempted to delete a method`

jrodbx:
I'm not following; this change isn't a solution for the Gradle caching issue highlighted in [#2038](https://github.com/cashapp/paparazzi/pull/2038).  FWIW, I'd also prefer to keep these tests where they are.

Geoff and I chatted offline about the underlying issue in https://github.com/cashapp/paparazzi/pull/2038.  In that chat, I theorized whether it might be related to this in the plugin:

https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L372-L376

i.e., code changes in the paparazzi module that aren't ABI changes do not trigger recompilations during test runs, resulting in CACHED result.  If that theory is accurate, we could try appending a SHA version in that case (although I'm not sure I want to endorse this, just thinking aloud).

To further investigate, I would try setting up a plugin test that simulates a "internal" dependency bump to an "internal" version of Paparazzi that highlights whatever issue you were facing and run with `-Dorg.gradle.caching.debug=true` per the [docs](https://docs.gradle.org/current/userguide/build_cache_debugging.html#helpful_data_for_diagnosing_a_cache_miss) and observing changes in task inputs/outputs.

jrodbx (commented):
This change isn't a solution for the Gradle caching issue highlighted in [#2038](https://github.com/cashapp/paparazzi/pull/2038).  FWIW, I'd also prefer to keep these tests where they are.

Geoff and I chatted offline about the underlying issue in https://github.com/cashapp/paparazzi/pull/2038.  In that chat, we theorized whether it might be related to [this logic in the plugin](https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L372-L376), i.e., code changes in the paparazzi module that aren't ABI changes do not trigger recompilations during test runs, resulting in CACHED result.  If that theory is accurate, we could try appending a SHA version in that case (although I'm not sure I want to endorse this, just thinking aloud).

To further investigate, maybe try setting up a plugin test that simulates a "internal" dependency bump to an "internal" version of Paparazzi that highlights whatever issue you were facing and run with `-Dorg.gradle.caching.debug=true` per the [docs](https://docs.gradle.org/current/userguide/build_cache_debugging.html#helpful_data_for_diagnosing_a_cache_miss) and observing changes in task inputs/outputs.

colinmarsch:
> This change isn't a solution for the Gradle caching issue highlighted in [#2038](https://github.com/cashapp/paparazzi/pull/2038). FWIW, I'd also prefer to keep these tests where they are.
> 
> Geoff and I chatted offline about the underlying issue in #2038. In that chat, we theorized whether it might be related to [this logic in the plugin](https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L372-L376), i.e., code changes in the paparazzi module that aren't ABI changes do not trigger recompilations during test runs, resulting in CACHED result. If that theory is accurate, we could try appending a SHA version in that case (although I'm not sure I want to endorse this, just thinking aloud).
> 
> To further investigate, maybe try setting up a plugin test that simulates a "internal" dependency bump to an "internal" version of Paparazzi that highlights whatever issue you were facing and run with `-Dorg.gradle.caching.debug=true` per the [docs](https://docs.gradle.org/current/userguide/build_cache_debugging.html#helpful_data_for_diagnosing_a_cache_miss) and observing changes in task inputs/outputs.

Yeah, this change isn't really a fix for the caching issue, more so an attempt at a workaround. I'll close this PR and try the suggestions you gave here!

PR #2051, opened 2025-07-22
Fix reporting in list when item doesn't have any accessibility identifiers

geoff-powell:
Repro screenshot:

addacc6507c0573f60a8ae2255227f81696048a6


colinmarsch (approved):
Thanks for doing this!

PR #2053, opened 2025-07-24
Fix in-list reporting for nested nodes for accessibility

geoff-powell:


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -193,6 +193,7 @@ public class AccessibilityRenderExtension : RenderExtension {
           .mapNotNull { it.accessibilityText() }
           .joinToString(", ")
           .ifEmpty { null }
+          .takeIf { it != IN_LIST_LABEL }
```
geoff-powell, line 196: this disallows any nodes to not have exactly `<in-list>`

colinmarsch (approved):
Is there a snapshot that failed because of this? If not maybe we should add one to ensure a regression doesn't slip back in?

geoff-powell:
> Is there a snapshot that failed because of this? If not maybe we should add one to ensure a regression doesn't slip back in?

Yes I updated the list test to catch this exact case. Let me split commits up to showcase this

geoff-powell:
@colinmarsch [Repro case](https://github.com/cashapp/paparazzi/pull/2053/commits/32eb4d844b173e3e254bf773d432a82a38a336dd): Build https://github.com/cashapp/paparazzi/actions/runs/16503506612

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -193,6 +193,7 @@ public class AccessibilityRenderExtension : RenderExtension {
           .mapNotNull { it.accessibilityText() }
           .joinToString(", ")
           .ifEmpty { null }
+//          .takeIf { it != IN_LIST_LABEL }
```
colinmarsch: Should this be commented out?

geoff-powell: nope, didn't push my last commit

PR #2067, opened 2025-08-12
Add support for overwriting snapshots when max percent diff threshold is reached

### LINKED ISSUE #1951: Record new screenshots only if changes exceed the `maxPercentDifference` option

geoff-powell:
Fixes #1951


colinmarsch (approved):
Cool! Nice to see this functionality!

paparazzi-gradle-plugin/src/test/projects/overwrite-on-max-percent-difference/gradle.properties
```diff
@@ -0,0 +1,4 @@
+android.useAndroidX=true
+android.nonTransitiveRClass=false
+
+app.cash.paparazzi.overwriteOnMaxPercentDifference=true
```
yogurtearl, line 4: @geoff-powell with the current name it is not clear if the type of the flag is a `boolean` or a `float`

i.e. someone might (incorrectly) try this: 
`app.cash.paparazzi.overwriteOnMaxPercentDifference=0.01`

If we call it something like `overwriteOnlyOnMaxPercentDifferenceOrGreater` it makes it clearer that it is a boolean.

Issue #2027, opened 2025-07-03
Custom font rendering in snapshots on Windows using default font (since 1.3.5)


appsoluut:
**Description**
After upgrading to 1.3.5 (from 1.3.4), the snapshots produced using Compose are using the default font instead of the custom font, which triggers all tests to fail and being different from the Previews and when launching the app.

This happens only on Windows.

I have tested against the latest 2.0.0-alpha (currently 02) version of Paparazzi too and the issue remains.

What I did notice that it is working when recording the snapshots on Linux or Mac. I have tried to use Windows WSL with the same settings and in there it also seems to work. My guess would be that the change (from 1.3.4 to 1.3.5) that fixed the `loadFont()` path broke something for Windows file structures.

**Steps to Reproduce**
Record a snapshot with a custom set font / textstyle and see the regular font (Arial?) being used instead.

**Expected behavior**
Using the correct font on Windows too.

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: Windows 11
- Compile SDK: 35
- Gradle Version: 8.14.2
- Android Gradle Plugin Version: 8.11.0

**Screenshots**
**Snapshot in 1.3.4 (Correct)**
![Snapshot in 1.3.4](https://github.com/user-attachments/assets/7426e39d-3019-4f9d-a3cd-5b250af2a1d9)

**Snapshot in 1.3.5 (Wrong)**
![Snapshot in 1.3.5](https://github.com/user-attachments/assets/8f147413-1008-4ef4-800b-5e3909c72d7f)

Notice the font difference in the `"S"` and `"C"` characters on the "Secondary" button style.

geoff-powell:
Hmm, yes we had a similar issue with loading the font path for windows since the file separator differs from Mac / Linux. 
This [ASM](https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/instrumentation/ResourcesCompatVisitorFactory.kt) change is supposed to fix font loading issues on windows but it seems like something isn't working correctly.

I can take a look. In the meantime, do you mind providing a sample project?

appsoluut:
Sure! I'll make a small sample project and report back to you! 👍

appsoluut:
Just a heads up, I didn't forget about this one. But when creating the small sample project, it is working with the custom font. So I'm trying to figure out what the difference is between the sample project and the actual project.

efemoney:
[...]

Issue #1701, opened 2024-11-19
Could not load font in multi module project on Windows (since 1.3.5)


tim-ebner-exxeta:
**Description**
Trying to use a custom font resource added in a second module on Windows with Paparazzi version 1.3.5 leads to following error:
```
java.lang.IllegalStateException: Could not load font
	at androidx.compose.ui.text.font.TypefaceRequestCache.runCached(FontFamilyResolver.kt:207)
	at androidx.compose.ui.text.font.FontFamilyResolverImpl.resolve(FontFamilyResolver.kt:92)
....
Caused by: java.lang.IllegalStateException: Unable to load font ResourceFont(resId=2131296256, weight=FontWeight(weight=400), style=Normal, loadingStrategy=Blocking)
	at androidx.compose.ui.text.font.FontListFontFamilyTypefaceAdapterKt.firstImmediatelyAvailable(FontListFontFamilyTypefaceAdapter.kt:193)
	at androidx.compose.ui.text.font.FontListFontFamilyTypefaceAdapterKt.access$firstImmediatelyAvailable(FontListFontFamilyTypefaceAdapter.kt:1)
...
Caused by: android.content.res.Resources$NotFoundException: Font resource ID #0x7f090000 could not be retrieved.
	at app//androidx.core.content.res.ResourcesCompat.loadFont(ResourcesCompat.java:572)
	at app//androidx.core.content.res.ResourcesCompat.getFont(ResourcesCompat.java:413)
...
```

The error only appears on Windows. With Linux it works.
It also worked with version 1.3.3 and 1.3.4 on Windows

**Steps to Reproduce**
On you Windows Machine:
- Create a new Android project with Paparazzi 1.3.5
- Add a second module with a custom Font with Typography file where font is used
- Import he second module in your first module and use the typography
- Run Paparazzi test in first module

Minimal Example:
[MinimalExample.zip](https://github.com/user-attachments/files/17817731/MinimalExample.zip)

**Expected behavior**
I don't get an error when using a custom font.

**Additional information:**
- Paparazzi Version: **1.3.5**
- OS: **Windows**
- Compile SDK:  35
- Gradle Version: 8.0
- Android Gradle Plugin Version: 8.7.2

Thanks & Best Regards
Tim

geoff-powell:
Verified issue is occuring on Mac OS as well. 

Thanks for reporting!

geoff-powell:
Looks like the file path that is failing is something like this

[...]

PR #2074, opened 2025-08-21
Repro issue with font loading on Windows

geoff-powell:
Uses AndroidStudio ASM fix to correctly fix ResourceCompat.loadFont
https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/classloading/ResourcesCompatTransform.kt;l=64-82

Repro of issue: https://github.com/cashapp/paparazzi/pull/2074/commits/dee0ed9f25b5d8309623b8a509d653917a9765b8

Fixes:
https://github.com/cashapp/paparazzi/issues/2027
#1701


jrodbx (approved):
nice!

Issue #2075, opened 2025-08-25
Some similar snapshots falsely pass even with maxPercentDifference set to zero


matejdro:
**Description**

It seems that in some cases, Paparazzi will ignore the `maxPercentDifference = 0.0` and will falsely pass the tests, even when the recorded image differs from the new image.

This seems to be happening when the result of the comparion is the `DiffResult.Similar`. It appears percentage is not forwarded in this case, but always hardcoded as zero: https://github.com/cashapp/paparazzi/blob/ca941225754abc3f768e3bf1378bdbf5c6a514b2/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt#L144

**Steps to Reproduce**

1. Record the snapshot with the following test:

```kotlin
class BugDemo {

    @get:Rule
    val paparazzi: Paparazzi = Paparazzi(deviceConfig = DeviceConfig.NEXUS_5, maxPercentDifference = 0.0)

    @Test
    fun test() {
        paparazzi.snapshot {
            Box(Modifier.size(1024.dp).background(Color.Black).padding(100.dp)) {
                Box(Modifier.size(400.dp, 1.dp).background(Color.White))
            }
        }
    }
}
```

2. Change the width of the inner box from `400.dp` to the `390.dp`
3. Run the verify task

**Expected behavior**

When `maxPercentDifference` is set to `0.0`, then any change in the image should fail the verification, no matter how small

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: Linux
- Compile SDK: 35
- Gradle Version: 8.11.1
- Android Gradle Plugin Version: 8.10.0

**Screenshots**

<img width="562" height="1000" alt="Image" src="https://github.com/user-attachments/assets/b004fecd-da33-4fcf-a93c-9381379e8b1c" />

Paparazzi does not detect line slightly shrinking in the above image

geoff-powell:
So when testing this in paparazzi and preview, it is working as Preview works.

For whatever reason this composable you are testing doesn't actually change the output image when changing from 
400 -> 390.

This is due because `NEXUS_5` device pixel size is 1080x1920.

If the outer Box is `1024.dp` then that would be way too large.

The following test fails as expected

```
class BugDemo {

  @get:Rule
  val paparazzi: Paparazzi = Paparazzi(deviceConfig = DeviceConfig.NEXUS_5, maxPercentDifference = 0.0)

  @Test
  fun test() {
    paparazzi.snapshot {
[...]

PR #2078, opened 2025-08-26
Fix differ comparison between black pixels with 100% alpha and black pixels with 0% alpha

geoff-powell:
#2075

Commit: https://github.com/cashapp/paparazzi/pull/2078/commits/1a42e0e14cd3b0f5fc9ad216ca4cbf887381e105 reproduces the issue and tests end up passing

### DIFF
--- a/paparazzi/api/paparazzi.api
+++ b/paparazzi/api/paparazzi.api
@@ -119,11 +119,11 @@ public final class app/cash/paparazzi/Flags {
 
 public final class app/cash/paparazzi/HtmlReportWriter : app/cash/paparazzi/SnapshotHandler {
 	public static final field $stable I
-	public fun <init> (DLapp/cash/paparazzi/internal/Differ;)V
-	public fun <init> (Ljava/lang/String;DLapp/cash/paparazzi/internal/Differ;)V
-	public fun <init> (Ljava/lang/String;Ljava/io/File;DLapp/cash/paparazzi/internal/Differ;)V
-	public fun <init> (Ljava/lang/String;Ljava/io/File;DLapp/cash/paparazzi/internal/Differ;Ljava/io/File;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/io/File;DLapp/cash/paparazzi/internal/Differ;Ljava/io/File;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (DLapp/cash/paparazzi/Differ;)V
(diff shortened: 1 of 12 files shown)


paparazzi/src/test/java/app/cash/paparazzi/internal/PixelPerfectTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test
+import java.awt.image.BufferedImage
+import java.awt.image.BufferedImage.TYPE_INT_ARGB
+
+class PixelPerfectTest {
+  @Test
+  fun `compare identical images`() {
+    val expected = BufferedImage(1, 1, TYPE_INT_ARGB)
+    val actual = BufferedImage(1, 1, TYPE_INT_ARGB)
+    val result = PixelPerfect.compare(expected, actual)
```
jrodbx: nice tests

paparazzi/src/main/java/app/cash/paparazzi/internal/PixelPerfect.kt
```diff
@@ -78,7 +78,13 @@ internal object PixelPerfect : Differ {
...
     // 3 different colors, 256 color levels
     val total = actualHeight.toLong() * actualWidth.toLong() * 3L * 256L
-    val percentDifference = (delta * 100 / total.toDouble()).toFloat()
+    var percentDifference = (delta * 100 / total.toDouble()).toFloat()
+
+    // If the delta diff is all black pixels, the computed difference is 0, but there are still
+    // different pixels. We can fallback to the amount of different pixels to less precise difference to ensure difference is reported.
+    if (differentPixels > 0 && percentDifference == 0f) {
+      percentDifference = differentPixels * 100 / (actualWidth * actualHeight.toDouble()).toFloat()
+    }
```
jrodbx: nit: it looks like we can avoid the `var` here and above, if the comparison uses `delta` (the numerator in percentDifference) instead.

```suggestion
    // If the delta diff is all black pixels, the computed difference is 0, but there are still
    // different pixels. We can fallback to the amount of different pixels to less precise difference to ensure difference is reported.
    val percentDifference = if (delta == 0 && differentPixels > 0) {
differentPixels * 100 / (actualWidth * actualHeight.toDouble()).toFloat()
    } else {
      (delta * 100 / total.toDouble()).toFloat()
    }
```

jrodbx (approved):

Issue #2093, opened 2025-09-08
Modifier.graphicsLayer {  alpha = 0f } nodes are still visible according to the A11y renderer


FrancoisBlavoet:
**Description**
If you make a composable invisible via `graphicsLayer` : 

```kt
Text(
            modifier = Modifier.graphicsLayer {
              alpha = 0f
              scaleX = 0.5f
            },
            text = "Text with zero alpha via graphicsLayer"
          )
```

Then Paparazzi's a11y renderer will still render it as if its alpha were not 0f
I am unsure of how Talkback does it, but it is able to discard such a node.
I have looked at how to access this data through the SemanticsNode, but no luck so far, there is SemanticsNode.isTransparent() but it is internal, and even calling it via reflection did not work for me

**Steps to Reproduce**
see above, I can also open a branch with an update to `AccessibilityRenderingTest` if that helps

**Expected behavior**
correctly hide this view from the a11y renderer

**Additional information:**
- Paparazzi Version: 2.0.0-alpha02
- OS: Mac
- Compile SDK: 34 
- Gradle Version: 8.10
- Android Gradle Plugin Version: 8.13
 
**Screenshots**

<img width="516" height="482" alt="Image" src="https://github.com/user-attachments/assets/2e39a877-fe92-4a12-a04e-04758c2aa545" />

FrancoisBlavoet:
nevermind, I have a very ugly fix .. I will open up a PR

FrancoisBlavoet:
I have opened https://github.com/cashapp/paparazzi/pull/2094

PR #2094, opened 2025-09-08
A11y renderer: handle nodes with alpha=0f  or HideFromAccessibility

FrancoisBlavoet:
Ugly fix for https://github.com/cashapp/paparazzi/issues/2093

The current approach to check for `modifier == Modifier.alpha(0f)` only works if you use this exact modifier. 
In particular, `Modifier.graphicsLayer { alpha = 0f }` won't match this check.

Unfortunately, Compose does not really expose anything that we could use to identify a transparent node.
There is however an internal function allowing us to detect this. 
Ideally we would get the compose team to make it public (or otherwise expose this information another way), but in the mean time I have written this ugly reflective way to access this information

This also adds a check for `HideFromAccessibility` since this is a new replacement for `InvisibleToUser`, we need to check for that one too.


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -349,6 +344,27 @@ public class AccessibilityRenderExtension : RenderExtension {
     )
   }
 
+  private fun SemanticsNode.hasZeroAlpha(): Boolean {
+    // try to reflectively query internal SemanticsNode.isTransparent() by iterating on kotlin field mangled names:
+    val candidateGetterNames = arrayOf(
+      "isTransparent",
```
FrancoisBlavoet: This is very ugly, unless there is an alternative way to get this info, we should get Compose to expose this property

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -349,6 +344,27 @@ public class AccessibilityRenderExtension : RenderExtension {
...
+      "isTransparent\$ui_release",
+      "getIsTransparent",
+      "getIsTransparent\$ui_release"
+    )
+    for (name in candidateGetterNames) {
```
TWiStErRob: Feels like this is near inner loop; executed on all nodes. I should be possible to cache this reflection filter query (companion object by lazy? because it's not dependent on any specific node) and only doing the invoke in hasZeroAlpha.

FrancoisBlavoet: fair enough, we now cache the resolved function

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/snapshots/images/app.cash.paparazzi.plugin.test_AccessibilityRenderingTest_dropDownMaterial3.png
FrancoisBlavoet, line 1: this dropdown is detected as transparent. That's not ideal, but the "before" screenshot is also incorrect and does not display the dropdown content, so the after is technically correct.

This happens because DropdownMenu is a `Popup` and AFAIK paparazzi can only render content in the main window 
The ideal state would be to display both windows, but that's out of the scope of this PR.

FrancoisBlavoet:
@TWiStErRob no hurry, but could I get another review on this? Ideally, if that PR looks good, then it would be nice to get it included in alpha3 🙏

colinmarsch:
I think we can resolve this issue without reflection on the `isTransparent` field through the use of checking `AccessibilityNodeInfo.isVisibleToUser` which internally in Compose is set via `info.isVisibleToUser = !semanticsNode.isHidden` here: https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidComposeViewAccessibilityDelegateCompat.android.kt#815. This takes into account the `isTransparent` state as well as both of `InvisibleToUser` and `HideFromAccessibility` per https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/semantics/SemanticsOwner.kt#140.

Could we try that against your repro here?

FrancoisBlavoet:
> I think we can resolve this issue without reflection on the `isTransparent` field through the use of checking `AccessibilityNodeInfo.isVisibleToUser` which internally in Compose is set via `info.isVisibleToUser = !semanticsNode.isHidden` here: https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidComposeViewAccessibilityDelegateCompat.android.kt#815. This takes into account the `isTransparent` state as well as both of `InvisibleToUser` and `HideFromAccessibility` per https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/semantics/SemanticsOwner.kt#140.
> 
> Could we try that against your repro here?

Thanks for the suggestion!
I implemented it, but I found that unfortunately it doesn't work. IsVisibleToUser always returns true in Paparazzi context, whether the node has alpha equal to 0 or not.
It correctly covers `InvisibleToUser` or `HideFromAccessibility`, but we can already query semantics node for these cases.

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/snapshots/images/app.cash.paparazzi.plugin.test_AccessibilityRenderingTest_dropDownMaterial3.png
FrancoisBlavoet: ^ Actually after looking at the code, the Dropdown uses an animatedContent animation and the frame we capture here has alpha = 0

so while I don't think this is what the test wants to capture, this is the correct a11y for that hierarchy

PR #2106, opened 2025-09-17
[DNM] AGP 9.0 API Changes

geoff-powell:


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -132,30 +145,44 @@ public class PaparazziPlugin @Inject constructor(
...
+       * 	at com.android.build.gradle.internal.dsl.InstrumentationImpl.getRegisteredDependenciesClassesVisitors(InstrumentationImpl.kt:72)
+       * 	at com.android.build.api.component.impl.features.InstrumentationCreationConfigImpl.getRegisteredDependenciesClassesVisitors(InstrumentationCreationConfigImpl.kt:118)
+       * 	at com.android.build.api.component.impl.features.InstrumentationCreationConfigImpl.getDependenciesClassesAreInstrumented(InstrumentationCreationConfigImpl.kt:56)
+       * 	at com.android.build.gradle.internal.dependency.AsmClassesTransform$Companion.registerAsmTransformForComponent(AsmClassesTransform.kt:68)
+       * 	at com.android.build.gradle.internal.DependencyConfigurator.configureVariantTransforms(DependencyConfigurator.kt:895)
+       * 	at com.android.build.gradle.internal.plugins.KotlinMultiplatformAndroidPlugin.afterEvaluate(KotlinMultiplatformAndroidPlugin.kt:427)
+       * 	at com.android.build.gradle.internal.plugins.KotlinMultiplatformAndroidPlugin.access$afterEvaluate(KotlinMultiplatformAndroidPlugin.kt:127)
+       */
+      if (!isMultiplatform) {
```
geoff-powell, line 167: need to log bug in IssueTracker. Android KMP plugin doesn't appropriately apply instrumentation correctly.

build.gradle
```diff
@@ -57,9 +60,9 @@ subprojects {
     targetCompatibility = libs.versions.javaTarget.get()
   }
 
-  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile).configureEach {
-    kotlinOptions {
-      jvmTarget = libs.versions.javaTarget.get()
+  tasks.withType(KotlinJvmCompile).configureEach {
```
TWiStErRob, line 69: is there a reason why this is not:
```suggestion
  kotlin {
```
like in all the other Gradle files changed?
(One thing that's better about the `kotlin {` is that we don't need to worry about KSP and other compiler tasks that may or may not pick this number up.)

PR #2112, opened 2025-09-22
Add support for Gradle 9.1.0

geoff-powell:


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
```diff
@@ -0,0 +1,16 @@
...
+internal class CodePanelRenderer : ReportRenderer<CodePanelRenderer.Data, SimpleHtmlWriter>() {
+  override fun render(data: Data, htmlWriter: SimpleHtmlWriter) {
+    htmlWriter.startElement("span").attribute("class", "code").startElement("pre").attribute("id", data.codePanelId)
+      .characters(data.text).endElement()
+    HtmlWriterTools.addClipboardCopyButton(htmlWriter, data.codePanelId)
+    htmlWriter.endElement()
+  }
+
+  data class Data(val text: String, val codePanelId: String)
+}
```
TWiStErRob:
```suggestion
internal class CodePanelRenderer : ReportRenderer<String, SimpleHtmlWriter>() {
  override fun render(text: String, htmlWriter: SimpleHtmlWriter) {
    htmlWriter.startElement("span").attribute("class", "code")
      .startElement("pre").characters(text).endElement();
    HtmlWriterTools.addClipboardCopyButton(htmlWriter);
    htmlWriter.endElement();
  }
}
```

see the left side of https://github.com/gradle/gradle/commit/1b56084b4717d37625394124cc2a3e5806c5129b (the other classes around this are expecting the behavior pre Gradle 9.1?)

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
```diff
@@ -0,0 +1,34 @@
...
+   */
+  @Throws(IOException::class)
+  private fun addClipboardCopyButton(writer: SimpleHtmlWriter): SimpleMarkupWriter {
+    return writer
+      .startElement("button")
+      .attribute("class", "clipboard-copy-btn")
+      .attribute("aria-label", "Copy to clipboard")
+      .characters("Copy")
+      .endElement()
+  }
```
octylFractal: This won't work. I don't know why you copied the old version of this code, but the new version (and the change that caused the breakage in the first place) provides `data-copy-element-id` which is used by our [`report.js`](https://github.com/gradle/gradle/blob/1b56084b4717d37625394124cc2a3e5806c5129b/subprojects/core/src/main/resources/org/gradle/reporting/report.js#L70). If you add this button without it but aren't emitting different JS, the button won't do anything. You may as well not emit it.

Personally, I think it would be preferable to provide a proper ID and emit it, but it's up to you.

ethancookeyouitv:
What is holding back merging this support for Gradle 9.1?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ClassPageRenderer.kt
```diff
@@ -117,7 +116,10 @@ internal class ClassPageRenderer(
           } else {
             failure.stackTrace
           }
-        codePanelRenderer.render(message, htmlWriter)
+        codePanelRenderer.render(
+          data = CodePanelRenderer.Data(text = message, codePanelId = "${test.id}-failure-$index-stacktrace"),
```
jrodbx: why add a second key (index) vs solely test id, per: https://github.com/gradle/gradle/commit/1b56084b4717d37625394124cc2a3e5806c5129b#diff-8ddb302f6da511b2e6f5c1ffc5f9107ed687ca37ab84f243b2a9d3266b30d89cR113

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
```diff
@@ -0,0 +1,38 @@
...
+    return writer
+      .startElement("button")
+      .attribute("class", "clipboard-copy-btn")
+      .attribute("aria-label", "Copy to clipboard")
+      .attribute("data-copy-element-id", copyElementId)
+      .characters("Copy").endElement()
+  }
+
+  data class Data(val text: String, val codePanelId: String)
+}
```
jrodbx: Remove and reference the identical internal sources? I believe the next Gradle bumps will force migrate us to the new test reporting anyway, so it's time-bound to just point to the internals, and being a plugin

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -972,7 +970,7 @@ class PaparazziPluginTest {
     )
     assertThat(config.aarExplodedDirs)
       .comparingElementsUsing(MATCHES_PATTERN)
-      .containsExactly("^caches/[0-9]{1,2}.[0-9]{1,2}(.[0-9])?/transforms/[0-9a-f]{32}/transformed/external/res\$")
+      .containsExactly("^caches/[0-9]{1,2}.[0-9]{1,2}(.[0-9])?/transforms/[0-9a-f]{32}/(workspace/)?transformed/external/res\$")
```
jrodbx, line 962: sigh...windows....

https://github.com/gradle/gradle/blob/8d9cbbb408ff3c4cd499ac46c358cff1fc490227/platforms/core-execution/execution/src/main/java/org/gradle/internal/execution/workspace/ImmutableWorkspaceProvider.java#L30

jrodbx (commented):
Let's rebase and target latest master?  Also, let's run `./gradlew wrapper` and commit the updated files.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
geoff-powell: :sads:

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ClassPageRenderer.kt
geoff-powell: So there is a potential for multiple failures per test so wanted to capture this in the id

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
geoff-powell: Unfortunately its a private static class so I can't references the `text and codePanelId` params without reflection which seems like a lot of work.

jrodbx: which class are you referring to?  `CodePanelRenderer.Data` is public static.

https://github.com/gradle/gradle/blob/master/platforms/core-runtime/report-rendering/src/main/java/org/gradle/reporting/CodePanelRenderer.java#L27

so in renderFailures:

```
        codePanelRenderer.render(
          CodePanelRenderer.Data(message, "test-message-$testId"),
          htmlWriter
        )
```

should just work

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -87,6 +85,17 @@ class PaparazziPluginTest {
...
+  fun kotlinMultiplatformPluginWithNewAndroidLibraryPlugin() {
+    val fixtureRoot = File("src/test/projects/multiplatform-plugin-with-new-android-library-plugin")
+
+    val result = gradleRunner
+      .withArguments("preparePaparazziAndroidMainResources", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+
+    assertThat(result.task(":preparePaparazziAndroidMainResources")).isNotNull()
+  }
+
```
jrodbx: revert

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -72,7 +70,7 @@ class PaparazziPluginTest {
 
     assertThat(result.task(":preparePaparazziDebugResources")).isNull()
     assertThat(result.output).contains(
-      "One of com.android.application, com.android.library, com.android.dynamic-feature must be applied for Paparazzi to work properly."
+      "One of com.android.application, com.android.library, com.android.dynamic-feature, com.android.kotlin.multiplatform.library must be applied for Paparazzi to work properly."
```
jrodbx: revert

paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-new-android-library-plugin/build.gradle
jrodbx, line 1: this test project can be deleted

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx, line 1: everything here can be reverted

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ClassPageRenderer.kt
jrodbx: as in multiple failing calls to `.snapshot` or `.gif`?  As an example, this test fails fast and only shows the first exception:

```
class CrashTest {

  @Test
  fun compose() {
    paparazzi.snapshot { Crash() }
    paparazzi.snapshot { Crash() }
  }
}

@Composable fun Crash() { throw Exception("Test exception for screenshot") }
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
geoff-powell: Sorry I meant the inner fields of `CodePanelRenderer.Data` are private and can't be accessed.
Are you suggesting instead of create our duplicate `CodePanelRenderer` we just use gradle's versions?

jrodbx (approved):
added a commit to add the clipboard copy buttons like Gradle does

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/CodePanelRenderer.kt
jrodbx: yes!

PR #2113, opened 2025-09-23
Ignore a missing androidx.legacy.coreutils.R class

swankjesse:
This package started showing up in the list of packages that have resources, but it has never had resources. It's safe to ignore if this class is missing.


swankjesse:
Hang on, I think I might have fixed the symptom and not the problem. I found the missing `R.class` here:

```
jar tvf /Volumes/Development/paparazzi/sample/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/debugUnitTest/generateDebugUnitTestStubRFile/R.jar
  ...
   729 Thu Jan 01 01:01:02 EST 1981 androidx/legacy/coreutils/R.class
```

PR #2114, opened 2025-09-23
Remove use of base extension from AGP ahead of 9.0

geoff-powell:
Fixes issue with BaseExtension usage with newer AGP apis to resolve android extension


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -122,6 +122,12 @@ public class PaparazziPlugin @Inject constructor(
       it.delete(files)
     }
 
+    // We need to pull target sdk as defined from DSL otherwise it gets set to some default value when resolving during [onVariants]
+    var targetSdkVersion: String? = null
+    extension.finalizeDsl { androidExtension ->
+      targetSdkVersion = androidExtension?.targetSdkVersion() ?: DEFAULT_COMPILE_SDK_VERSION.toString()
+    }
+
```
geoff-powell, line 130: This is the bulk of the change where `BaseExtension` is replaced with this dsl extension handling.

Issue #472, opened 2022-06-24
lateinit property mergeResourcesTask has not been initialized


PaulWoitaschek:
**Description**
We have several gradle modules that are pure compose. Our project default is `android.defaults.buildfeatures.resvalues=false
`.
If you run paparrazi on a module without res values, it crashes in the configuration phase.

```stacktrace
Caused by: kotlin.UninitializedPropertyAccessException: lateinit property mergeResourcesTask has not been initialized
        at com.android.build.gradle.internal.scope.MutableTaskContainer.getMergeResourcesTask(MutableTaskContainer.kt:67)
        at com.android.build.gradle.internal.api.BaseVariantImpl.getMergeResourcesProvider(BaseVariantImpl.java:384)
        at com.android.build.gradle.internal.api.LibraryVariantImpl_Decorated.getMergeResourcesProvider(Unknown Source)
        at app.cash.paparazzi.gradle.PaparazziPlugin.setupPaparazzi$lambda-22(PaparazziPlugin.kt:68)
        at org.gradle.configuration.internal.DefaultUserCodeApplicationContext$CurrentApplication$1.execute(DefaultUserCodeApplicationContext.java:123)
        at org.gradle.api.internal.DefaultCollectionCallbackActionDecorator$BuildOperationEmittingAction$1.run(DefaultCollectionCallbackActionDecorator.java:110)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.run(DefaultBuildOperationRunner.java:47)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:68)
[...]

PR #2115, opened 2025-09-23
Add support for new android library multiplatform plugin

geoff-powell:
Ahead of AGP 9.0 removal of `com.android.library plugin` support with KMP`

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -22,28 +22,24 @@ import app.cash.paparazzi.gradle.utils.artifactViewFor
 import app.cash.paparazzi.gradle.utils.capitalize
 import app.cash.paparazzi.gradle.utils.relativize
 import com.android.build.api.component.analytics.AnalyticsEnabledComponent
+import com.android.build.api.component.analytics.AnalyticsEnabledLibraryVariant
 import com.android.build.api.dsl.CommonExtension
-import com.android.build.api.dsl.KotlinMultiplatformAndroidLibraryExtension
 import com.android.build.api.instrumentation.FramesComputationMode
 import com.android.build.api.instrumentation.InstrumentationScope
 import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.api.variant.ApplicationAndroidComponentsExtension
 import com.android.build.api.variant.Component
 import com.android.build.api.variant.DynamicFeatureAndroidComponentsExtension
-import com.android.build.api.variant.GeneratesApk
 import com.android.build.api.variant.HasHostTests
 import com.android.build.api.variant.HasUnitTest
-import com.android.build.api.variant.HasUnitTestBuilder
+import com.android.build.api.variant.HostTest
 import com.android.build.api.variant.HostTestBuilder
 import com.android.build.api.variant.KotlinMultiplatformAndroidComponentsExtension
 import com.android.build.api.variant.LibraryAndroidComponentsExtension
-import com.android.build.api.variant.LibraryVariant
 import com.android.build.api.variant.TestComponent
-import com.android.build.api.variant.Variant
 import com.android.build.api.variant.impl.HasHostTestsCreationConfig
-import com.android.build.api.variant.impl.HasTestSuitesCreationConfig
 import com.android.build.gradle.internal.component.TestCreationConfig
-import com.android.build.gradle.internal.component.VariantCreationConfig
+import com.android.builder.model.Version.ANDROID_GRADLE_PLUGIN_VERSION
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
@@ -110,69 +106,15 @@ public class PaparazziPlugin @Inject constructor(
           // exhaustive to avoid potential breaking changes in future AGP releases
           else -> error("${androidComponents.javaClass.name} from $plugin is not supported in Paparazzi")
         }
-        AndroidProjectScope(project = project, extension = androidComponents).setupPaparazzi()
+        project.setupPaparazzi(androidComponents)
       }
     }
   }
 
-  private class AndroidProjectScope(
-    val project: Project,
-    val extension: AndroidComponentsExtension<*, *, *>
-  ) {
-    val isMultiplatformProject: Boolean =
-      extension is KotlinMultiplatformAndroidComponentsExtension ||
-        project.plugins.hasPlugin(KOTLIN_MULTIPLATFORM_PLUGIN)
-
-    private fun Project.isInternal(): Boolean = providers.gradleProperty("app.cash.paparazzi.internal").orNull == "true"
-
-    fun addTestDependency() {
-      val dependency = if (project.isInternal()) {
-        project.dependencies.project(mapOf("path" to ":paparazzi"))
-      } else {
-        project.dependencies.create("app.cash.paparazzi:paparazzi:$VERSION")
-      }
-      val configurationName = if (isMultiplatformProject) "commonTestImplementation" else "testImplementation"
-      project.configurations.getByName(configurationName).dependencies.add(dependency)
-    }
-
-    fun snapshotDir(testComponent: TestComponent): Provider<Directory> {
-      val testSources = requireNotNull(testComponent.sources.java?.all ?: testComponent.sources.kotlin?.all) {
-        "No test sources found for test component: ${testComponent.name}"
-      }
-      return testSources.map { sourceDirs ->
-        val defaultTestDir = sourceDirs.first().asFile.parentFile
-        val defaultTestDirectory = project.layout.projectDirectory.dir(defaultTestDir.path)
-        (
-          sourceDirs.firstNotNullOfOrNull { sourceDir ->
-            // Sources usually in `/src/test/java/` so need to move to parent `/src/test/`
-            val parentFile = sourceDir.asFile.parentFile
-            val testFiles = sourceDir.asFileTree.relativize(project.layout.projectDirectory)
-            val isNonBuildDirectorySources =
-              parentFile.path.contains(project.layout.buildDirectory.get().asFile.path).not()
-            val containsTestFiles = testFiles.isPresent
-
-            if (isNonBuildDirectorySources && containsTestFiles) {
-              project.layout.projectDirectory.dir(parentFile.path)
-            } else {
-              null
-            }
-          } ?: defaultTestDirectory
-          ).dir("snapshots")
-      }
-    }
-
-    fun Provider<Boolean>.provideOnEnabled(provider: Provider<*>): Provider<*> =
-      flatMap { record ->
-        if (record) {
-          provider
-        } else {
-          project.objects.directoryProperty()
-        }
-      }
-  }
-
-  private fun AndroidProjectScope.setupPaparazzi() {
-    addTestDependency()
+  private fun Project.setupPaparazzi(extension: AndroidComponentsExtension<*, *, *>) {
+    val isMultiplatformProject: Boolean = extension is KotlinMultiplatformAndroidComponentsExtension ||
+      project.plugins.hasPlugin(KOTLIN_MULTIPLATFORM_PLUGIN)
+    addTestDependency(isMultiplatformProject)
 
     val layoutlibNativeRuntimeFileCollection = project.setupLayoutlibRuntimeDependency()
     val layoutlibResourcesFileCollection = project.setupLayoutlibResourcesDependency()
@@ -204,8 +146,8 @@ public class PaparazziPlugin @Inject constructor(
     extension.onVariants { variant ->
       val variantSlug = variant.name.capitalize()
       val testComponent = when (variant) {
-        is HasUnitTest -> (variant as HasUnitTest).unitTest
         is HasHostTests -> variant.hostTests[HostTestBuilder.UNIT_TEST_TYPE]
+        is HasUnitTest -> (variant as HasUnitTest).unitTest
         else -> null
       } ?: return@onVariants
       val snapshotOutputDir = snapshotDir(testComponent)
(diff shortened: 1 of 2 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -67,15 +69,21 @@ public class PaparazziPlugin @Inject constructor(
...
       check(supportedPlugins.any { project.plugins.hasPlugin(it) }) {
         "One of ${supportedPlugins.joinToString(", ")} must be applied for Paparazzi to work properly."
       }
+      val supportsNewAndroidKMPPlugin = project.plugins.hasPlugin(ANDROID_KOTLIN_MULTIPLATFORM_LIBRARY_PLUGIN)
       project.plugins.withId("org.jetbrains.kotlin.multiplatform") {
         val kmpExtension = project.extensions.getByType(KotlinMultiplatformExtension::class.java)
-        check(kmpExtension.targets.any { target -> target is KotlinAndroidTarget }) {
-          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin"
+        check(supportsNewAndroidKMPPlugin || kmpExtension.targets.any { target -> target is KotlinAndroidTarget }) {
+          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin. You can also try adding then new Android Library KPM plugin: https://developer.android.com/kotlin/multiplatform/plugin. "
```
colinmarsch:
```suggestion
          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin. You can also try adding then new Android Library KMP plugin: https://developer.android.com/kotlin/multiplatform/plugin. "
```

paparazzi-gradle-plugin/src/test/projects/multiplatform-plugin-with-new-android-library-plugin/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    namespace = 'app.cash.paparazzi.plugin.test'
+    compileSdk = libs.versions.compileSdk.get() as int
+    minSdk = libs.versions.minSdk.get() as int
+
+    withJava()
+    withHostTest {
+    }
+  }
+
+//  compileOptions {
```
colinmarsch: Can we remove these commented out lines for now?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -205,8 +234,9 @@ public class PaparazziPlugin @Inject constructor(
 
       val overwriteOnMaxPercentDifferenceProvider = project.overwriteOnMaxPercentDifferenceProvider()
       val failureDir = buildDirectory.dir("paparazzi/failures")
-      val testTaskProvider =
-        project.tasks.withType(Test::class.java).named { it == "test$testVariantSlug" }
+
+      @Suppress("UNCHECKED_CAST")
+      val testTaskProvider = testTasks.matching { it is Test } as TaskCollection<Test>
```
TWiStErRob: What's wrong with withType? as far as I know matching is eager.

gabrielittner:
I've tried this out and it's generally working well, I've found 2 small things:
- I've had to enable `isIncludeAndroidResources` on `withHostTest` or the `KotlinMultiplatformAndroidHostTestCompilation`, without there would be a `ClassNotFoundException` during paparazzi initialization for the `R` class of the module
- The expected directory for the snapshots is still `src/test/snapshots` and not `src/androidHostTest/snapshots`, not a blocker just slightly confusing

geoff-powell:
> I've tried this out and it's generally working well, I've found 2 small things:
> 
> * I've had to enable `isIncludeAndroidResources` on `withHostTest` or the `KotlinMultiplatformAndroidHostTestCompilation`, without there would be a `ClassNotFoundException` during paparazzi initialization for the `R` class of the module

Yes you bring up a good point, [this issue logged with this here that seems related](https://github.com/cashapp/paparazzi/issues/472). With newer versions of AGP it may have manifested itself differently since we don't rely on the `mergeResources` task and build the resource graph ourselves.
- [ ] Investigate

> * The expected directory for the snapshots is still `src/test/snapshots` and not `src/androidHostTest/snapshots`, not a blocker just slightly confusing

Yep, this is a good point to bring up. I tried to pull the host test source directory and that seems to not be working. We were hardcoding `src/test/snapshots` which doesn't work for multiplatform or custom test suite configurations.
- [x] Investigate

Thanks for testing this 🙏🏽

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -138,14 +150,29 @@ public class PaparazziPlugin @Inject constructor(
...
 
-      val testInstrumentation = testVariant.instrumentation
-      testInstrumentation.transformClassesWith(
-        ResourcesCompatVisitorFactory::class.java,
-        InstrumentationScope.ALL
-      ) { }
-      testInstrumentation.setAsmFramesComputationMode(
-        FramesComputationMode.COMPUTE_FRAMES_FOR_INSTRUMENTED_METHODS
-      )
+      // TODO: Fixup with for [Android Multiplatform Library Plugin] doesn't have exception with instrumentation transformations for KMP
```
gabrielittner: Since I couldn't find an issue for thin on the Google tracker I looked into this a bit and opened one https://issuetracker.google.com/issues/462919291. Without deep AGP knowledge it just looks like the plugin is calling 2 methods in the wrong order

geoff-powell: Yep, would assume not a ton of consumers are using KMP and ASM transformations. 
Thanks for reporting the issue. Will add it to the comment
This method will eventually go away and rewrites of ASM will be done at runtime instead of compile time

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: Good point, reverting this back to `.withType()`

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -94,7 +104,9 @@ public class PaparazziPlugin @Inject constructor(
   }
 
   private fun setupPaparazzi(project: Project, extension: AndroidComponentsExtension<*, *, *>) {
-    project.addTestDependency()
+    val newMultiplatformPlugin = extension is KotlinMultiplatformAndroidComponentsExtension
+    val isMultiplatform = newMultiplatformPlugin || project.plugins.hasPlugin("org.jetbrains.kotlin.multiplatform")
+    project.addTestDependency(isMultiplatform)
```
jrodbx: push the "isMultiplatform" logic down into `addTestDependency`?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -376,13 +406,14 @@ public class PaparazziPlugin @Inject constructor(
       .files
   }
 
-  private fun Project.addTestDependency() {
+  private fun Project.addTestDependency(isMultiplatform: Boolean) {
```
jrodbx:
```suggestion
  private fun Project.addTestDependency() {
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -376,13 +406,14 @@ public class PaparazziPlugin @Inject constructor(
...
     val dependency = if (isInternal()) {
       dependencies.project(mapOf("path" to ":paparazzi"))
     } else {
       dependencies.create("app.cash.paparazzi:paparazzi:$VERSION")
     }
-    configurations.getByName("testImplementation").dependencies.add(dependency)
+    val configurationName = if (isMultiplatform) "commonTestImplementation" else "testImplementation"
+    configurations.getByName(configurationName).dependencies.add(dependency)
   }
```
jrodbx:
```suggestion
    val testConfiguration = if (isMultiplatform()) "commonTestImplementation" else "testImplementation"
    configurations.getByName(testConfiguration).dependencies.add(dependency)
  }

  private fun Project.isMultiplatform(): Boolean = plugins.hasPlugin("org.jetbrains.kotlin.multiplatform")
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -65,15 +68,21 @@ public class PaparazziPlugin @Inject constructor(
...
         "One of ${supportedPlugins.joinToString(", ")} must be applied for Paparazzi to work properly."
       }
+      val supportsNewAndroidKMPPlugin = project.plugins.hasPlugin(ANDROID_KOTLIN_MULTIPLATFORM_LIBRARY_PLUGIN)
       project.plugins.withId("org.jetbrains.kotlin.multiplatform") {
         val kmpExtension = project.extensions.getByType(KotlinMultiplatformExtension::class.java)
-        check(kmpExtension.targets.any { target -> target is KotlinAndroidTarget }) {
-          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin"
+        check(supportsNewAndroidKMPPlugin || kmpExtension.targets.any { target -> target is KotlinAndroidTarget }) {
+          "There must be an Android target configured when using Paparazzi with the Kotlin Multiplatform Plugin. You can also try adding then new Android Library KMP plugin: https://developer.android.com/kotlin/multiplatform/plugin. "
         }
```
jrodbx, line 94:
```suggestion
        val usesAndroidKmpLibraryPlugin = project.plugins.hasPlugin("com.android.kotlin.multiplatform.library")
        if (!usesAndroidKmpLibraryPlugin) {
          val kmpExtension = project.extensions.getByType(KotlinMultiplatformExtension::class.java)
          check(kmpExtension.targets.any { target -> target is KotlinAndroidTarget }) {
            "There must be an androidTarget() configured when using Paparazzi with the Kotlin Multiplatform Plugin"
          }
        }
```

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -174,8 +201,10 @@ public class PaparazziPlugin @Inject constructor(
 
       val testVariantSlug = testVariant.name.capitalize()
 
-      project.tasks.named { it == "test$testVariantSlug" }
-        .configureEach { it.dependsOn(writeResourcesTask) }
+      val testTasks = project.tasks.named {
+        it == "test$testVariantSlug" || (isMultiplatform && it == "testAndroidHostTest")
```
jrodbx: - is it possible to have testAndroidHostTest in a non-multiplatform world, i.e., is the `isMultiplatform` necessary?
- is it possible to infer `testAndroidHostTest` via variants?  i'll need to read up on what this actually does, feel free to supply docs/pointers.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -138,14 +150,29 @@ public class PaparazziPlugin @Inject constructor(
...
+       */
+      if (!newMultiplatformPlugin) {
+        val testInstrumentation = testVariant.instrumentation
+        testInstrumentation.transformClassesWith(
+          ResourcesCompatVisitorFactory::class.java,
+          InstrumentationScope.ALL
+        ) { }
+        testInstrumentation.setAsmFramesComputationMode(
+          FramesComputationMode.COMPUTE_FRAMES_FOR_INSTRUMENTED_METHODS
+        )
```
jrodbx, line 185: worth checking/duplicating what AGP actually does here, or are they actively working to fix this?

geoff-powell: Looks like Android multiplatform tests make sure this works as expected.
http://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/integration-test/application/src/test/java/com/android/build/gradle/integration/multiplatform/v2/KotlinMultiplatformAndroidVariantApiTest.kt;l=56?q=COMPUTE_FRAMES_FOR_INSTRUMENTED_METHODS&start=31

geoff-powell: Looks like AGP 9.0.0-beta05 fixes this issue! 🎉

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
geoff-powell: Yes, I just pushed changes that make this simpler and solve for multiplatform android projects and normal Android library/application modules.

eygraber:
Is there any way to support this pre AGP 9?

Kotlin 2.3.0 warns about using `com.android.library`, but Paparazzi doesn't support `com.android.kotlin.multiplatform.library` yet.

gabrielittner:
With the limitations I posted above the PR does work with 8.13 with the main limitation being that custom fonts don't work without the test instrumentation and that requires AGP 9.

geoff-powell:
> Is there any way to support this pre AGP 9?
> 
> Kotlin 2.3.0 warns about using `com.android.library`, but Paparazzi doesn't support `com.android.kotlin.multiplatform.library` yet.

Yep, the plan is to get this out in soon. AGP 9.0 will be a fast follow

alexandrucaraus:
Didn't get into 2.0.0-alpha04 :(, why?

geoff-powell:
> Didn't get into 2.0.0-alpha04 :(, why?

It will be in 2.0.0-alpha04. We just released 2.0.0-alpha03 yesterday and will have alpha04 as a fast follow

eygraber:
I think alpha03 was skipped, and alpha04 was released.

geoff-powell:
Ahh I see, there was a release issue with 03 so we went to 04.

This change will be in alpha05 then, my mistake

IAmShaarawy:
@geoff-powell when it is expected to have alpha05 or at least a snapshot version to use.
we upgraded to AGP 9 and we are blocked.

PR #2149, opened 2025-11-13
Register missing task inputs for Develocity test distribution

### LINKED ISSUE #1790: Missing task input
**Description**
The `paparazzi.test.resources` system property points to the output of `writeResourcesTask` but it's not registered as an input.

**Steps to Reproduce**
When using test distribution from Develocity it will fail to run because that file isn't uploaded.

**Expected behavior**
All inputs should be registered.

**Additional information:**
- Paparazzi Version:  1.3.5
- OS: macOS 15.2
- Compile SDK: 35
- Gradle Version: 8.12
- Android Gradle Plugin Version: 8.7.0

**Screenshots**
n/a

colinmarsch:
## Summary
Fixes #1790

This PR registers two missing task inputs that were causing failures when using Develocity's test distribution feature:

1. **paparazzi.test.resources** - The output file from `writeResourcesTask` that's used as a system property
2. **paparazzi.resources** - The `layoutlibResourcesFileCollection` used in system properties

Both are now properly registered as task inputs with `PathSensitivity.NONE`, ensuring these files are uploaded when using Develocity test distribution.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -243,6 +243,12 @@ public class PaparazziPlugin @Inject constructor(
         test.inputs.files(layoutlibNativeRuntimeFileCollection)
           .withPropertyName("paparazzi.nativeRuntime")
           .withPathSensitivity(PathSensitivity.NONE)
+        test.inputs.files(layoutlibResourcesFileCollection)
+          .withPropertyName("paparazzi.resources")
+          .withPathSensitivity(PathSensitivity.NONE)
+        test.inputs.files(writeResourcesTask.flatMap { it.paparazziResources.asFile })
+          .withPropertyName("paparazzi.test.resources")
+          .withPathSensitivity(PathSensitivity.NONE)
 
         test.inputs.dir(
           isVerifyRun.flatMap {


colinmarsch:
> Nice, do you have an example of before and after where this is working? Maybe via a build scan

I'm not exactly sure how to manually test this out. Do you have any thoughts here how to trigger a build scan or another way to verify this before/after?

TWiStErRob:
Are we sure NONE is the right sensitivity? If the files are relocated, wouldn't we need to invalidate the inputs and re-run the task? Or is that not what it means?

geoff-powell:
> I'm not exactly sure how to manually test this out. Do you have any thoughts here how to trigger a build scan or another way to verify this before/after?

You can run whatever task like `export CI="true" && ./gradlew :sample:verifyPaparazziDebug` to publish build scan

colinmarsch:
> > I'm not exactly sure how to manually test this out. Do you have any thoughts here how to trigger a build scan or another way to verify this before/after?
> 
> You can run whatever task like `export CI="true" && ./gradlew :sample:verifyPaparazziDebug` to publish build scan

I got a build scan here: https://scans.gradle.com/s/u37l6sna4uhao, tried looking around for the inputs but I can't seem to find them 🤔

colinmarsch:
> Are we sure NONE is the right sensitivity? If the files are relocated, wouldn't we need to invalidate the inputs and re-run the task? Or is that not what it means?

I don't think the paths are relevant here and only the content of these matter for the invalidation. Happy to change this if there is something I'm missing here though!

colinmarsch:
Closing in favour of https://github.com/cashapp/paparazzi/pull/2205

PR #2150, opened 2025-11-13
Support custom view accessibility hierarchy ordering in accessibility snapshots

colinmarsch:
Adds support for views similar to the custom accessibility hierarchy ordering support already added for compose ui back here: https://github.com/cashapp/paparazzi/pull/2019


colinmarsch:
> This looks good to me 👍🏽 . At some point, I want to look into pulling the accessibility info from layout lib as they might take this into account so we don't have to re-invent the wheel.

Yeah, happy to look into this in the near future too!

PR #2162, opened 2025-11-20
Configure default differ when none is specified

joshskeen:
Differs are internal classes, but copying them to provide one for the expected argument of public SnapshotVerifier isn't ideal. 
Instead, read a system property and select from default if no argument is present for differ.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
@@ -19,8 +19,12 @@ import app.cash.paparazzi.Differ
 import app.cash.paparazzi.SnapshotHandler.FrameHandler
 import app.cash.paparazzi.internal.ImageUtils
 import app.cash.paparazzi.internal.apng.ApngVerifier
+import app.cash.paparazzi.internal.differs.DeltaE2000
+import app.cash.paparazzi.internal.differs.Flip
+import app.cash.paparazzi.internal.differs.Mssim
 import app.cash.paparazzi.internal.differs.OffByTwo
 import app.cash.paparazzi.internal.differs.PixelPerfect
+import app.cash.paparazzi.internal.differs.Sift
 import okio.Path.Companion.toOkioPath
 import java.awt.image.BufferedImage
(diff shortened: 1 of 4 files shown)


paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
```diff
@@ -103,3 +105,17 @@ public class SnapshotVerifier @JvmOverloads constructor(
       }
   }
 }
+
+/**
+ * Convenience method to select a built-in differ via system property.
+ * This allows selection of built-in differs without requiring direct access to internal implementations.
+ */
```
jrodbx:
```suggestion
```

paparazzi/src/main/java/app/cash/paparazzi/SnapshotVerifier.kt
```diff
@@ -103,3 +105,17 @@ public class SnapshotVerifier @JvmOverloads constructor(
...
+private fun determineDiffer() =
+  System.getProperty("app.cash.paparazzi.differ")?.lowercase().let { differ ->
```
jrodbx, line 114: can we remove the equivalent elsewhere and possibly update determineHandler accordingly?

geoff-powell: +1

geoff-powell:
@joshskeen can we update https://github.com/cashapp/paparazzi/blob/72c3463c8bb398e7afcd8d9ea7e8686402bf709b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt#L191-L205C17 to use `determineDiffer` logic 🙏🏽

jrodbx:
> @joshskeen can we update https://github.com/cashapp/paparazzi/blob/72c3463c8bb398e7afcd8d9ea7e8686402bf709b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt#L191-L205C17 to use `determineDiffer` logic 🙏🏽

since `determineDiffer` moved to `SnapshotVerifier`, we can remove the `val differ` and the associated parameter passing since that arg now has a default.

PR #2173, opened 2025-11-29
Introduce MainDispatcherValidator to Paparazzi tests

nishatoma:
Hello! 

This is a concept PR that introduces `Strict Dispatcher` mode, which basically prevents developers from using `Dispatchers.setMain(...)` in Paparazzi tests. Typically, the `Dispatchers.set/reset` methods are used to help testing coroutines on JVM tests in which you do not have access to the main thread. It's beneficial as it allows you to control coroutines executions.

However, this control means flaky tests when you combine this functionality with Snapshot testing. In particular, I've observed the following 2 main issues in our large, multi-module code base:

- Often times, some developers forget to call `advanceUntilIdle()` or even forget to use `UnconfinedTestDispatcher()` which means certain coroutines remain scheduled, but never executed, even after your paparazzi tests finish. This can lead to `OutOfMemory` errors in modules with large amounts of snapshot tests.

- Flaky tests that sometimes just fail because the coroutine that naturally was supposed to behave as expected did not get to finish because the main dispatcher was modified.

Since `Paparazzi` uses its own `AndroidUi` dispatcher anyway, I never saw any benefit to using the `Dispatchers.setMain` here, so, I've introduced an optional system property that when turned on, will catch if the `Main` dispatcher was tampered with, and throw exceptions early on let developers refactor their tests.

I've put multiple test cases that demonstrate the behaviour, let me know what are your thoughts!


geoff-powell:
Thanks for the contribution this makes a lot of sense to me!

- [ ] My only feedback would be instead of a `:sample` test, could you create a test project in `:paparazzi-gradle-plugin/src/test/test-projects/*` and add the test task to `PaparazziGradlePluginTest.kt`?

paparazzi/src/main/java/app/cash/paparazzi/internal/validation/MainDispatcherValidator.kt
```diff
@@ -0,0 +1,50 @@
...
+ */
+internal class MainDispatcherValidator(
+  private val defaultHandlerClassName: String = "HandlerContext",
+  private val strictMode: Boolean
+) {
+
+  internal fun checkMainCoroutineDispatcher() {
+    if (isMainDispatcherModified()) {
+      throw MainDispatcherAlteredException(
+        message = "Main dispatcher was modified. Please do not change Main dispatcher in paparazzi tests."
```
geoff-powell: We can add to this exception message something around the calling `Dispatchers.resetMain()` before `Paparazzi.snapshot()` is called if it is needed for test setup reasons like async operations (networking, persistence queries, etc) in UI layer.

`If overriding Main dispatched is needed, please call "Dispatchers.resetMain()" before Paparazzi.snapshot is called`

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -166,6 +170,7 @@ public class Paparazzi @JvmOverloads constructor(
     frameCount: Int = 1,
     fps: Int = -1
   ): SnapshotHandler.FrameHandler {
+    mainDispatcherValidator.checkMainCoroutineDispatcher()
```
geoff-powell: nit: Can we move this into [PaparazziSdk.takeSnapshots()](https://github.com/cashapp/paparazzi/blob/bdd87b571e0b6a311a03eeb19838875f6ced151b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt#L273)

nishatoma:
> Thanks for the contribution this makes a lot of sense to me!
> 
> 
> 
> - [ ] My only feedback would be instead of a `:sample` test, could you create a test project in `:paparazzi-gradle-plugin/src/test/test-projects/*` and add the test task to `PaparazziGradlePluginTest.kt`?

Thank you for the feedback! I will look into this once back home!

paparazzi/src/main/java/app/cash/paparazzi/internal/validation/MainDispatcherValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+      )
+    }
+  }
+
+  private fun isMainDispatcherModified(): Boolean {
+    return try {
+      strictMode && !Dispatchers.Main.immediate.javaClass.name.contains(defaultHandlerClassName, ignoreCase = true)
+    } catch (_: Exception) {
+      // If an exception happens when trying to access Dispatchers.Main then that means it was
+      // never altered to begin with.
```
jrodbx, line 49:
```suggestion
      // If an exception happens while accessing Dispatchers.Main, then it was never altered to begin with.
```

paparazzi/src/main/java/app/cash/paparazzi/internal/validation/MainDispatcherValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+ * If someone uses `Dispatchers.setMain(...)` within a Paparazzi test, there is a chance
+ * that they make the test pass, but not really execute some of the scheduled coroutines within
+ * the test dispatcher. This in turn can pile up in large modules with lots of snapshot tests, and
+ * eventually can cause `java.lang.OutOfMemoryError.`
+ *
+ * Setting the `Dispatchers.Main` to a `TestDispatcher` can also lead to flaky tests if the developer
+ * doesn't properly call `advanceUntilIdle()`.
+ *
+ * Finally, Paparazzi already has access to its own `AndroidUiDispatcher,` so there
+ * aren't many benefits to use `Dispatchers.setMain()` within Paparazzi tests.
```
jrodbx, line 21: Thank you for the contribution ❤️ 

I'm still trying to understand the use cases here though, namely:
* In what case would someone deliberately modify Dispatchers.Main?
* Even so, is it appropriate for this library to be the watchdog for this potential scenario versus a custom test rule in your codebase that would check for this?

Additionally, I'm not understanding where preventing Dispatchers.Main from being set prevents OOMEs from occurring.

nishatoma: Thank you @jrodbx for the feedback!

Re: is it needed on Paparazzi code

This is a really good call out! Im not aware of other companies or teams but we use custom rules in our codebase, so I thought it would be useful to have a centralized place! Although I am okay as well with other options, hence this being a concept PR!

Re: dispatcher rules set deliberately

I believe some developers did this by accident, not deliberate. For example maybe a new person joins and copies some code from other tests.

For context, we have Paparazzi, Robolectric and other JVM tests theoughout 50+ large modules, and I was able to fix 'OOME's by doing this. Sadly Im not able to reproduce on fresh projects, so I can't prove it is 100% being caused by the dispatcher.

geoff-powell: hmm, in that case I feel like the responsibility of managing main dispatcher is less something the `Paparazzi` should manage and more on the plugin consumers. Curious on your thoughts here @jrodbx

geoff-powell:
Closing this PR as we are not sure this is Paparazzi's responsibility to manage main dispatchers

PR #2176, opened 2025-12-03
Update DeviceConfigs

jrodbx:
Claude-assisted fetch, transform, and diff of: https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-main/sdklib/src/main/resources/com/android/sdklib/devices/nexus.xml

Release dates and other bits pulled from:
  Sources:
  - https://en.wikipedia.org/wiki/Google_Pixel
  - https://endoflife.date/pixel
  - https://en.wikipedia.org/wiki/Pixel_9_Pro_Fold


geoff-powell (approved):
nice 💪🏽 🎉

Issue #2136, opened 2025-10-24
Add support for com.android.kotlin.multiplatform.library plugin


eboudrant:
Paparazzi is working fine in a KMP project if you apply `com.android.library` plugin but `com.android.kotlin.multiplatform.library` will be the new recommended way by Google to build a KMP Android library: 
- https://developer.android.com/kotlin/multiplatform/plugin

With `com.android.kotlin.multiplatform.library` paparazzi fails with:
```
One of com.android.application, com.android.library, com.android.dynamic-feature must be applied for Paparazzi to work properly.
```

nishatoma:
@geoff-powell is it okay if I make an attempt at this?

jrodbx:
@nishatoma #2115 once landed will support this

nishatoma:
That's awesome, thank you!

geoff-powell:
Unfortunately there is a bug in AGP 8.x.x for supporting this with our ASM rewrite for compat font loading. 
I plan on investigating moving away from ASM rewrites in the gradle infra and doing it at runtime with ByteBuddy.
This would make us less reliant on AGP and helps in moving toward other build systems or using Paparazzi as a standalone application

geoff-powell:
I have this #2211 to avoid ASM requirements and unblocks the dependency for this to land after AGP 9.0 🎉

oldergod:
Do we know what's left for this ticket?

#2115 added support for `com.android.kotlin.multiplatform.library`, yes? We also list it: https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L80

frett:
I just configured Paparazzi in an Android & iOS KMP project that uses the multiplatform plugin yesterday using the SNAPSHOT build.

I did have to manually disable the paparazzi dependency for the iOS source set because paparazzi is being added to the common source set and not the android specific one [here](https://github.com/cashapp/paparazzi/blob/master/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt#L417). This was causing gradle to fail because it couldn't find the library for the ios targets.

The source set the plugin gets added to probably needs to be updated to target the android test source set only and not just the commonTest one.

IAmShaarawy:
@frett i encountered the same as well

PR #2198, opened 2025-12-18
Bump AGP to 9.0.0

geoff-powell:


settings.gradle
```diff
@@ -50,7 +51,8 @@ new File(rootDir, "paparazzi-gradle-plugin/src/test/projects").eachDir {
 }
 
 def isIdea = !(System.getProperty("idea.version") ?: "").isEmpty()
-if (isIdea) {
+if (isIdea || providers.gradleProperty("app.cash.paparazzi.includeTestProjectModules").orNull == "true") {
```
geoff-powell: Added a prop to test projects from terminal

jrodbx: what's the context here?  also, can this be bumped independently vs rebased on the current target?

TWiStErRob: What's the benefit of:
```
gradlew -Papp.cash.paparazzi.includeTestProjectModules=true
```
Vs
```
gradlew -Didea.version=anything
```

paparazzi-gradle-plugin/src/test/projects/verify-resources-kotlin/consumer/build.gradle
```diff
@@ -14,8 +13,11 @@ android {
     sourceCompatibility = libs.versions.javaTarget.get()
     targetCompatibility = libs.versions.javaTarget.get()
   }
-  libraryVariants.configureEach {
-    it.registerGeneratedResFolders(project.layout.buildDirectory.files("generated/res/extra"))
+}
+
+androidComponents {
+  onVariants(selector().all()) { variant ->
+    variant.sources.res.addStaticSourceDirectory(project.layout.buildDirectory.dir("generated/res/extra").get().asFile.toString())
```
TWiStErRob, line 20: This migration is possible in 8.x too as far as I know, this could go directly to main.

settings.gradle
geoff-powell: reverting this

paparazzi-gradle-plugin/src/test/projects/verify-resources-kotlin/consumer/build.gradle
geoff-powell: Pulled that out here: https://github.com/cashapp/paparazzi/pull/2199

paparazzi-gradle-plugin/src/test/projects/verify-resources-java/consumer/build.gradle
```diff
@@ -26,3 +28,8 @@ dependencies {
   implementation projects.module1
   implementation projects.module2
 }
+
+abstract class SourceGenTask extends DefaultTask {
```
TWiStErRob: Is this unused?

gradle/libs.versions.toml
```diff
@@ -59,7 +59,7 @@ okio = { module = "com.squareup.okio:okio", version = "3.16.4" }
 
 tools-common = { module = "com.android.tools:common", version.ref = "androidTools" }
 tools-layoutlib = { module = "com.android.tools.layoutlib:layoutlib", version.ref = "layoutlib" }
-tools-layoutlib-api = { module = "com.android.tools.layoutlib:layoutlib-api", version = "31.13.2" }
+tools-layoutlib-api = { module = "com.android.tools.layoutlib:layoutlib-api", version = "32.0.0" }
```
SimonMarquis: @geoff-powell shouldn't we enforce this dependency to use the `androidTools` version ref?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -139,29 +139,19 @@ class PaparazziPluginTest {
     fixtureRoot.resolve("build-cache").registerForDeletionOnExit()
 
     val firstRun = gradleRunner
-      .withArguments("testRelease", "testDebug", "--build-cache", "--stacktrace")
```
SimonMarquis, line 163: Can we keep all release-related stuff? AGP 9 decided to keep only a single default variant for unit tests, which is set to debug by default. This is very unfortunate since testing "debug" code is usually not what you would expect. (I know Android build variants are usually misused).

Keeping these would only require to restore the previous value default value of `android.onlyEnableUnitTestForTheTestedBuildType` ([docs](https://developer.android.com/build/releases/agp-9-0-0-release-notes))

geoff-powell: Ya I went back and forth between using AGP defaults and removing. The only test that really needs this property would be ensuring the `verifyPaparazzi` test and `resourceCache` test runs release and and debug.

gradle/libs.versions.toml
geoff-powell: We don't enforce this due to api dependency tends to be updated more frequent and we don't necessarily need these to be tied together.

colinmarsch (approved):

MontyKirner:
Hey, is there any chance we could please get this merged now that it's been approved, and then have an `alpha5` version pushed out?

We use many 3rd party libraries on our project, but currently Paparazzi is the only one which is blocking us from updating to AGP 9.0.0 😅

eboudrant:
Hello, I'm curious, is this change going to add support to the new KMP android plugin `com.android.kotlin.multiplatform.library`? We're blocked by that and the [PR](https://github.com/cashapp/paparazzi/pull/2211) linked to #2136 issue was closed. Thanks!

difiTran:
Hello!
Is there any estimation when we can expect this? :) 
@geoff-powell

montykapadmi:
Pinging this PR again, it's been another month since a user asked for an update on this, are there other related PRs which prevent this version bump to 9.0.0?

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -108,6 +108,62 @@ public class PaparazziPlugin @Inject constructor(
     }
   }
 
+  private class AndroidProjectScope(
```
colinmarsch: It doesn't look like this is used anywhere, do we need this in this PR?

oldergod:
Superseded by https://github.com/cashapp/paparazzi/pull/2318 ?

colinmarsch:
Just rebased this branch on the latest master to resolve conflicts and removed the unused `AndroidProjectScope` class.

colinmarsch:
Just rebased to keep this conflict free

jrodbx:
updated the multiplatform tests because the AGP bump resulted in failing IDE syncs, saying:

```                                                                                                                                                  
  An exception occurred applying plugin request [id: 'com.android.library']                                                                                                                                
  > Failed to apply plugin 'com.android.internal.library'.                                                                                                                                                 
     > java.lang.IllegalStateException: The 'com.android.library' (or 'com.android.application') plugin is not compatible with the 'org.jetbrains.kotlin.multiplatform' plugin since AGP 9.0.              
       Solution:                                                                                                                                                                                           
         - [Recommended] Replace the 'com.android.library' plugin with the 'com.android.kotlin.multiplatform.library' plugin (see https://developer.android.com/r/tools/built-in-kotlin).                  
         - Or set the Gradle property 'android.builtInKotlin=false' and 'android.newDsl=false' to temporarily bypass this issue.     
```

however, 'android.builtInKotlin=false' and 'android.newDsl=false' didn't resolve the issue.

see commit https://github.com/cashapp/paparazzi/pull/2198/changes/433ad5ca5297f98b06e000a29f647be39d265363

thoughts?

eygraber:
> thoughts

Is there an issue with using `com.android.kotlin.multiplatform.library`?

jrodbx:
> Is there an issue with using `com.android.kotlin.multiplatform.library`?

I don't think so; in fact, the recent push updates things to do exactly that.  I guess I didn't realize that AGP 9.0 now forces this, despite adding the flags to maintain legacy behavior.

jrodbx (approved):

colinmarsch (approved):
The latest changes you made here look good to me @jrodbx!

PR #2199, opened 2025-12-19
Migrate legacy variant api to androidComponents

geoff-powell:


jrodbx (commented):
Note that the previous legacy approach uses a (lazy) FileCollection which isn't resolved until later in the AGP lifecycle.

Does this modern approach of extracting the provider's value in `onVariants` result in any suboptimal Gradle perf issues, like eager task resolution?  

A quick before/after build scan comparison should show that.

Not a blocker or cause for revert, but would be curious for a quick follow-up comment here.

In either case ✅

geoff-powell:
Ya good point. Instead of `addStaticSourceDirectory` I could use [addGeneratedSourceDirectory](https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/variant/SourceDirectories#addGeneratedSourceDirectory(org.gradle.api.tasks.TaskProvider,kotlin.Function1)) with the `generateSource` to allow for future resolution

PR #2202, opened 2025-12-25
Update gradle-wrapper to RC as soon as available

TWiStErRob:
This will open RC and stable PRs.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -47,6 +47,13 @@
       ],
     },
     {
+      description: "Disable Gradle wrapper milestone releases, only interested stables and RCs for now.",
+      matchPackageNames: [
+        "gradle",
+      ],
+      allowedVersions: "!/-milestone-\\d+$/",
+    },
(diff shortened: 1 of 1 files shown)


geoff-powell:
Looks good, thanks!

jrodbx:
do we want to start updating to Gradle release candidates vs only stable versions?  also, will this only update the wrapper vs the Gradle binary/scripts?  if so, any concerns there?

TWiStErRob:
@jrodbx exactly the same as other Renovate PRs, but also includes RCs:
https://github.com/cashapp/paparazzi/pull/2204/files

The point of this is that:
1. You get a notification when RC is released
2. CI is immediately executed, so any breakages you have a chance to address **before** stable

Note: the PR doesn't have to be merged. We can call it poor man's shadow build. 😅

Note 2: if there are no internal API usages, the relevance of this goes waaaay down.

PR #2203, opened 2025-12-27
Introduce defaultLocale system property

### LINKED ISSUE #1982: Add a way to specific the default locale used in Paparazzi
If no locale is specified in the the `DeviceConfig` it seems like it picks up `Locale.getDefault()` ? 

It would be nice if Paparazzi had a way to force a default locale for all paparazzi tests (overridable on a per test basis). 

```
app.cash.paparazzi.localeDefault=en-rUS
```

This follows the pattern of `app.cash.paparazzi.maxPercentDifferenceDefault=0.001`

nishatoma:
This PR fixes #1982 by adding a `app.cash.paparazzi.defaultLocale` system property.

If this is a concept the team wants to go ahead with, I'm also open to suggestions on whether this default value should be left empty (and up to the consumer) or whether to actually set it to something like `en-rUS` as per @yogurtearl's suggestion!

Thank you in advance for any feedback.

Closes #1982

### DIFF
--- a/paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleTest.kt
+++ b/paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleTest.kt
@@ -23,7 +23,9 @@ class DefaultLocaleTest(@TestParameter val locale: Locale) {
       object : Statement() {
         override fun evaluate() {
           try {
-            System.setProperty("app.cash.paparazzi.defaultLocale", locale.tag.orEmpty())
+            locale.tag?.let { localeTag ->
+              System.setProperty("app.cash.paparazzi.defaultLocale", localeTag)
+            }
             base.evaluate()
           } catch (_: Exception) {
             System.clearProperty("app.cash.paparazzi.defaultLocale")


paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
```diff
@@ -0,0 +1,40 @@
...
+  @get:Rule
+  val paparazzi = Paparazzi()
+
+  @After
+  fun tearDown() {
+    System.clearProperty("app.cash.paparazzi.localeDefault")
+  }
+
+  @Test
+  fun `GIVEN system prop WHEN snapshot THEN use provided locale`() {
```
geoff-powell: nit: We don't use this structure in other tests, so can this be something like `Verify system prop default locale`

nishatoma: Sorry about that, fixed!

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
```diff
@@ -0,0 +1,40 @@
...
+) {
+
+  internal enum class SystemPropLocale(val tag: String?) {
+    Default(null),
+    EN_GB("en-rGB"),
+    FR_FR("fr-rFR")
+  }
+
+  init {
+    System.setProperty("app.cash.paparazzi.localeDefault", defaultLocale.tag.orEmpty())
```
jrodbx: nit: move to a `@Before fun setup` pattern for symmetry and technically more accurate lifecycle

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+internal class DefaultLocaleSnapshotTest(
+  @TestParameter private val defaultLocale: SystemPropLocale
+) {
+
+  internal enum class SystemPropLocale(val tag: String?) {
```
jrodbx:
```suggestion
  internal enum class Locale(val tag: String?) {
```

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
```diff
@@ -0,0 +1,40 @@
...
+  @get:Rule
+  val paparazzi = Paparazzi()
+
+  @After
+  fun tearDown() {
+    System.clearProperty("app.cash.paparazzi.localeDefault")
+  }
+
+  @Test
+  fun `Verify system prop default locale`() {
```
jrodbx:
```suggestion
  fun `verify system property sets default locale`() {
```

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/snapshots/images/app.cash.paparazzi.plugin.test_LocaleQualifierTest_locale[GB].png
Comment (jrodbx) [inline paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/snapshots/images/app.cash.paparazzi.plugin.test_LocaleQualifierTest_locale[GB].png:1]: irrelevant updates, remove?

paparazzi/src/main/java/app/cash/paparazzi/SnapshotHandler.kt
```diff
@@ -28,3 +30,23 @@ public interface SnapshotHandler : Closeable {
...
+ *
+ * Uses the provided `locale` if passed in, otherwise falls back to the
+ * `app.cash.paparazzi.localeDefault` system property, and finally the
+ * default `LocaleQualifier()` if neither are available.
+ *
+ * @param locale Optional Locale to use for snapshot tests.
+ * @return LocaleQualifier for snapshot tests.
+ */
+internal fun detectLocaleQualifierDefault(locale: String?): LocaleQualifier {
+  val propLocale = System.getProperty("app.cash.paparazzi.localeDefault")
```
jrodbx:
```suggestion
  val propLocale = System.getProperty("app.cash.paparazzi.defaultLocale")
```

paparazzi/src/main/java/app/cash/paparazzi/SnapshotHandler.kt
```diff
@@ -28,3 +30,23 @@ public interface SnapshotHandler : Closeable {
...
+ * Returns a LocaleQualifier for snapshots.
+ *
+ * Uses the provided `locale` if passed in, otherwise falls back to the
+ * `app.cash.paparazzi.localeDefault` system property, and finally the
+ * default `LocaleQualifier()` if neither are available.
+ *
+ * @param locale Optional Locale to use for snapshot tests.
+ * @return LocaleQualifier for snapshot tests.
+ */
+internal fun detectLocaleQualifierDefault(locale: String?): LocaleQualifier {
```
jrodbx: why is this method in SnapshotHandler?  seems like a companion object method in DeviceConfig would be a better home?

jrodbx: arguably, `detectMaxPercentDifferenceDefault` shouldn't be there either 😅

paparazzi/src/main/java/app/cash/paparazzi/DeviceConfig.kt
jrodbx:
```suggestion
  public val locale: String? = detectLocaleProperty(),
```

paparazzi/src/main/java/app/cash/paparazzi/SnapshotHandler.kt
```diff
@@ -28,3 +30,23 @@ public interface SnapshotHandler : Closeable {
...
+  val propLocale = System.getProperty("app.cash.paparazzi.localeDefault")
+    ?.takeIf { it.isNotEmpty() }
+  return when {
+    locale != null -> LocaleQualifier.getQualifier(locale)
+    propLocale != null -> LocaleQualifier.getQualifier(propLocale)
+    else -> LocaleQualifier()
```
jrodbx: let's minimize this method, reverting the LocaleQualifier.getQualifier logic to where it was, and making this a "system property value [string] or null" operation.

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import app.cash.paparazzi.Paparazzi
+import com.google.testing.junit.testparameterinjector.TestParameter
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.After
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+internal class DefaultLocaleSnapshotTest(
```
jrodbx:
```suggestion
class DefaultLocaleTest(
```

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
nishatoma: Had to move to `@BeforeClass` since now we load the locale property earlier, I also had to reduce to 1 system property since the parametrized test wouldn't work with the `@BeforeClass.` Is this okay?

paparazzi/src/main/java/app/cash/paparazzi/SnapshotHandler.kt
nishatoma: Moved to `DeviceConfig`, is it okay to handle the `detectMaxPercentDifferenceDefault` in a separate PR?

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
jrodbx: ah, I forget that `@Before`/`@After` has a separate lifecycle from `@Rule`s.  That being said, you can replace them with a RuleChain to get around this:

```
  @get:Rule 
  val chain: RuleChain = RuleChain
    .outerRule { base, _ ->
      object : Statement() {
        override fun evaluate() {
          try {
            System.setProperty("app.cash.paparazzi.defaultLocale", "fr-rFR")                    
            base.evaluate()
          } catch (e: Exception) {
            System.clearProperty("app.cash.paparazzi.defaultLocale")
          }
        }
      }
    }
    .around(paparazzi)
```

This should also allow you to re-introduce the parametrized test approach.

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/gradle.properties
```diff
@@ -0,0 +1,2 @@
+android.useAndroidX=true
+app.cash.paparazzi.defaultLocale=""
```
jrodbx:
```suggestion
```

shouldn't be necessary given the default mode?

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
nishatoma: @jrodbx Thank you! I've tried this as well + ordering of rules, and the `DeviceConfig` being a companion object makes this more difficult than it seems.. we could however use a fresh copy of the config each time.. something like this works for example:

```
  @Before
  fun setup() {
    System.setProperty("app.cash.paparazzi.defaultLocale", locale.tag.orEmpty())
    paparazzi.unsafeUpdateConfig(
        locale = System.getProperty("app.cash.paparazzi.defaultLocale")?.takeIf { it.isNotEmpty() }
      )
    )
  }

  @After
  fun teardown() {
    System.clearProperty("app.cash.paparazzi.defaultLocale")
  }
```

Would this approach be acceptable?

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/gradle.properties
nishatoma: Removed, thank you!

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleSnapshotTest.kt
jrodbx: can you elaborate on what you mean by "the DeviceConfig being a companion object makes this more difficult than it seems"?   if you inline the Paparazzi rule into the chain, then the locale property setting in the 1st custom rule should take effect before the Paparazzi rule is instantiated.

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/gradle.properties
jrodbx, line 1: this file can be deleted

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.google.testing.junit.testparameterinjector.TestParameterInjector
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(TestParameterInjector::class)
+class DefaultLocaleTest(@TestParameter val locale: Locale) {
+
```
jrodbx:
```suggestion
```

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    )
+  }
+
+  @After
+  fun teardown() {
+    System.clearProperty("app.cash.paparazzi.defaultLocale")
+  }
+
+  @get:Rule
+  val paparazzi = Paparazzi()
```
jrodbx:
```suggestion
 @get:Rule 
  val chain: RuleChain = RuleChain
    .outerRule { base, _ ->
      object : Statement() {
        override fun evaluate() {
          try {
            System.setProperty("app.cash.paparazzi.defaultLocale", "fr-rFR")                    
            base.evaluate()
          } catch (e: Exception) {
            System.clearProperty("app.cash.paparazzi.defaultLocale")
          }
        }
      }
    }
    .around(Paparazzi())
```

nishatoma: Ahh I misunderstood, it wasn't a config issue, it was because I kept the global paparazzi variable! After applying your changes it works now:) Thank you so much ❤️

paparazzi-gradle-plugin/src/test/projects/locale-qualifier/src/test/java/app/cash/paparazzi/plugin/test/DefaultLocaleTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    GB("en-rGB")
+  }
+
+  @get:Rule
+  val chain: RuleChain = RuleChain
+    .outerRule { base, _ ->
+      object : Statement() {
+        override fun evaluate() {
+          try {
+            System.setProperty("app.cash.paparazzi.defaultLocale", locale.tag.orEmpty())
```
jrodbx: This sets the property to blank vs not setting the property.  Is that what you're intending to test with the null value?

nishatoma: Yah the logic should work for either empty or null, but I can do like a ?.let if you want to for the first scenario so we could test exactly null!

Issue #2206, opened 2026-01-01
Progress Bar Accessibility Information Incorrectly Ordered


Jonas-I:
**Description**
When applying the `progressBarRangeInfo` to the semantics of a composable, the screenshot generates the accessibility information out of order. The screenshot shows the text first, then the semantic property information. However, TalkBack states the `progressBarRangeInfo` first, then the text value.

**Steps to Reproduce**
We can implement the following sample test and apply the `progressBarRangeInfo` modifier:
```kotlin
class ProgressBarRangeInfoTest {
    
    @get:Rule
    val paparazzi = Paparazzi(renderExtensions = setOf(AccessibilityRenderExtension()))

    @Test
    fun progressBarRangeInfoSample() {
        paparazzi.snapshot {
            Text(
                text = "Progress Bar",
                modifier = Modifier.semantics {
                    progressBarRangeInfo = ProgressBarRangeInfo(0.5f, 0f..1f)
                }
            )
        }
    }
}
```

**Expected behavior**
We would expect the accessibility information in the screenshot to reflect what is mentioned in TalkBack:

Current:
```
Progress Bar, <progress>: 50%
```

Expected:
```
<progress>: 50%, Progress Bar
```

**Additional information:**
- Paparazzi Version: 2.0.0-alpha02
- OS: MacOS
- Compile SDK: 36.1.0
- Gradle Version: 9.0.0
- Android Gradle Plugin Version: 8.13.12

**Screenshots**
| `ProgressBarRangeInfo` Accessibility Screenshot  |
| --- |
| <img width="1000" height="888" alt="Image" src="https://github.com/user-attachments/assets/5c2dddc2-1949-4c56-93e8-32c93b99abc1" /> |

| TalkBack w/ Caption (Device: Pixel 8 Pro) |
| --- |
| <img width="439" height="795" alt="Image" src="https://github.com/user-attachments/assets/56eada80-04db-49fd-b431-6370b5143bfc" /> |

PR #2208, opened 2026-01-05
Fix progress bar accessibility ordering

colinmarsch:
Fixes the ordering of the progress info not aligning with Talkback. Verified locally that the reported Talkback ordering is accurate and the `AccessibilityRenderExtension` was in error here.

Resolves https://github.com/cashapp/paparazzi/issues/2206


geoff-powell:
Curious if we have usage of progress bar accessibility text before this change?

Like the `verify progress semantic information represented` test?

geoff-powell:
- [x] ~wait for 2.0.0-alpha03 release before merging~

colinmarsch:
> Curious if we have usage of progress bar accessibility text before this change?
> 
> Like the `verify progress semantic information represented` test?

I just moved the change from this new test case into that one so that we don't need the new snapshot here. Thanks for bringing this up!

colinmarsch:
> * [ ]  wait for 2.0.0-alpha03 release before merging

When is this planned to be released?

geoff-powell:
> > * [ ]   wait for 2.0.0-alpha03 release before merging
> 
> When is this planned to be released?

Should be soon. This seems safe to include in the release so will merge this now. Thanks! 🙏🏽

PR #2209, opened 2026-01-07
Allow Paparazzi usage for multiple frameworks (Jupiter, TestNg, Kotest)

### LINKED ISSUE #282: Decouple Paparazzi from JUnit 4 (TestRule)

nishatoma:
## Credits & Overview

Credits to: @TWiStErRob and community members for inspiration based off of #282 and #293. 

Although, this approach is a bit different.. basically we extract the logic for setting up`PaparazziSdk` and let consumers decide which test framework they want to use, while still keeping the `TestRule`.

This has few added benefits:

1) Existing consumers can still use JUnit 4 as is, since `Paparazzi` still implements `TestRule.`
2) If a new library `JUnit X` comes out tomorrow, then `Paparazzi` doesn't have to handle it. Consumers can use the public APIs to configure their test lifecycle.

I understand that the long term goal is to decouple from JUnit completely, however, this is an interim step that fixes #282 and allows consumers to achieve that end goal without breaking the library.

## Tests

- Added the following for demonstration:
&nbsp;&nbsp; 1) Jupiter example tests (extensions, parametrized tests)
&nbsp;&nbsp; 2) Kotest example test
&nbsp;&nbsp; 3) TestNg example test

## README overview:

https://github.com/nishatoma/paparazzi/tree/expose-sdk-config


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -119,6 +105,32 @@ public class Paparazzi @JvmOverloads constructor(
     sdk.prepare()
   }
 
+  public fun beforeTest(testName: TestName) {
```
geoff-powell: nit: can we name this `setup`?

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -119,6 +105,32 @@ public class Paparazzi @JvmOverloads constructor(
...
+      validateAccessibility = validateAccessibility,
+      onNewFrame = { frameHandler.handle(it) },
+      useDeviceResolution = useDeviceResolution
+    )
+    sdk.setup()
+    this.testName = testName
+    sdk.prepare()
+  }
+
+  public fun afterTest() {
```
geoff-powell: nit : can we name this `teardown` instead?

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
geoff-powell, line 1: This class will be renamed eventually (maybe in final 2.0.0) to better reflect its junit test rule nature and the test name tracking will be generalized. This will make things easier to manage with multiple test suites but thanks for doing the initial work here 🎉 🙏🏽

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -60,6 +60,30 @@ class PaparazziPluginTest {
...
+  }
+
+  @Test
+  fun supportsTestNgLibrary() {
+    val fixtureRoot = File("src/test/projects/supports-testng")
+
+    gradleRunner.withArguments("verifyPaparazziDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+  }
+
```
geoff-powell, line 86: We might want to point out these sample projects in the README to point folks in the right direction on how you can set up test suites outside of JUnit. Can be a fast follow

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
nishatoma: Appreciate the comment, hope people find this useful, I know Jupiter support will definitely help bring down test runtimes :)

geoff-powell:
- [x] Wait for 2.0.0-alpha04 release before approving / merging

nishatoma:
@geoff-powell I see there is a new milestone, should that be reflected here?

geoff-powell (approved):

jrodbx:
```
   1) Jupiter example tests (extensions, parametrized tests)
   2) Kotest example test
   3) TestNg example test
```

I don't think we should support every test framework, as this adds a wide surface area.  Should we limit to Junit 4 and 5?

nishatoma:
> ```
> 
>    1) Jupiter example tests (extensions, parametrized tests)
> 
>    2) Kotest example test
> 
>    3) TestNg example test
> 
> ```
> 
> 
> 
> I don't think we should support every test framework, as this supports a wide support surface area.  Should we limit to Junit 4 and 5?

@jrodbx can you elaborate? The test projects are just examples in this case. In practice, having 'setup' and 'teardown' APIs means the consumer maintains those test frameworks on their end as opposed to having it maintained by Paparazzi. What do you think?

karsie:
How come I can't find this code in a downloaded 2.0.0-alpha05?

matejdro:
I am successfuly using it on alpha 05, so something might be on your side?

karsie:
I was looking at this PR and obviously the beforeTest and afterTest methods have been renamed to setup and teardown

PR #2211, opened 2026-01-08
Move ASM fix from gradle instrumentation to ByteBuddy runtime instrumentation

geoff-powell:
AGP Instrumentation ASM changes for AGP Multiplatform Library plugin has a bug in 8.13.x for ASM transformations.
This is fixed in AGP 9.0 (see [here](https://github.com/cashapp/paparazzi/pull/2115#discussion_r2582313440) and [here](https://github.com/cashapp/paparazzi/issues/2136#issuecomment-3723454107)) but it would be beneficial for Paparazzi to move away from AGP apis to do this instrumentation. We already are using ByteBuddy to intercept other methods, so we can provide the same ASM changes at runtime


paparazzi/src/main/java/app/cash/paparazzi/agent/InterceptorRegistrar.kt
```diff
@@ -15,6 +27,11 @@ internal object InterceptorRegistrar {
 
   private val methodInterceptors = mutableListOf<() -> Unit>()
 
+  private const val RESOURCES_COMPAT_CLASS_NAME = "androidx.core.content.res.ResourcesCompat"
+  private const val LOAD_FONT_METHOD_NAME = "loadFont"
+  private const val LOAD_FONT_METHOD_DESCRIPTOR =
+    "(Landroid/content/Context;Landroid/content/res/Resources;Landroid/util/TypedValue;IILandroidx/core/content/res/ResourcesCompat\$FontCallback;Landroid/os/Handler;ZZ)Landroid/graphics/Typeface;"
```
colinmarsch, line 33: For my understanding, where does this value come from?

colinmarsch:
Is there a way to test / verify a change like this?

paparazzi/src/main/java/app/cash/paparazzi/agent/InterceptorRegistrar.kt
geoff-powell: These values come from the [ASM rewrite done in Layoutlib.](https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:rendering/src/com/android/tools/rendering/classloading/ResourcesCompatTransform.kt;drc=5bb41b6d5e519c891a4cd6149234138faa28e1af)

geoff-powell:
> Is there a way to test / verify a change like this?

We have a custom font test that specifically calls `ResourcesCompat.loadFont()` to ensure this works.
Without the ASM fixes, the test fails with an exception

jrodbx:
> This is fixed in AGP 9.0 but it would be beneficial for Paparazzi to move away from AGP apis to do this instrumentation. We already are using ByteBuddy to intercept other methods, so we can provide the same ASM changes at runtime

You might recall from an earlier sync that we actually would prefer to move away from Bytebuddy 😅 and move towards more ASM (as you did with the original ResourceFont implementation), since ASM has more documented samples, and I suspect that ByteBuddy misconfiguration causes more pain than it's worth (e.g., constant Robolectric clashes).

Do we have a sense on how long until the AGP fix is scheduled to land in release?  If we want to land this to fill the gap sooner, then I'm game, provided we revert as soon as it's fixed upstream in AGP, since the ASM approach is the better way.

Side note: IIRC, if we migrate the ViewEditMode interceptor to ASM, then Bytebuddy can be removed!

Side note 2: On the flip side, it would be nice to replace the AGP hooks and instead hook in the ASM stuff with the classloader.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -143,6 +143,7 @@ public class PaparazziSdk @JvmOverloads constructor(
   public fun setup() {
     if (!isInitialized) {
       registerViewEditModeInterception()
+      InterceptorRegistrar.registerResourcesCompatFontLoadFix()
```
jrodbx, line 146: (assuming this is a temporary fix that will be landed, vs closing this out until waiting for AGP fix to land)

let's move the resource compat bits out of InterceptorRegistrar, keeping it dumb, and let the SDK call a generic method similar to ViewEditMode?

geoff-powell:
> > This is fixed in AGP 9.0 but it would be beneficial for Paparazzi to move away from AGP apis to do this instrumentation. We already are using ByteBuddy to intercept other methods, so we can provide the same ASM changes at runtime
> 
> You might recall from an earlier sync that we actually would prefer to move away from Bytebuddy 😅 and move towards more ASM (as you did with the original ResourceFont implementation), since ASM has more documented samples, and I suspect that ByteBuddy misconfiguration causes more pain than it's worth (e.g., constant Robolectric clashes).
> 
> Do we have a sense on how long until the AGP fix is scheduled to land in release? If we want to land this to fill the gap sooner, then I'm game, provided we revert as soon as it's fixed upstream in AGP, since the ASM approach is the better way.
> 
> Side note: IIRC, if we migrate the ViewEditMode interceptor to ASM, then Bytebuddy can be removed!
> 
> Side note 2: On the flip side, it would be nice to replace the AGP hooks and instead hook in the ASM stuff with the classloader.

Yep so my thoughts moving forward align with
```
Side note 2: On the flip side, it would be nice to replace the AGP hooks and instead hook in the ASM stuff with the classloader.
```

This gets us to a place removing the reliance on AGP for ASM rewrites which is great!
My only motivation here is to land this to unblock multi-platform PR.

But with AGP at a RC it might not be worth it to try and land before alpha03

geoff-powell:
Closing for future ASM changes at runtime

Issue #2127, opened 2025-10-09
Tests pass despite exception thrown in Compose Launched Effect during rendering


elihart:
**Description**
When snapshotting compose UI code, the test will succeed even if a LaunchedEffect within the UI code throws an error. This can lead to part of the UI being missing. The error is printed in the test logs, but the test will still pass and the issue can easily be missed.

**Steps to Reproduce**

Snapshotting a composable with a launched effect that throws an error. The test won't fail.
```
@Composable
fun SomeUIWithAnEffect() {

    LaunchedEffect(Unit) {
            error("Throw an exception")
    }
}
```

**Expected behavior**
The test should fail with the thrown error. I expect the dispatcher set on the recomposer in the paparazzi sdk should attach a coroutine exception handler

**Additional information:**
- Paparazzi Version: 1.3.5

**Screenshots**
N/A

For now I am working around this by using a rule to detect uncaught exceptions. This seems to work to catch the coroutine errors.

```
@OptIn(ExperimentalAtomicApi::class)
class FailOnUncaughtExceptionRule : TestRule {
    private val throwable = AtomicReference<Throwable?>(null)

    override fun apply(base: Statement, description: Description) = object : Statement() {
        override fun evaluate() {
            val prev: Thread.UncaughtExceptionHandler? = Thread.getDefaultUncaughtExceptionHandler()
            Thread.setDefaultUncaughtExceptionHandler { t, e ->
                throwable.compareAndSet(null, e)
                prev?.uncaughtException(t, e)
            }
            try {
                base.evaluate()
                throwable.load()?.let { throw it }
            } finally {
                Thread.setDefaultUncaughtExceptionHandler(prev)
            }
        }
    }
}
```

nishatoma:
@elihart  we have similar issue as well in our org as we have Paparazzi + Robolectric tests sometimes throw exceptions silently, then later on fail in our pure JVM tests.

Do you want to put up a PR for review? Or if you would like, I could also open a PR for this issue, with credits towards you of course.

elihart:
@nishatoma feel free to put up a PR - no credit needed 👍

nishatoma:
@jrodbx @geoff-powell I wanted to get your thoughts on this before going ahead and making a PR, does this rule make sense to be added to Paparazzi?

geoff-powell:
Hmm, need to think through this some more but at a hight level, it makes sense to try and not swallow or obfuscate exceptions thrown during snapshot execution.

[...]

PR #2214, opened 2026-01-13
Fail test if exception is thrown during effect

colinmarsch:
This change ensures that exceptions thrown during test runs will be logged and that the snapshot test task will fail instead of silently succeeding. 

Resolves https://github.com/cashapp/paparazzi/issues/2127

### DIFF
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -1670,7 +1670,7 @@ class PaparazziPluginTest {
 
     assertThat(result.task(":testDebugUnitTest")).isNotNull()
     assertThat(result.output).contains("LaunchedEffectExceptionTest > launchedEffectExceptionPropagates FAILED")
-    assertThat(result.output).contains("java.lang.IllegalStateException")
+    assertThat(result.output).contains("java.lang.IllegalStateException: Exception thrown in LaunchedEffect")
   }
 
   @Test
--- a/paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/build.gradle
+++ b/paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/build.gradle
@@ -26,3 +26,9 @@ android {
 dependencies {
   implementation libs.composeUi.material
 }
+
+tasks.withType(Test).configureEach {
+  testLogging {
+    showStandardStreams true
+  }
+}


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1660,6 +1660,19 @@ class PaparazziPluginTest {
       .runFixture(fixtureRoot) { build() }
   }
 
+  @Test
+  fun composeLaunchedEffectExceptionPropagates() {
+    val fixtureRoot = File("src/test/projects/compose-launched-effect-exception")
+
+    val result = gradleRunner
+      .withArguments("testDebug", "--stacktrace")
+      .runFixture(fixtureRoot) { buildAndFail() }
```
nishatoma, line 1669: [minor, non-blocking]
Instead of failing the build, can we 'assertThrows' within the test instead?

Edit for clarity: I meant assertThrows from the test project (the one taking the snapshot), my apologies for the confusion!

nishatoma:
Thank you for fixing this! 

One thing I wanted to call out: you might wanna consider an optional build parameter or a system property. 

If you call this handler as is, then all consumers with silent exceptions will have failing tests immediately.

It would be a nice thing to have, the ability to gradually have "stricter" tests as you fix errors, especially in large codebases.

paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/src/test/java/app/cash/paparazzi/plugin/test/LaunchedEffectExceptionTest.kt
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+@Composable
+fun ComposableWithFailingLaunchedEffect() {
+  LaunchedEffect(Unit) {
+    error("Exception thrown in LaunchedEffect")
+  }
+  Text("Hello")
+}
```
geoff-powell: nit: can this be inlined with the Test class?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: I think `buildAndFail` is the right option here as the exception thrown in the code under test is turned into a logged error and a test failure instead of a re-thrown exception in the internal Paparazzi layer. Let me know if I've missed something here however!

colinmarsch:
> One thing I wanted to call out: you might wanna consider an optional build parameter or a system property.
> 
> If you call this handler as is, then all consumers with silent exceptions will have failing tests immediately.
> 
> It would be a nice thing to have, the ability to gradually have "stricter" tests as you fix errors, especially in large codebases.

That's a good callout, however I'm not sure the additional overhead of a new build parameter for this is worth it as the state of a test throwing an exception is likely already degraded and should be fixed. Curious your thoughts as well @geoff-powell @jrodbx

paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/src/test/java/app/cash/paparazzi/plugin/test/LaunchedEffectExceptionTest.kt
colinmarsch: Updated this to inline in the test function!

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
nishatoma: No worries! Suppose, the test fails for another reason, maybe due to `sdk` update or another breaking change, then how should we verify if it failed explicitly because of that coroutine, as opposed to something else? This is the reason I mentioned `assertThrows` might be helpful to check for that scenario explicitly and it makes you more confident on what failed exactly:)

colinmarsch: I see your point. In the current approach we are also verifying that the exception is shown in the output of the test task when it has failed, so that should give some extra safety that this is behaving correctly.

Unfortunately, I don't think `assertThrows` will help for this situation as the exception isn't being re-thrown, its just being turned into an error log in the test output. `paparazzi.snapshot` in the test here completes without exception but the `logger` contains errors after the test run which results in the failed build

paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/src/test/java/app/cash/paparazzi/plugin/test/LaunchedEffectExceptionTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * This test verifies that exceptions thrown in LaunchedEffect are properly propagated
+ * and cause the test to fail. See https://github.com/cashapp/paparazzi/issues/2127
+ *
+ * The test is expected to FAIL - the plugin test wraps this with buildAndFail() to
+ * verify that the exception propagation is working correctly.
+ */
```
jrodbx:
```suggestion
```

already explained in the implementation site

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: > Instead of failing the build, can we 'assertThrows' within the test instead?

AFAIK, the Gradle process will swallow the thrown exception preventing the runner/test code here from being able to assert on it.  If there is a way to propagate the error through, I think the additional assertion on that (along with the logged warning) would be nice.

jrodbx:
> > One thing I wanted to call out: you might wanna consider an optional build parameter or a system property.
> > If you call this handler as is, then all consumers with silent exceptions will have failing tests immediately.
> > It would be a nice thing to have, the ability to gradually have "stricter" tests as you fix errors, especially in large codebases.
> 
> That's a good callout, however I'm not sure the additional overhead of a new build parameter for this is worth it as the state of a test throwing an exception is likely already degraded and should be fixed. Curious your thoughts as well @geoff-powell @jrodbx

@nishatoma I think of this as more of a bug being fixed vs a feature being added, the distinction being that these tests should have already been failing.

Also, see my soon-to-be-posted comment re: implementation detail.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -306,8 +308,15 @@ public class PaparazziSdk @JvmOverloads constructor(
         // Since this dispatcher does not provide its own implementation of Delay, it will default to using DefaultDelay which runs
         // async to our test Handler. By initializing Recomposer with Dispatchers.Main, Delay will now be backed by our test Handler,
         // synchronizing expected behavior.
+        //
+        // Additionally, we attach a CoroutineExceptionHandler to ensure that uncaught exceptions from
+        // LaunchedEffect and other coroutines are propagated to the test framework rather than being
+        // silently swallowed. See https://github.com/cashapp/paparazzi/issues/2127
+        val exceptionHandler = CoroutineExceptionHandler { _, throwable ->
+          logger.error(throwable, throwable.message)
+        }
```
jrodbx: nice to implement this in the SDK, but this means that it'd be limited to coroutine-based exceptions, whereas the OP/underlying issue registers an generalized uncaught exception handler in the test rule, which would catch all exceptions.

I guess if we design this further, there's:
1) this implementation, CoroutineExceptionHandler in sdk
2) the OP suggestion, UncaughtExceptionHandler in rule
3) UncaughtExceptionHandler in sdk?

3 would allow for same behavior in potential future use cases that use the SDK directly, while also expanding the catch-all for non-coroutine exception cases.

As an addition, it might also be nice to allow that to part of the API, i.e., configurable, for consumers who want to route such exceptions to their reporting tool of choice (Bugsnag, Crashlytics, etc.).  Thoughts?

nishatoma: In the OP issue approach, after implementing that in our own codebase as a custom rule, it actually fails the test with any exception, not just coroutine, similar to runTest! We were able to fix some silently failing test cases. 

It also allowed us to write tests that explicitly assert the errors thrown from that custom rule!

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
nishatoma: Sorry to clarify, I meant 'assertThrows' from within the test project, (the test that is taking the snapshot), would that also swallow the exception?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: Thanks for pointing this out! I just updated this to use a similar approach with the `UncaughtExceptionHandler` here. As for supporting configurable reporting logic in here, I think thats a good feature however I think it could be extracted out of this PR. I'll note something down to track it for a follow up though!

paparazzi-gradle-plugin/src/test/projects/compose-launched-effect-exception/src/test/java/app/cash/paparazzi/plugin/test/LaunchedEffectExceptionTest.kt
colinmarsch: Removed this comment

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: I spent some time trying to get both paths working here (logging the exception in the paparazzi test output and being able to assert on the exception being thrown in the test itself) and wasn't able to figure out a clean way to support both at once.

For context, the below does not work in the current state as the `snapshot()` function does not throw an exception in this situation. The test is just failed due to the `logger.assertNoErrors()` call in the `teardown()` function [here](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt#L203).
```
val exception = assertThrows(IllegalStateException::class.java) {
  paparazzi.snapshot { ... }
}
```

If either of you have any thoughts on how to accomplish both paths I'm happy to try implementing them!

jrodbx: > would that also swallow the exception?

@nishatoma: @colinmarsch clarifies this above.

> Sorry to clarify, I meant 'assertThrows' from within the test project, (the test that is taking the snapshot)

...but even so, the project test fixtures are meant to mimic actual consumer usage, and the functional plugin tests are where the project repo actually tests behavior -- it's essentially a black box test.  so parsing the test output is the best way for now.

I guess one *could* instead provide an error log artifact that we control and assert against that independently, vs sifting through the Junit output, but probably not worth the minimal gain to refactor this at the time.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -1660,6 +1660,19 @@ class PaparazziPluginTest {
...
+
+    assertThat(result.task(":testDebugUnitTest")).isNotNull()
+    assertThat(result.output).contains("LaunchedEffectExceptionTest > launchedEffectExceptionPropagates FAILED")
+    assertThat(result.output).contains("java.lang.IllegalStateException")
```
jrodbx: nit: can we include the exception message "Exception thrown in LaunchedEffect" to avoid false positives from potentially other ISEs being thrown?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
nishatoma: > For context, the below does not work in the current state as the `snapshot()`

I believe you would need to essentially store the suppressed exception and then throw it after your test is done., kinda like so:

```
val unCaughtException = // this is null
Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
      logger.error(throwable, throwable.message)
      previousUncaughtExceptionHandler?.uncaughtException(thread, throwable)
      // assign it here
    }
```
After removing the logger, the test throws the exception as expected instead of logging (which confused me a bit as I saw the exception in red so I thought it was being thrown):

```
Thread.currentThread().uncaughtExceptionHandler = Thread.UncaughtExceptionHandler { thread, throwable ->
      unCaughtException = throwable
      previousUncaughtExceptionHandler?.uncaughtException(thread, throwable)
    }
```

And now this passes as expected:

```
@Test
  fun launchedEffectExceptionPropagates() {
    assertThrows(Throwable::class.java) {
      paparazzi.snapshot {
        LaunchedEffect(Unit) {
          error("Exception thrown in LaunchedEffect")
        }
        Text("Hello")
      }
    }
  }
```

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
nishatoma: @colinmarsch thank you for the quick feedback! just a qq, should we override the `Thread.defaultExceptionHandler` or `Thread.currentThread()`'s handler here?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: https://github.com/cashapp/paparazzi/pull/2214/changes/55c9dbeb5cbeadbf7eb2bf7f09a573a9fafd3810

colinmarsch: Thanks for adding the change!

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: Thanks for code snippets! I had tried something similar and ran into what you mentioned where the logger swallowed the exception being thrown so I couldn't effectively have both the logged error and the thrown exception from the test. Overall, I agree with John above that the current state treating the test as a black box is ideal.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: The `Thread.defaultExceptionHandler` should be what we want here in my understanding. Paparazzi does not control the threads created that might be throwing exceptions here.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
nishatoma: I have more clarity now, It takes me slightly longer to understand this so I appreciate the patience and discussion, awesome work:)

PR #2240, opened 2026-01-27
Remove byte buddy and apply ASM changes and Java Agent at runtime

geoff-powell:
* Adds a java agent for runtime and legacy separate jvm process agent to allow for instrumentation manipulation
* Moves ASM instrumentation logic applied through AGP to be applied at runtime
* Add ASM rewrite for `isInEditMode` method interceptor
* Removes ByteBuddy dependency


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 agp = "8.13.2"
 androidTools = "31.13.2" # == 23.0.0 + agp version
-bytebuddy = "1.18.4"
```
colinmarsch, line 4: 🎉

colinmarsch (approved):
What would be the process to verify a change like this?

geoff-powell:
> What would be the process to verify a change like this?

We have test coverage for both the ResourceCompat asm fix and the isEditMode tests so those specific projects would fail if this didn't work.
Will wait for @jrodbx to get his 👀 on it

paparazzi/src/main/java/app/cash/paparazzi/agent/ResourcesCompatTransform.kt
geoff-powell, line 1: Moved from `ResourcesCompatVisitorFactory`

PR #2241, opened 2026-01-27
Fix configuration-cache issue resolving `project` in provider mapping directly

geoff-powell:
image


geoff-powell:
> Do we need to change the test projects in this repo to catch these issues?

Yes, unfortunately I wasn't able to repro this issue in paparazzi alone. It required using cash-android to fix and unblock

geoff-powell:
Task with successful compilation https://gradle-enterprise.aws.squarecloudservices.com/s/oqkt3hsaneggi

Issue #2239, opened 2026-01-26
Create accessibility manifest to render accessibility info in test report


geoff-powell:
Currently Accessibility information requires using AccessibilityRenderExtension. It would be great to get accessibility support out of the box (opt out) by generating accessibility manifest for all snapshots and use it for `verifyPaparazzi` step to compare accessibility identifier changes.

PR #2254, opened 2026-02-05
Extract accessibility element collection into reusable collector

colinmarsch:
This change updates the accessibility element collection logic to refactor it into a reusable class to support efforts towards the work described here: https://github.com/cashapp/paparazzi/issues/2239.


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityElementCollector.kt
geoff-powell, line 1: Love that this was pulled out of the render extension.
Maybe we can separate a little more like so:
* Pull data from views/semanticsNodes to own class `AccessibilityElement`
* Sorting logic for elements that specify previous/next element
* Parsing logic for `AccessibilityElement` -> String representation

geoff-powell: Can be done in follow up to this

colinmarsch: Yes! Breaking this down further makes sense to me, I'll handle these in follow ups.

PR #2260, opened 2026-02-07
Add accessibility traversal neighbor metadata

colinmarsch:
This change adds before/after metadata to the `AccessibilityElement` for verifying the ordering of the accessibility hierarchy via the metadata of the elements. This is part of the work to contribute to the implementation of https://github.com/cashapp/paparazzi/issues/2239


geoff-powell (approved):

colinmarsch:
@geoff-powell After your review I added an additional test case to verify that the ordering is correctly represented from taking an actual snapshot

oldergod (approved):
beforeElementId and afterElementId are copied from AccessibilityElement.id which label derived, right? thus uniqueness isn't guaranteed I believe (like 2 "OK" buttons in the screen)

We should do something to confirm we're always traversing in the same order, either some kind of prefix or fallback when we have dups.

geoff-powell (approved):
LGTM!

PR #2261, opened 2026-02-07
Move detectMaxPercentDifferenceDefault method to Paparazzi.kt

nishatoma:
Hello! As part of feedback from here:

https://github.com/cashapp/paparazzi/pull/2203/changes/BASE..82b4fe2b566ffb055ea434af9b65e2c46f5c5ea1#diff-2c10da07633a221584f8c9dd4d75419da14f458fbc5d0eed23cba5f49a530273R44

This is a follow-up PR that moves the `detectMaxPercentDifferenceDefault` to `Paparazzi.kt.`

@jrodbx let me know if you have any additional feedback on that.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
@@ -21,6 +21,7 @@ import android.view.LayoutInflater
 import android.view.View
 import androidx.annotation.LayoutRes
 import androidx.compose.runtime.Composable
+import app.cash.paparazzi.Paparazzi.PaparazziUtils.detectMaxPercentDifferenceDefault
 import com.android.ide.common.rendering.api.SessionParams.RenderingMode
 import org.junit.rules.TestRule
 import org.junit.runner.Description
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -187,5 +187,8 @@ public class Paparazzi @JvmOverloads constructor(
       } else {
         HtmlReportWriter(maxPercentDifference = maxPercentDifference)
       }
+
+    private fun detectMaxPercentDifferenceDefault(): Double =
```
geoff-powell: nit: still need this public. We use this method at cashapp

paparazzi/src/main/java/app/cash/paparazzi/Paparazzi.kt
```diff
@@ -188,4 +189,9 @@ public class Paparazzi @JvmOverloads constructor(
         HtmlReportWriter(maxPercentDifference = maxPercentDifference)
       }
   }
+
+  public object PaparazziUtils {
```
geoff-powell, line 193: nit: this seems unnecessary to wrap this public fun in an object. Can we just make it class level?

nishatoma: Oh okay, did you mean something like this?
```
public fun detectMaxPercentDifferenceDefault(): Double =
    System.getProperty("app.cash.paparazzi.maxPercentDifferenceDefault")?.toDoubleOrNull() ?: 0.01
```

It would be an instance method within `Paparazzi()` is that what you meant? If so, it would not be usable anymore as is on lines 37 and 56 within the constructor.

Or do you mean like a top-level function within `Paparazzi.kt` file itself?

PR #2268, opened 2026-02-09
Generate JSON accessibility hierarchy after snapshot completion

colinmarsch:
This changes adds JSON generation logic for the accessibility metadata that can be used to create a diff between the previously stored golden accessibility hierarchy diff and the current validation run. This storage of the metadata JSON on record and validation during verify will be added in a subsequent PR behind an opt-in gradle configuration flag.

Part of the changes for https://github.com/cashapp/paparazzi/issues/2239.

---
**Update Aug 4, 14:12 PDT:** Moved hierarchy generation out of `RenderExtension` following review feedback.
- Generation is controlled by `app.cash.paparazzi.accessibility.hierarchy.artifacts.enabled` and remains disabled by default.
- Paparazzi collects the original content hierarchy per rendered frame, then reports the frame list after the snapshot run, including image verification failure paths.
- Record, verify, and report persistence remain follow-up work.
- Verified with focused rendering and accessibility tests, API compatibility checks, and Spotless.

Generated with Codex

## Property ownership update (2026-08-04)

This PR now owns the single public property key, `Flags.ACCESSIBILITY_HIERARCHY_ARTIFACTS_ENABLED`, as well as the generation gate that consumes it. The generator references that constant directly, so the key is declared once.

## Update (2026-08-12)

Accessibility hierarchy collection no longer uses `WindowManager`, `WindowManagerImpl`, or `currentRootView`. `PaparazziSdk` supplies all ordered Layoutlib render roots to the generator for each image frame, the collector traverses each root exactly once, and serialization consumes the resulting `AccessibilityFrame` elements. Native dialog output remains byte-for-byte compatible, with added coverage for dialog, popup, and primary roots appearing exactly once in supplied order.


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -68,12 +82,19 @@ public class AccessibilityRenderExtension : RenderExtension {
...
           )
+          collectedElements = elements
           overlayDrawable.updateElements(elements)
           overlayDetailsView.updateElements(elements)
         }
       }
     }
   }
+
+  internal fun onSnapshotRunCompleted() {
```
geoff-powell: hmm, I wonder if this should be part of `RenderExtension` interface. Not to block this, just wanted to pose the question.

geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
colinmarsch: I just updated this branch to make this change. I agree that this is likely a cleaner way to go about this in a more standardized pattern rather than filtering for the `AccessibilityRenderExtension` instances in the PaparazziSdk logic

oldergod (approved):
Same uniqueness problem that #2260, otherwise good

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -371,6 +371,8 @@ public class PaparazziSdk @JvmOverloads constructor(
         }
         onNewFrame(scaleImage(frameImage(image)))
       }
+
+      renderExtensions.forEach { it.onSnapshotRunCompleted() }
```
oldergod: btw we won't go through this if the diff fails above `onNewFrame` for this one throws in verify mode when not the same

geoff-powell:
Hmm, ya lets try and move this to another mechanism and move away from RenderExtension.

Maybe add a hook to control this via a gradle prop and disable it by default. We can enable it in a follow up PR with Jrod's test reporting work.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -330,48 +335,73 @@ public class PaparazziSdk @JvmOverloads constructor(
...
-              "Pending work detected. This may cause unexpected results in your generated snapshots. ${recomposerInstance.changeCount}"
+          val image = bridgeRenderSession.image
+          if (validateAccessibility) {
+            require(renderExtensions.isEmpty()) {
+              "Running accessibility validation and render extensions simultaneously is not supported."
+            }
+            validateLayoutAccessibility(modifiedView, image)
+          }
+          val windowRoots = bridgeRenderSession.rootViews.map { it.viewObject as View }
+          accessibilityHierarchyGenerator.generate(
```
colinmarsch, line 383: This is the main update in my latest commit

PR #2277, opened 2026-02-12
Add generic artifact plumbing hooks

colinmarsch:
This changes adds artifact plumbing hooks for use in https://github.com/cashapp/paparazzi/pull/2278 for the accessibility hierarchy json string file to be plumbed through. Part of the work for https://github.com/cashapp/paparazzi/issues/2239

## Update (2026-08-04)

Artifact delivery now uses an internal `ArtifactFrameHandler` capability instead of extending the public `SnapshotHandler.FrameHandler` API. This removes the RenderExtension completion hook and PaparazziSdk constructor churn while leaving an internal sink for the record/verify work above it.

Verified with the focused `ArtifactFrameHandlerTest`, `:paparazzi:apiCheck`, and `:paparazzi:spotlessCheck`.


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -92,7 +92,8 @@ public class PaparazziSdk @JvmOverloads constructor(
   private val supportsRtl: Boolean = false,
   private val showSystemUi: Boolean = false,
   private val useDeviceResolution: Boolean = false,
-  private val onNewFrame: (BufferedImage) -> Unit
+  private val onNewFrame: (BufferedImage) -> Unit,
+  private val onArtifact: (String, String) -> Unit = { _, _ -> }
 ) {
   private var validateAccessibility = false
```
oldergod: This is a breaking change. unless we're happy doing it, we could create another constructor to decrease the damage and define `onArtifact` as a private property.


```suggestion
  private val onNewFrame: (BufferedImage) -> Unit,
) {
  private var validateAccessibility = false
  private var onArtifact: (String, String) -> Unit = { _, _ -> }

  public constructor(
    environment: Environment,
    deviceConfig: DeviceConfig,
    theme: String,
    renderingMode: RenderingMode,
    appCompatEnabled: Boolean,
    renderExtensions: Set<RenderExtension>,
    supportsRtl: Boolean,
    showSystemUi: Boolean,
    useDeviceResolution: Boolean,
    onNewFrame: (BufferedImage) -> Unit,
    onArtifact: (String, String) -> Unit
  ) : this(
    environment,
    deviceConfig,
    theme,
    renderingMode,
    appCompatEnabled,
    renderExtensions,
    supportsRtl,
    showSystemUi,
    useDeviceResolution,
    onNewFrame
  ) {
    this.onArtifact = onArtifact
  }
```

oldergod (approved):

geoff-powell:
Hmm not sure if this needs to be run once for gif. Its a poor usecase but I would assume that we could do one of the following for multiple frames. 
1. Create a json file per rendered frame (seems like overkill)
2. Create a json file with the root element as a list / key value pair of frame number -> Accessibility Json Blob

I like 2 personally but open to any other thoughts here.

PR #2278, opened 2026-02-12
Add opt-in accessibility artifact record and verify support

colinmarsch:
This change adds opt-in functionality (behind a gradle configuration property) to handle recording and verifying an accessibility hierarchy json string artifact for the https://github.com/cashapp/paparazzi/issues/2239 feature.

Example output as follows:
```
[
  {
    "id": "Number 1",
    "beforeElementId": null,
    "afterElementId": "Number 2",
    "bounds": {
      "left": 57,
      "top": 57,
      "right": 215,
      "bottom": 616
    },
    "legendText": "Number 1"
  },
  {
    "id": "Number 2",
    "beforeElementId": "Number 1",
    "afterElementId": "Number 3",
    "bounds": {
      "left": 57,
      "top": 680,
      "right": 215,
      "bottom": 1239
    },
    "legendText": "Number 2"
  },
  {
    "id": "Number 3",
    "beforeElementId": "Number 2",
    "afterElementId": null,
    "bounds": {
      "left": 57,
      "top": 1303,
      "right": 215,
      "bottom": 1863
    },
    "legendText": "Number 3"
  }
]
```

## Update (2026-08-04)

Accessibility JSON is generated by the Paparazzi snapshot lifecycle when `app.cash.paparazzi.accessibility.hierarchy.artifacts.enabled=true`, then routed to the built-in record/verify handlers through the internal artifact capability. Generation is disabled by default and no longer depends on `AccessibilityRenderExtension`; using that extension remains compatible but is not what generates the JSON.

Focused hierarchy lifecycle, report writer, and verifier tests pass. API and formatting checks also pass.

## Stack update (2026-08-04)

This PR now also owns the internal `ArtifactFrameHandler` capability previously proposed in #2277. It is stacked directly on #2268, so the public API remains unchanged while the record/verify implementation and its internal delivery boundary are reviewed together.

## Property ownership update (2026-08-04)

The property declaration and public API addition now live entirely in #2268. This PR inherits that property and is limited to the internal artifact sink plus record, verify, and report persistence.

## Update (2026-08-12)

Restacked on the ordered render-root hierarchy implementation from #2268. Added coverage for one hierarchy per GIF frame, late user callback assignment without replacing artifact delivery, independent artifact and callback failure handling, and cleanup of stale actual/diff artifacts after a successful retry. Core tests, API checks, and formatting checks pass.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
@@ -70,9 +70,11 @@ public class HtmlReportWriter @JvmOverloads constructor(
   private val runsDirectory: File = File(rootDirectory, "runs")
   private val imagesDirectory: File = File(rootDirectory, "images")
   private val videosDirectory: File = File(rootDirectory, "videos")
+  private val reportArtifactsDirectory = File(rootDirectory, ARTIFACTS_DIRECTORY_NAME)
 
   private val goldenImagesDirectory = File(snapshotRootDirectory, "images")
   private val goldenVideosDirectory = File(snapshotRootDirectory, "videos")
+  private val goldenArtifactsDirectory = File(snapshotRootDirectory, ARTIFACTS_DIRECTORY_NAME)
 
   private val shots = mutableListOf<Snapshot>()
 
@@ -85,6 +87,8 @@ public class HtmlReportWriter @JvmOverloads constructor(
     runsDirectory.mkdirs()
     imagesDirectory.mkdirs()
     videosDirectory.mkdirs()
+    reportArtifactsDirectory.mkdirs()
+    goldenArtifactsDirectory.mkdirs()
     writeStaticFiles()
     writeRunJs()
     writeIndexJs()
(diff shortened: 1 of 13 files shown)


paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -85,6 +87,8 @@ public class HtmlReportWriter @JvmOverloads constructor(
     runsDirectory.mkdirs()
     imagesDirectory.mkdirs()
     videosDirectory.mkdirs()
+    reportArtifactsDirectory.mkdirs()
+    goldenArtifactsDirectory.mkdirs()
```
oldergod: This creates a folder unconditionally during initialization in all cases but this mutates the snapshot root just by constructing the writer, regardless of whether artifact recording is enabled or any artifact is written. `goldenFile.parentFile.mkdirs()` inside the `isRecording` below is good enough I think

oldergod (approved):

geoff-powell:
Yes like that its opt-in initially. Once we deprecate the render extension this should be used to ensure accessibility changes. If turned off, we can provide the accessibility information within the test report and generate the json in our paparazzi reports folder.

PR #2310, opened 2026-03-17
Replace byte-buddy with ASM for runtime method interception

colinmarsch:
## Summary

Byte-buddy (`byte-buddy` + `byte-buddy-agent`, ~3.5 MB) was used solely for one thing: redefining `View.isInEditMode()` at runtime via `InterceptorRegistrar`. This replaces both dependencies with direct ASM bytecode transformation and the JDK Attach/Instrumentation APIs.

**Why:** byte-buddy is a large, general-purpose bytecode library. Paparazzi's interception needs are minimal (replace a method body with a static delegate call), so ASM's lower-level API is a better fit — smaller dependency footprint, fewer transitive deps, and more transparent bytecode manipulation.

### What changed

- **`AgentInstaller.kt`** (new) — self-attaches to the current JVM via `VirtualMachine.attach()` to obtain a `java.lang.instrument.Instrumentation` instance. Replaces `ByteBuddyAgent.install()`.
- **`InterceptorRegistrar.kt`** (rewritten) — uses ASM `ClassReader`/`ClassWriter`/`ClassVisitor` to rewrite target method bodies with `INVOKESTATIC` calls to interceptor classes. Uses `Instrumentation.retransformClasses()` (not `redefineClasses`) to apply transforms, which ensures compatibility with classes already modified by build-time ASM transforms.
- **`PaparazziSdk.kt`** — removed `ByteBuddyAgent.install()` call; agent installation is now handled internally by `InterceptorRegistrar.registerMethodInterceptors()`.
- **`PaparazziPlugin.kt`** — adds `-Djdk.attach.allowAttachSelf=true` and `-XX:+EnableDynamicAgentLoading` JVM flags to consumer test tasks (required for the JDK Attach API).
- **`libs.versions.toml` / `build.gradle`** — replaced `bytebuddy-agent` + `bytebuddy-core` with `org.ow2.asm:asm:9.7.1`.

## Test plan

- [x] `InterceptorRegistrarTest` passes — verifies ASM-based method interception and class retransformation
- [x] All `:paparazzi:test` tests pass
- [x] `:sample:testDebug` passes — full end-to-end: agent install → ASM transform of `android.view.View.isInEditMode()` → layoutlib rendering → snapshot comparison

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 agp = "8.13.2"
 androidTools = "31.13.2" # == 23.0.0 + agp version
-bytebuddy = "1.18.7"
+asm = "9.7.1"
 compose = "1.10.5"
 coroutines = "1.10.2"
 javaTarget = "21"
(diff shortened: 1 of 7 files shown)


saket (approved):

geoff-powell (approved):
Awesome, thanks for tackling this. Had this change locally once the AGP 9 stuff was merged but glad to see its up and ready!

PR #2009, opened 2025-06-23
Introduce additional differ types

jrodbx:
Closes https://github.com/cashapp/paparazzi/issues/1568.

Crafted mostly by GPT-4o, prompted with the following resources.  In early anecdotal measurements (i.e., measureTime), it looks like SIFT (or my implementation) isn't very performant/usable.  MSSIM/FLIP/DE2000 look better, but would need to dig in more.

Resources:
SSIM
* https://medium.com/@akp83540/structural-similarity-index-ssim-c5862bb2b520
* https://ece.uwaterloo.ca/~z70wang/publications/ssim.pdf
 
DeltaE
* https://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e
* CIEDE2000 variation: https://hajim.rochester.edu/ece/sites/gsharma/ciede2000/ciede2000noteCRNA.pdf

FLIP
* https://research.nvidia.com/publication/flip
* https://research.nvidia.com/sites/default/files/node/3260/FLIP_Paper.pdf
* https://github.com/NVlabs/flip

SIFT
* https://www.cs.ubc.ca/~lowe/papers/iccv99.pdf
* https://www.cs.ubc.ca/~lowe/papers/ijcv04.pdf
* https://docs.opencv.org/4.x/da/df5/tutorial_py_sift_intro.html

PR #2315, opened 2026-04-08
Fix perceptual differs

### LINKED ISSUE #2314: New differs use `Similar` incorrectly
**Description**
The `OffByTwo` differ uses the `Similar` result to mean that there are pixels in the image that are slightly different (off by two) from the target image. That can be the case if just a single pixel is similar or if every pixel is. The `Similar` result is interpreted by Paparazzi to be a passed test, since it's meant to cover slight differences due to things like cross-platform rendering differences. If even one pixel is different by more than two, the differ returns `Different`.

It seems like the new differs use `Similar` to mean that there is a small fraction of the images that are entirely different. Here's the implementation for `DeltaE2000`:

```kotlin
return when {
  percentDifference == 0f -> DiffResult.Identical(deltaImage)
  percentDifference < 5f -> DiffResult.Similar(deltaImage, numSimilar)
  else -> DiffResult.Different(deltaImage, percentDifference, numDifferent)
}
```

Since `Similar` is actually a passing result, this essentially puts a floor on `maxPercentDifference` of 5%.

**Steps to Reproduce**
Using `app.cash.paparazzi.differ=de2000`, change the text of `HelloPaparazzi` in `HelloComposeTest` to something else. So long as it doesn't cause a line break, it will probably pass because the image won't change by more than 5%.

**Expected behavior**
The test should fail because we shouldn't report different images that are even very close as similar. "Similar" should only mean that the two images have *pixels* which are similar, not the images themselves.

The fix should be to return `Similar` only if all the different pixels are within the perceptible difference.

**Additional information:**
- Paparazzi Version: 2.0.0-alpha04

VirtualParticle:
Fixes #2314

The new differs added in #2009 return `Similar` (a passing result) when the images have some degree of difference. They should only do that when some *pixels* are similar. If even one pixel is sufficiently different, we should return `Different`.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/differs/DeltaE2000.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/differs/DeltaE2000.kt
@@ -15,6 +15,9 @@ import kotlin.math.sin
 import kotlin.math.sqrt
 
 internal object DeltaE2000 : Differ {
+
+  private const val THRESHOLD = 2.3 // JND threshold for perceptible difference
+
   override fun compare(expected: BufferedImage, actual: BufferedImage): DiffResult {
     require(expected.width == actual.width && expected.height == actual.height)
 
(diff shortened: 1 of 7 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/differs/DeltaE2000.kt
```diff
@@ -31,17 +34,17 @@ internal object DeltaE2000 : Differ {
...
-        val lab1 = rgbToLab(rgb1)
-        val lab2 = rgbToLab(rgb2)
-
-        val deltaE = deltaE2000(lab1, lab2)
-
-        val isDifferent = deltaE > 2.3 // JND (just noticeable difference) threshold
-
-        if (isDifferent) {
-          numDifferent++
+        val deltaE = if (rgb1 != rgb2) {
```
VirtualParticle, line 37: This is an optimization to avoid doing the more complex logic if the colors are identical.

PR #2319, opened 2026-04-11
Fix flaky overwriteSnapshotOnMaxPercentDiff test

greg7gkb:
## Summary

- `overwriteSnapshotOnMaxPercentDiff` had a copy-paste bug where `recordLastModified` was capturing `dontRecordFile.lastModified()` instead of `recordFile.lastModified()`
- This caused a flaky assertion on CI: when both files share the same checkout timestamp (as they always do on a fresh `git checkout`), and the build completes within the same filesystem timestamp resolution, the `isNotEqualTo` check would spuriously fail
- One-line fix: capture the correct file's timestamp

## Test plan

- [ ] Verify `overwriteSnapshotOnMaxPercentDiff` passes consistently across all platforms/JDK combos in CI


geoff-powell (approved):
LGTM, thanks for the contribution!

Issue #2322, opened 2026-04-21
accept workflow - verifyPaparazziDebug actual images not identical with recordPaparazziDebug images


eboudrant:
Paparazzi's documented accept workflow (`mv actual.png golden.png`, printed on every failure) changes the repo with byte-different PNGs for same images Basically same pixels → different PNG bytes → `mv actual golden` always changes the file hash → LFS/git shows a modification for every accepted file.

Root cause: two different PNG encoders on the two sides:
- **Record** (`HtmlReportWriter`) writes goldens via `ApngWriter` — `Deflater.BEST_COMPRESSION` (level 9), row filter `None`.
- **Verify** (`ImageUtils.assertImageSimilar`) writes `actual.png` via `ImageIO.write(image, "PNG", …)` — stock Java PNG plugin, default zlib level, adaptive filters.

Simple step to repro : 
- run `verifyPaparazziDebug`
- accept the new actual images with `mv`
- commit to git
- run `recordPaparazziDebug`
- git is dirty, visually same images but diff exists because because different compression, etc...

PR #2323, opened 2026-04-21
Make verifyPaparazziDebug actual.png identical to recorded png

eboudrant:
Fix for #2322

Simple fix, just use `ApngWriter` to write the actual png.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/ImageUtils.kt
@@ -20,6 +20,8 @@ import app.cash.paparazzi.Differ
 import app.cash.paparazzi.Differ.DiffResult.Different
 import app.cash.paparazzi.Differ.DiffResult.Identical
 import app.cash.paparazzi.Differ.DiffResult.Similar
+import app.cash.paparazzi.internal.apng.ApngWriter
+import okio.Path.Companion.toPath
 import java.awt.AlphaComposite
 import java.awt.Color
 import java.awt.Graphics2D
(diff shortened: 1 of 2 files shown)


jrodbx (commented):
could re-recording the golden values as canonical source of truth work here?  ApngWriter was introduced for video recordings, and I think it doesn't result in the same binary output for single frames/images.  That in itself is worth exploring to potentially merge the image/video paths if you're interested!

eboudrant:
> could re-recording the golden values as canonical source of truth work here? ApngWriter was introduced for video recordings, and I think it doesn't result in the same binary output for single frames/images. That in itself is worth exploring to potentially merge the image/video paths if you're interested!

It could but need to update the message that suggest the `mv` command when failed. I'd love the move approach because it save a re-record pass (can be expensive on large project).

But regardless, it seems the actual in the build folder should match perfectly the recorded golden?

oldergod:
> ApngWriter doesn’t result in the same binary output for single frames/images
if the recorded single-frame golden is already produced by ApngWriter, why would we have a different binary?

The fix looks good to me. I created a test here that confirm the behavior is fixed https://github.com/eboudrant/paparazzi/pull/10

oldergod:
Let's merge tomorrow unless we have something better until then

eboudrant:
> ApngWriter was introduced for video recordings, and I think it doesn't result in the same binary output for single frames/images.

From my testing (and looking at the code again) looks like `ApngWriter` is used for single frame image. And when I found the issue I was testing using single frame image (no `paparazzi.gif { ... }`) . I'm happy to plan a refactor for later but so far looks like safe to merge to me too.

oldergod (approved):

Issue #1922, opened 2025-04-11
Content Resolver implementation changed in unrelated unit tests


sromanuk:
**Description**
Some of our existing unit tests have need to mock `contentResolver.query` method. For that we are using MockK, which worked right until Paparazzi was introduced into the module. With it we are getting the following exception in every single test:
```
java.lang.UnsatisfiedLinkError: 'java.lang.String android.os.SystemProperties.native_get(java.lang.String, java.lang.String)'
```
This error appeared because implementation of ContentResolver is switched to `layoutlib` for some reason, even though the test itself is not a screenshot test and Paparazzi rule is nowhere near this test.

**Steps to Reproduce**
1. have tests which rely on mocking of the `contentResolver.query` with MockK
2. Add paparazzi screenshot tests into the same module
3. Run the tests from step 1
4. Rejoice at seeing this exception

**Expected behavior**
Unit tests are working properly without the need too find a workaround, since they are unchanged by the addition of Paparazzi.

**Steps to work around this issue**
For some unknown reason we need to add paparazzi test rule to the test class and mock not only `ContentResolver`, but also a `Cursor`.

You can add the following to the test class, even if it does not make any sense to do so:
```kotlin
    @MockK lateinit var cursor: Cursor
    @get:Rule
    val paparazzi = Paparazzi(
        deviceConfig = DeviceConfig.PIXEL_4A,
        renderingMode = SessionParams.RenderingMode.SHRINK,
        renderExtensions = emptySet(),
        maxPercentDifference = 0.0,
    )

    @Before
    fun before() {
        MockKAnnotations.init(this)

        every { app.contentResolver.query(any(), any(), any(), any(), any()) } returns cursor
        every { app.contentResolver.getType(any()) } returns Constants.MIME_TYPE_JPEG

        every { cursor.moveToFirst() } returns true
        every { cursor.getLong(any()) } returns 0
        every { cursor.getString(any()) } returns ""
        every { cursor.getColumnIndex(any()) } returns 0
        every { cursor.getColumnIndexOrThrow(any()) } returns 0
        every { cursor.close() } just Runs
    }

    @After
    fun after() {
        unmockkAll()
    }
``` 

**Additional information:**
- Paparazzi Version: 1.3.5
- OS: macOS 15.4
- Compile SDK: 35
- Gradle Version: 8.13
- Android Gradle Plugin Version: 8.9.1

**Stacktrace**
```
[...]

PR #2324, opened 2026-04-30
Isolate Paparazzi test execution

colinmarsch:
## Why
Running Paparazzi through the Android unit test task can load Robolectric or other regular unit test state into the same JVM, which can break layoutlib initialization and fail snapshot verification. This change is a replacement for https://github.com/cashapp/paparazzi/pull/2313, which was a different approach to resolve the same issue https://github.com/cashapp/paparazzi/issues/1922.

## What
- Add a dedicated `testPaparazzi<Variant>` task for record and verify flows
- Copy unit test classpath and execution options onto the isolated Paparazzi task
- Filter the isolated task to Paparazzi snapshot tests, including inherited base classes on the runtime classpath
- Add a regression fixture with regular Robolectric tests and inherited Paparazzi tests

Verified in Cash App Android to confirm this resolves paparazzi build failures for us when running the latest paparazzi master version internally on modules with both robolectric and paparazzi tests.


saket (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -244,81 +249,286 @@ public class PaparazziPlugin @Inject constructor(
...
+      project.gradle.projectsEvaluated {
+        val unitTestTask = testTaskProvider.singleOrNull()
+        if (unitTestTask != null) {
+          paparazziTestTaskProvider.configure {
+            it.configureExecutionOptionsFromUnitTestTask(unitTestTask)
+          }
+        }
+      }
+
+      fun Test.configureForPaparazzi() {
```
geoff-powell, line 281: nit: is this change need for the feature. Lets make this diff smaller to better understand the changes.

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -244,81 +249,286 @@ public class PaparazziPlugin @Inject constructor(
       val overwriteOnMaxPercentDifferenceProvider = project.overwriteOnMaxPercentDifferenceProvider()
       val failureDir = buildDirectory.dir("paparazzi/failures")
       val testTaskProvider = testTasks.withType(Test::class.java)
-      testTaskProvider.configureEach { test ->
+      val paparazziTestTaskProvider = project.tasks.register("testPaparazzi$variantSlug", Test::class.java) {
```
geoff-powell, line 252: nit: Would like to follow the old task name structure here. Can we update this to `testPaparazzi${variantSlug}UnitTest`?

geoff-powell:
Updating context here.
@jrodbx @colinmarsch and I talked through this feature with the option of using a separate test source set, similar to how compose screenshot testing plugin works. We opted against it as it requires a lot more gradle work to get test reporting and all the features we currently support. Still investigating this approach as the collisions between Paparazzi mocked android frameworks and others (Robolelectric, etc) are high priority.

PR #2325, opened 2026-05-01
Destroy lifecycle owner after snapshots

### LINKED ISSUE #1867: UI stops recomposing after first test when using animations in lifecycle coroutine scope
**Description**
When paparazzi's injected lifecycle is used for launching an animation with a frame clock, the UI stops recomposing in subsequent tests. Since paparazzi shares the same renderer instance across all tests, could some important state be failing to reset correctly?

**Steps to Reproduce**
When the two tests in the following code are run, they'll produce two animated snapshots. The first one will display an animated counter whereas the second one will remain stuck at `0`.

```
class FooTest {
  @get:Rule
  val paparazzi = Paparazzi()

  @Test fun first() {
    takeSnapshot()
  }

  @Test fun second() {
    takeSnapshot()
  }

  private fun takeSnapshot() {
    paparazzi.gif {
      val counter by produceState(initialValue = 0) {
        while (isActive) {
          delay(100)
          this.value += 1
          println("writing value = ${this.value}")
        }
      }

      SideEffect {
        println("reading value = $counter")
      }

      val lifecycleOwner = LocalLifecycleOwner.current
      DisposableEffect(Unit) {
        val animatable = Animatable(0f)
        onDispose {
          lifecycleOwner.lifecycle.coroutineScope.launch {
            withContext(AndroidUiDispatcher.Main) {
              animatable.animateTo(1f)
            }
          }
        }
      }

      Text(
        modifier = Modifier
          .fillMaxSize()
          .background(Color(0xFF073858))
          .padding(16.dp),
        text = "Counter = $counter",
        color = Color.Yellow,
      )
    }
  }
}
```

**Additional information:**
- Paparazzi Version: https://github.com/cashapp/paparazzi/commit/1d2e1db453216c3329a232d7109813e6e6b9bbe7
- OS: macOS
- Compile SDK: 35
- Gradle Version: 8.12.1
- Android Gradle Plugin Version: 8.8.1

colinmarsch:
The `PaparazziLifecycleOwner` was never being reset after each test run and ended up being stuck at `RESUMED` after the first `gif` run. Due to this, the clean up in `onDispose`, (for example in the repro given in the linked issue) would be run but the lifecycle coroutine scope stayed alive and the disposal-time animation could keep running. That leftover coroutine could interfere with the next GIF.

## Summary of changes
- move Paparazzi's synthetic lifecycle owner to DESTROYED during snapshot cleanup
- add a Compose gif regression fixture for lifecycle coroutine work launched during disposal
- add the fixture-only Compose animation dependency via the version catalog

Fixes https://github.com/cashapp/paparazzi/issues/1867

### DIFF
--- a/paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
+++ b/paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
@@ -1,0 +1,77 @@
+package app.cash.paparazzi.plugin.test
+
+import androidx.compose.animation.core.Animatable
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.Text
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.produceState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.LocalLifecycleOwner
+import androidx.lifecycle.coroutineScope
+import app.cash.paparazzi.Paparazzi
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+
+class LifecycleCoroutineScopeTest {
+  @get:Rule
+  val paparazzi = Paparazzi()
+
+  @Test fun lifecycleCoroutineDoesNotBlockSubsequentGifs() {
+    assertCounterAdvances()
+    assertCounterAdvances()
+  }
+
+  private fun assertCounterAdvances() {
+    var maxCounter = 0
+
+    val view = ComposeView(paparazzi.context)
+    view.setContent {
+      val counter by produceState(initialValue = 0) {
+        while (isActive) {
+          delay(100)
+          value += 1
+        }
+      }
+
+      SideEffect {
+        if (counter > maxCounter) {
+          maxCounter = counter
+        }
+      }
+
+      val lifecycleOwner = LocalLifecycleOwner.current
+      DisposableEffect(Unit) {
+        val animatable = Animatable(0f)
+        onDispose {
+          lifecycleOwner.lifecycle.coroutineScope.launch {
+            animatable.animateTo(1f)
+          }
+        }
+      }
+
+      Text(
+        modifier = Modifier
+          .fillMaxSize()
+          .background(Color(0xFF073858))
+          .padding(16.dp),
+        text = "Counter = $counter",
+        color = Color.Yellow
+      )
+    }
+
+    paparazzi.gif(view, end = 500L, fps = 10)
+
+    assertTrue("Expected counter to advance, maxCounter=$maxCounter", maxCounter >= 1)
+  }
+}
--- a/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
@@ -288,6 +288,8 @@ public class PaparazziSdk @JvmOverloads constructor(
       previousUncaughtExceptionHandler?.uncaughtException(thread, throwable)
     }
 
+    lateinit var lifecycleOwner: PaparazziLifecycleOwner
+
     try {
       withTime(0L) {
         // Initialize the choreographer at time=0.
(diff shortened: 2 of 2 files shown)


saket (approved):
Good find!

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -288,6 +288,8 @@ public class PaparazziSdk @JvmOverloads constructor(
       previousUncaughtExceptionHandler?.uncaughtException(thread, throwable)
     }
 
+    var lifecycleOwner: PaparazziLifecycleOwner? = null
```
jrodbx:
```suggestion
    lateinit var lifecycleOwner: PaparazziLifecycleOwner
```

avoid the nullable type?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.withContext
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+@Suppress("DEPRECATION")
```
jrodbx: what's deprecated?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.math.max
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.withContext
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
jrodbx: is test ordering needed?  looks like `a` and `b` do the same thing?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
```diff
@@ -0,0 +1,97 @@
...
+@Suppress("DEPRECATION")
+@OptIn(InternalComposeUiApi::class, ExperimentalCoroutinesApi::class)
```
jrodbx: which apis require this?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/build.gradle
```diff
@@ -24,8 +24,9 @@ android {
 }
 
 dependencies {
+  implementation libs.composeAnimation.core
+  implementation libs.androidx.lifecycle.runtime
```
jrodbx: nit: would it be possible to craft a test that highlights the problem without having to pull it additional dependencies?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -372,6 +374,7 @@ public class PaparazziSdk @JvmOverloads constructor(
         onNewFrame(scaleImage(frameImage(image)))
       }
     } finally {
+      lifecycleOwner?.registry?.currentState = Lifecycle.State.DESTROYED
```
jrodbx: The PR description specifies what actions were taken in this PR but not the root cause.

Is the issue that the 1st test that runs walks the lifecycle from INITIALIZED to RESUMED, and leaves it there, so when the 2nd test runs, it never transitions (since it's still stuck on RESUMED from the 1st test) thereby causing the bug?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: updated

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: Updated the PR description here, that is correct though, it was getting stuck at `RESUMED` and not being in the expected state for the subsequent run

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/LifecycleCoroutineScopeTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.produceState
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.LocalLifecycleOwner
+import androidx.lifecycle.coroutineScope
+import app.cash.paparazzi.Paparazzi
+import org.junit.Assert.assertTrue
```
jrodbx, line 19:
```
import com.google.common.truth.Truth.assertThat
```

fyi, the project uses Truth for assertions

jrodbx (approved):
nice fix, thanks!

PR #2326, opened 2026-05-05
Support accessibility shrink snapshots

### LINKED ISSUE #1350: `AccessibilityRenderExtension` doesn't work correctly with `RenderingMode.SHRINK`

colinmarsch:
To support SHRINK mode for accessibility snapshots this change introduces a second render pass to re-calculate the size needed for the legend after the view under test has been laid out and we have determined the bounds of it and the semantic information needed for the legend creation. Previously, the combination of `SHRINK` with accessibility snapshots was forbidden and just resulted in a thrown exception.

Fixes https://github.com/cashapp/paparazzi/issues/1350

## Why
`AccessibilityRenderExtension` currently cannot be used with `RenderingMode.SHRINK`, even though developers want to capture accessibility legends for shrink-rendered content.

## What
- Allow accessibility rendering to wrap and measure content in `RenderingMode.SHRINK`
- Rerender shrink accessibility snapshots after legend content is collected
- Add fixture snapshots for standard, wider-than-content, and taller-than-content legends


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -265,13 +265,13 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+    }
     val modifiedView = renderExtensions.fold(view) { currentView, renderExtension ->
-      val currentSessionRenderingMode = sessionParamsBuilder.build().renderingMode
-      if (currentSessionRenderingMode == RenderingMode.SHRINK && renderExtension is AccessibilityRenderExtension) {
-        throw IllegalStateException(
-          "AccessibilityRenderExtension cannot be used with the SHRINK rendering mode. " +
-            "See https://github.com/cashapp/paparazzi/issues/1350 for more context."
-        )
+      if (renderExtension is AccessibilityRenderExtension) {
+        renderExtension.renderView(currentView, currentSessionRenderingMode)
```
jrodbx, line 274: Doesn't this defeat the purpose of the RenderExtension abstraction? ☹️ 

I recall us mentioning replacing the accessibility implementation as a reporting artifact (similar to junit reporting) a few times, and simplify this out of the pipeline.  Is it finally time to break ground on that?  I think it would solve several of these pain points around accessibility vs continuing to double down on the current implementation and introducing workarounds like this.

Fwiw, i'd vote yes, given we're on 2.0-alphas, allowing us to make breaking api changes before the bigger release.

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/java/app/cash/paparazzi/plugin/test/AccessibilityRenderingTest.kt
```diff
@@ -213,16 +214,55 @@ class AccessibilityRenderingTest {
...
     paparazzi.snapshot {
       Column(Modifier.background(Color.LightGray)) {
-        Text("SHRINK and AccessibilityRenderExtension are not supported together")
+        Text("SHRINK and AccessibilityRenderExtension are supported together")
       }
     }
   }
 
+  @Test
+  fun renderingModeSHRINKIncludesLegendWiderThanContent() {
```
saket, line 228: these test names will really benefit from spaces and backticks.

saket (approved):

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: Yeah I do agree this is hacky and shouldn't be here. To get a fix in here for the current structure another option would be to introduce the `renderView(currentView, currentSessionRenderingMode)` on the `RenderExtension` interface itself?

Going forward I do agree that moving the accessibility implementation here out of this rendering pipeline would be ideal and maybe this is the time to do this. I do have an open stack of PRs starting here from earlier to implement https://github.com/cashapp/paparazzi/issues/2239. In it's current state that still relies on the render extension system but it could be refactored further to be standalone.

jrodbx: Putting renderingMode on the RenderExtension interface would remove the is check, but it bakes a concern only accessibility cares about into every implementer — and we'd still need the SHRINK-aware second pass living in PaparazziSdk. Feels like we'd be locking the leak into the public API.

Let's use this moment to do the refactor since alphas are the right window for that breaking change, and it dissolves SHRINK, legend sizing, and the validateAccessibility situation in one go rather than one workaround at a time.  Happy to pair on the API shape.

colinmarsch:
closing this in favour of a larger refactor of the shape of the accessibility snapshot handling per https://github.com/cashapp/paparazzi/pull/2326#discussion_r3228407049

PR #2329, opened 2026-05-13
Improvments for 2205

oldergod:
- JSON are uploaded but are missing the external AAR it points to.
- `build/paparazzi/failures` is shared between different variants + not cleaned before verify runs


colinmarsch (approved):
Looks good to me!

PR #2332, opened 2026-05-14
Follow-up KMP support, drop internal AGP types

jrodbx:
- Replace finalizeDsl + mutable-var targetSdk with a Project.targetSdk(variant) Provider mirroring AGP's HostTestImpl.targetSdkVersion algorithm.
- Replace provideOnEnabled with a Provider<T>.presentWhen(condition) extension built on Provider.filter + flatMap.
- Use (variant as? HasUnitTest)?.unitTest in place of the when(variant){is HasHostTests; is HasUnitTest} switch.
- Reshape the KMP precheck
- Simplify snapshotDir(testVariant: UnitTest) to read .static source dirs and take the first parent as the source-set root. Stable across src/test, src/androidUnitTest, src/androidHostTest.
- Drop internal-AGP imports no longer needed (AnalyticsEnabledComponent, Component, HasHostTests, HostTestBuilder, TestComponent, TestCreationConfig).

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -21,21 +21,16 @@ import app.cash.paparazzi.gradle.reporting.PaparazziTestReporter
 import app.cash.paparazzi.gradle.utils.artifactViewFor
 import app.cash.paparazzi.gradle.utils.capitalize
 import app.cash.paparazzi.gradle.utils.relativize
-import com.android.build.api.component.analytics.AnalyticsEnabledComponent
 import com.android.build.api.dsl.CommonExtension
 import com.android.build.api.instrumentation.FramesComputationMode
 import com.android.build.api.instrumentation.InstrumentationScope
 import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.api.variant.ApplicationAndroidComponentsExtension
-import com.android.build.api.variant.Component
 import com.android.build.api.variant.DynamicFeatureAndroidComponentsExtension
-import com.android.build.api.variant.HasHostTests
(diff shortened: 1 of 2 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -418,38 +401,19 @@ public class PaparazziPlugin @Inject constructor(
...
 
-  private fun Project.provideOnEnabled(filterProvider: Provider<Boolean>, sourceProvider: Provider<*>): Provider<*> {
-    val emptyDir = objects.directoryProperty()
-    return filterProvider.flatMap { enabled ->
-      if (enabled) sourceProvider else emptyDir
-    }
-  }
+  @Suppress("UnstableApiUsage")
+  private fun <T : Any> Provider<T>.presentWhen(condition: Provider<Boolean>): Provider<T> =
+    condition.filter { it }.flatMap { this }
```
colinmarsch, line 406: Should we just keep the `flatMap` approach from before here to avoid adding in the `UnstableApiUsage`?

colinmarsch (approved):
One comment, otherwise this looks good to me!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: Attempted, but `Provider.filter` is the cleanest way to express "make this provider absent when the condition is false"; the alternative is to expose and hoist an empty DirectoryProperty

```
  private fun <T> Provider<T>.presentWhen(
    condition: Provider<Boolean>,
    fallback: Provider<T>
  ): Provider<T> =
    condition.flatMap { enabled -> if (enabled) this else fallback }
```

to the call sites, which makes them noisier and re-introduces the Project dependency.

PR #2205, opened 2025-12-27
Gradle Task Cleanup

geoff-powell:
Add missing task inputs and better support using provider for gradleProps

Fixes: #1790

PR #2334, opened 2026-05-19
Trim redundant inputs, inline failure-dir cleanup

jrodbx:
- Inline failure-dir cleanup into the existing `test.doFirst`; remove the dedicated `deletePaparazzi{Variant}Failures` task.
- Replace `inputs.files(layoutlibNativeRuntime)` + `inputs.files(layoutlibResources)` with `inputs.property("paparazzi.layoutlib.version", NATIVE_LIB_VERSION)`. Content is version-determined; avoids fingerprinting multi-MB binaries on every up-to-date check.
- Drop redundant `inputs.property("paparazzi.test.overwriteOnMaxPercentDifference", ...)` — already covered by `paparazziGradlePropertiesProvider`.
- `SnapshotVerifier`: require `paparazzi.failures.dir` with a clear error; drop the `paparazzi.build.dir` fallback that NPE'd on null anyway.
- Comment the two-layer (source-dir / metadata-file) input design so future readers don't try to consolidate it away.
- Tests: tighten `verifyDeletesFailures` (assert task outcome, drop the misleading `.png` filename); add `recordPreservesFailures` to pin the verify-only scoping.

Refs: #2205, #2329

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -229,21 +229,11 @@ public class PaparazziPlugin @Inject constructor(
       val paparazziGradlePropertiesProvider =
         project.providers.gradlePropertiesPrefixedBy("app.cash.paparazzi")
       val failureDir = buildDirectory.dir("paparazzi/failures/${variant.name}")
-      val deleteFailuresTaskProvider = project.tasks.register(
-        "deletePaparazzi${variantSlug}Failures",
-        Delete::class.java
-      ) {
-        it.group = VERIFICATION_GROUP
-        it.description = "Delete failed screenshot comparison images for variant '${variant.name}'"
-        it.delete(failureDir)
-        it.onlyIf { isVerifyRun.get() }
-      }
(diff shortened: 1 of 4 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
colinmarsch, line 293: Just wondering if this `failureDir` output should only be `presentWhen(isVerifyRun)` now to make sure we align with the intention of the `recordPreservesFailures` test case?

Is there a case where this failureDir is cleaned during a record if the underlying `Test` task re-writes the contents of it?

colinmarsch (approved):
One non-blocking comment, otherwise looks good!

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
jrodbx: oops, enabling the repo setting to "not require being on latest" allowed the auto-merge of this PR to land, before addressing this comment.  sorry about that!

good callout; though, i think that's where the `optional()` would come in, so I _think_ it's a non-issue.

Issue #1968, opened 2025-05-24
`ClassNotFoundException` when upgrading `androidx.navigation` from `2.8.9` to `2.9.0`


yogurtearl:
**Description**

When upgrading `androidx.navigation` from `2.8.9` to `2.9.0` I get this error. 

It looks like `androidx.navigation` removed a transitive dep on `androidx.navigation:navigation-common-ktx` in the new version. 

 `navigation-common-ktx-2.9.0.aar` is essentially empty with no classes (no R class) 

Not sure why paparazzi is still trying to load this class that no longer exists. 

```
    java.lang.ClassNotFoundException: androidx.navigation.common.ktx.R
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:421)
        at java.base/java.lang.Class.forName(Class.java:412)
        at app.cash.paparazzi.internal.PaparazziCallback.initResources(PaparazziCallback.kt:59)
        at app.cash.paparazzi.PaparazziSdk.prepare(PaparazziSdk.kt:124)
        at app.cash.paparazzi.Paparazzi.prepare(Paparazzi.kt:89)
        at app.cash.paparazzi.Paparazzi$apply$1.evaluate(Paparazzi.kt:77)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
```

**Expected behavior**

Don't try to load `R` classes that don't exist. 

**Additional information:**
- Paparazzi Version: 1.3.5

yogurtearl:
[...]

PR #2343, opened 2026-05-20
Handle missing R classes from external artifacts

### LINKED ISSUE #2286: ClassNotFoundException for AndroidX "anchor" module R classes in initResources
## Description

After updating AndroidX dependencies (DataStore 1.1.x+, Lifecycle 2.10.x+), Paparazzi tests crash with `ClassNotFoundException` because `PaparazziCallback.initResources()` tries to load R classes for AndroidX **anchor modules** that don't have any.

Anchor modules (e.g. `datastore-preferences-android`, `lifecycle-viewmodel-android`) are empty marker artifacts used by AndroidX for dependency resolution. They declare a package name (e.g. `androidx.datastore.datastore.preferences.anchor`) which ends up in `resources.json` via `PrepareResourcesTask`, but they don't generate an R class.

## Stack trace

`java.lang.ClassNotFoundException: androidx.datastore.datastore.preferences.anchor.R at java.base/java.lang.Class.forNameClass.java:377 at app.cash.paparazzi.internal.PaparazziCallback.initResourcesPaparazziCallback.kt:59 at app.cash.paparazzi.PaparazziSdk.preparePaparazziSdk.kt:122 at app.cash.paparazzi.Paparazzi.preparePaparazzi.kt:86 at app.cash.paparazzi.Paparazziapply1.evaluatePaparazzi.kt:74`

## Root cause

In [`PaparazziCallback.initResources()`](https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt), `Class.forName("$rPackageName.R")` is called without a try-catch for `ClassNotFoundException`:

```kotlin
fun initResources() {
    for (rPackageName in resourcePackageNames) {
        val rClass = Class.forName("$rPackageName.R") // crashes if R class doesn't exist
        // ...
    }
}
```

The problematic package names in `resources.json` (generated by `PrepareResourcesTask`):
- `androidx.datastore.datastore.preferences.anchor`
- `androidx.lifecycle.lifecycle.viewmodel.anchor`

## Expected behavior

Paparazzi should gracefully skip packages that don't have an R class, instead of crashing.

## Suggested fix

Either:

1. **In `PaparazziCallback.initResources()`** — wrap `Class.forName` in a try-catch and `continue`:
```kotlin
for (rPackageName in resourcePackageNames) {
    val rClass = try {
        Class.forName("$rPackageName.R")
    } catch (_: ClassNotFoundException) {
        continue
    }
    // ...
}
```

2. **In `PrepareResourcesTask`** — filter out package names ending with `.anchor` from `resourcePackageNames` before writing `resources.json`.

## Workaround

We work around this by adding a `doLast` action to the `preparePaparazzi*Resources` task that filters anchor packages from the generated `resources.json`:

```kotlin
tasks.matching { it.name.startsWith("preparePaparazzi") && it.name.endsWith("Resources") }
    .configureEach {
        doLast {
            val resourcesFile = outputs.files.singleFile
            if (!resourcesFile.exists()) return@doLast
            // filter out entries ending with ".anchor" from resourcePackageNames in the JSON
        }
    }
```

## Environment

- **Paparazzi**: 2.0.0-alpha02 (also reproduced by inspecting alpha04 bytecode — same code path)
- **AGP**: 8.13.2
- **Gradle**: 9.2.1
- **AndroidX DataStore**: 1.1.x (transitive via Firebase BOM 34.9.0)
- **AndroidX Lifecycle**: 2.10.0

SimonMarquis:
Catch missing R classes in the same way unknown field types and malformed classes are in the function.

While this is not fixing the root issue brought up by #1968, this will at least prevent crashes in the plugin.

The issue (confirmed? because in our case it was room-ktx and it was not "dropped" as transitive dependency) seems un-fixable right now since we rely on [`org.gradle.api.artifacts.ArtifactCollection#getArtifactFiles(): FileCollection`](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-artifact-collection/index.html#-812691026%2FFunctions%2F-1793262594).

Although, there is a [`org.gradle.api.artifacts.ArtifactCollection#getResolvedArtifacts(): Provider<Set<ResolvedArtifactResult>>`](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.artifacts/-artifact-collection/index.html#606612266%2FFunctions%2F-1793262594) that could allow us to access the resolved artifacts. But it will require changing some exposed types.

Let me know what you think about this.

Closes #2286

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/PaparazziCallback.kt
@@ -53,10 +53,14 @@ internal class PaparazziCallback(
 
   private val loadedClasses = mutableMapOf<String, Class<*>>()
 
-  @Throws(ClassNotFoundException::class)
   fun initResources() {
     for (rPackageName in resourcePackageNames) {
-      val rClass = Class.forName("$rPackageName.R")
+      val rClass = try {
+        Class.forName("$rPackageName.R")
+      } catch (e: ClassNotFoundException) {
+        logger.error(e, "R class not found %1\$s", "$rPackageName.R")
+        continue
+      }
       for (resourceClass in rClass.declaredClasses) {
         val resourceType = ResourceType.fromClassName(resourceClass.simpleName) ?: continue
 


jrodbx (commented):
Do we know what the root cause is?  Swallowing the exception isn't guaranteed to make the plugin behave properly, whereas failing fast is a good signal

SimonMarquis:
It seems like it comes from a missing dependency in the `debugUnitTestRuntimeClasspath` configuration, but I'm unsure why the `-ktx` was not included.

image



Forcing it with testRuntimeOnly makes the issue go away.

```kotlin
testRuntimeOnly(libs.androidx.room.ktx)
```

SimonMarquis:
I've created a basic reproducer on the `:sample` module with a build scan:
- https://github.com/SimonMarquis/paparazzi/pull/3

PR #2317, opened 2026-04-10
Upgrade layoutlib

tcmulcahy:
Upgrade layoutlib to 16.2.3 and layoutlib-api to 32.1.0. Fix breaking changes.

Here is the commit that introduced the 16.2.2 prebuilt: https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/2a1f75b7665fe7510c86b80b92c1e878b8a5f1

It specifies it was built the following commit, which is accessible: https://android.googlesource.com/platform/frameworks/layoutlib/+/e211004e241ca515cb25381adf57f6e792e54646

Here is the commit that introduced the 16.2.3 prebuilt: https://android.googlesource.com/platform/prebuilts/studio/layoutlib/+/0f5c8c2736213742c7b150c9f2fe72c2dee202

It specifies that it was built from the following commit, which is non-accessible:
https://android.googlesource.com/platform/frameworks/layoutlib/+/d0eb3c1279abd884bc1cd7f1aaf57f7feb1f7e19

Since the source for 16.2.3 is non-accessible, we have to disassemble to see the breaking changes:

layoutlib 16.2.3 removed the prepareThread/cleanupThread APIs:

```
% diff <(javap -classpath /tmp/layoutlib-16.2.2.jar com.android.layoutlib.bridge.Bridge) <(javap -classpath /tmp/layoutlib-16.2.3.jar com.android.layoutlib.bridge.Bridge)
15,16d14
<   public static synchronized void prepareThread();
<   public static synchronized void cleanupThread();
```

and also removed the getCurrentRootView API:

```
% diff <(javap -classpath /tmp/layoutlib-16.2.2.jar android.view.WindowManagerImpl) <(javap -classpath /tmp/layoutlib-16.2.3.jar android.view.WindowManagerImpl) | grep -A5 -B5 getCurrentRootView
16a23,24
>   public int getDisplayImePolicy(int);
>   public boolean isGlobalKey(int);
19,24c27,51
<   public void setBaseRootView(android.view.ViewGroup);
<   public android.view.ViewGroup getCurrentRootView();
<   public void createOrUpdateDisplayFrames(android.view.InsetsState);
<   public void setupDisplayCutout();
<   public android.view.WindowManager createLocalWindowManager(android.view.Window);
<   static {};
---
```

PR #2361, opened 2026-05-27
Add synthetic window insets Compose fixture

colinmarsch:
## Why
Add upstream coverage for fake system UI (upstreaming test coverage for a error case we are hitting on the 16.2.3 layoutlib bump in Cash App) that dispatches synthetic status, navigation, and IME insets through a Compose-hosted Android view.

Merging this test coverage in ahead of #2317 will help ensure the changes made there don't break insets applied in this way.

## What
- Add a Compose fixture snapshot that renders fake system UI chrome
- Dispatch synthetic window insets and animation progress to the Compose root and hosted Android view
- Add the syntheticWindowInsets golden snapshot


jrodbx (commented):
It appears that the actual signal comes from InsetAwareView reading topInset and bottomInset and positioning text accordingly. That's what would visibly break if #2317 regresses.

On the other hand, the fake chrome (FakeStatusBar, FakeSoftKeyboard, FakeGestureNavigation) doesn't contribute to what's being tested. It's drawn on top after the insets are built from hardcoded dp constants, so it has no effect on inset propagation.  

If that's accurate, can we drop the chrome and keep just the dispatch + InsetAwareView? Failure mode becomes more legible (text shifts vertically = insets broke) and the fixture shrinks substantially.

colinmarsch:
> It appears that the actual signal comes from InsetAwareView reading topInset and bottomInset and positioning text accordingly. That's what would visibly break if #2317 regresses.
> 
> On the other hand, the fake chrome (FakeStatusBar, FakeSoftKeyboard, FakeGestureNavigation) doesn't contribute to what's being tested. It's drawn on top after the insets are built from hardcoded dp constants, so it has no effect on inset propagation.
> 
> If that's accurate, can we drop the chrome and keep just the dispatch + InsetAwareView? Failure mode becomes more legible (text shifts vertically = insets broke) and the fixture shrinks substantially.

Good call here, I simplified this PR to the real cause.

jrodbx (approved):

PR #2362, opened 2026-06-01
Use unit test variant's runtimeConfiguration instead of main variant's

### LINKED ISSUE #1968: `ClassNotFoundException` when upgrading `androidx.navigation` from `2.8.9` to `2.9.0`

### LINKED ISSUE #2286: ClassNotFoundException for AndroidX "anchor" module R classes in initResources

SimonMarquis:
This ensures only the resolved runtime **unit test** dependencies are used.

image

Fixes #1968
Fixes #2286

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/AndroidVariantSources.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/AndroidVariantSources.kt
@@ -1,8 +1,10 @@
 package app.cash.paparazzi.gradle
 
 import app.cash.paparazzi.gradle.utils.artifactsFor
+import com.android.build.api.variant.UnitTest
 import com.android.build.api.variant.Variant
 import com.android.build.gradle.internal.publishing.AndroidArtifacts
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.file.Directory
 import org.gradle.api.file.FileCollection
@@ -12,7 +14,8 @@ import org.gradle.api.provider.Provider
  * All the relevant sources for a given Android variant.
  */
 internal class AndroidVariantSources(
-  private val variant: Variant
+  private val variant: Variant,
+  private val unitTest: UnitTest
 ) {
   val localResourceDirs: Provider<List<Directory>>? by lazy {
     variant.sources.res?.all?.map { layers -> layers.flatten() }?.map { it.asReversed() }
@@ -20,13 +23,13 @@ internal class AndroidVariantSources(
 
   // https://android.googlesource.com/platform/tools/base/+/96015063acd3455a76cdf1cc71b23b0828c0907f/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/MergeResources.kt#875
   val moduleResourceDirs: FileCollection by lazy {
-    variant.runtimeConfiguration
+    unitTest.runtimeConfiguration
       .artifactsFor(AndroidArtifacts.ArtifactType.ANDROID_RES.type) { it is ProjectComponentIdentifier }
       .artifactFiles
   }
 
   val aarExplodedDirs: FileCollection by lazy {
-    variant.runtimeConfiguration
+    unitTest.runtimeConfiguration
       .artifactsFor(AndroidArtifacts.ArtifactType.ANDROID_RES.type) { it !is ProjectComponentIdentifier }
       .artifactFiles
   }
@@ -37,19 +40,19 @@ internal class AndroidVariantSources(
 
   // https://android.googlesource.com/platform/tools/base/+/96015063acd3455a76cdf1cc71b23b0828c0907f/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/MergeResources.kt#875
   val moduleAssetDirs: FileCollection by lazy {
-    variant.runtimeConfiguration
+    unitTest.runtimeConfiguration
       .artifactsFor(AndroidArtifacts.ArtifactType.ASSETS.type) { it is ProjectComponentIdentifier }
       .artifactFiles
   }
 
   val aarAssetDirs: FileCollection by lazy {
-    variant.runtimeConfiguration
+    unitTest.runtimeConfiguration
       .artifactsFor(AndroidArtifacts.ArtifactType.ASSETS.type) { it !is ProjectComponentIdentifier }
       .artifactFiles
   }
 
   val packageAwareArtifactFiles: FileCollection by lazy {
-    variant.runtimeConfiguration
+    unitTest.runtimeConfiguration
       .artifactsFor(AndroidArtifacts.ArtifactType.SYMBOL_LIST_WITH_PACKAGE_NAME.type)
       .artifactFiles
   }
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -946,6 +946,7 @@ class PaparazziPluginTest {
     assertThat(config.mainPackage).isEqualTo("app.cash.paparazzi.plugin.test")
     assertThat(config.resourcePackageNames).containsExactly(
       "app.cash.paparazzi.plugin.test",
+      "app.cash.paparazzi.plugin.test",
       "com.example.mylibrary",
       "app.cash.paparazzi.plugin.test.module1",
       "app.cash.paparazzi.plugin.test.module2"
@@ -956,6 +957,7 @@ class PaparazziPluginTest {
       "build/generated/res/extra"
     )
     assertThat(config.moduleResourceDirs).containsExactly(
+      "build/intermediates/packaged_res/debug/packageDebugResources",
       "../module1/build/intermediates/packaged_res/debug/packageDebugResources",
       "../module2/build/intermediates/packaged_res/debug/packageDebugResources"
     )
@@ -981,6 +983,7 @@ class PaparazziPluginTest {
     assertThat(config.mainPackage).isEqualTo("app.cash.paparazzi.plugin.test")
     assertThat(config.resourcePackageNames).containsExactly(
       "app.cash.paparazzi.plugin.test",
+      "app.cash.paparazzi.plugin.test",
       "com.example.mylibrary",
       "app.cash.paparazzi.plugin.test.module1",
       "app.cash.paparazzi.plugin.test.module2"
@@ -991,6 +994,7 @@ class PaparazziPluginTest {
       "build/generated/res/extra"
     )
     assertThat(config.moduleResourceDirs).containsExactly(
+      "build/intermediates/packaged_res/debug/packageDebugResources",
       "../module1/build/intermediates/packaged_res/debug/packageDebugResources",
       "../module2/build/intermediates/packaged_res/debug/packageDebugResources"
     )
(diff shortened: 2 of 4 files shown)


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -945,6 +945,7 @@ class PaparazziPluginTest {
     val config = resourcesFile.loadConfig()
     assertThat(config.mainPackage).isEqualTo("app.cash.paparazzi.plugin.test")
     assertThat(config.resourcePackageNames).containsExactly(
+      "app.cash.paparazzi.plugin.test",
       "app.cash.paparazzi.plugin.test",
```
SimonMarquis: Should we use a `buildSet {...}` to avoid duplicated values that will result in the same corresponding `R` class?

https://github.com/cashapp/paparazzi/blob/c640cbbdddae046dcc7868c729867630f491834e/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PrepareResourcesTask.kt#L73-L78

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -956,6 +957,7 @@ class PaparazziPluginTest {
       "build/generated/res/extra"
     )
     assertThat(config.moduleResourceDirs).containsExactly(
+      "build/intermediates/packaged_res/debug/packageDebugResources",
```
SimonMarquis, line 959: Any comments about these additions in the `moduleResourceDirs` & `projectAssetDirs` that are now also linking the consumer project dirs?

SimonMarquis: Is this because the 'main' variant is now seen as a 'dependency'?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: Deduping seems reasonable here to me. Maybe we can use `distinct` or `distinctBy` on the `buildList` to keep the type the same for less change here?

geoff-powell (approved):

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/AndroidVariantSources.kt
```diff
@@ -1,8 +1,10 @@
 package app.cash.paparazzi.gradle
 
 import app.cash.paparazzi.gradle.utils.artifactsFor
+import com.android.build.api.variant.UnitTest
 import com.android.build.api.variant.Variant
 import com.android.build.gradle.internal.publishing.AndroidArtifacts
+import org.gradle.api.artifacts.Configuration
```
colinmarsch: nit: this is unused

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch: > Is this because the 'main' variant is now seen as a 'dependency'?

Yeah, the unit test runtime configuration includes the main component. Could we filter out the current component in this change to avoid this duplication?

SimonMarquis: Done in https://github.com/cashapp/paparazzi/pull/2362/commits/246127648b6d7fc566ae06dd44fc2298858bba4b

SimonMarquis:
@colinmarsch the review comments have been addressed, PTAL 🙏

paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/AndroidVariantSources.kt
```diff
@@ -12,21 +13,25 @@ import org.gradle.api.provider.Provider
...
     variant.sources.res?.all?.map { layers -> layers.flatten() }?.map { it.asReversed() }
   }
 
   // https://android.googlesource.com/platform/tools/base/+/96015063acd3455a76cdf1cc71b23b0828c0907f/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/MergeResources.kt#875
   val moduleResourceDirs: FileCollection by lazy {
-    variant.runtimeConfiguration
-      .artifactsFor(AndroidArtifacts.ArtifactType.ANDROID_RES.type) { it is ProjectComponentIdentifier }
+    unitTest.runtimeConfiguration
+      .artifactsFor(AndroidArtifacts.ArtifactType.ANDROID_RES.type) {
+        it is ProjectComponentIdentifier && it.projectPath != projectPath
```
colinmarsch: Could we distinguish the containing build as well as the `projectPath` here? `projectPath` is only unique within a build, so this could incorrectly exclude an included-build dependency whose path matches the consumer project, for example if both are `:library`. `buildTreePath` would address this on Gradle 8.3+, though we may need a compatibility-safe equivalent for older supported Gradle versions. Same with the two places below.

SimonMarquis: Comparing `Project.path` and `Gradle.buildPath` seems to be good enough.
Let me know if using a `Project` reference like this is banned, but as far as i check, values are not ending up in the configuration cache 👍

SimonMarquis: Actually, forcing these properties in the unit tests, brings up CC errors.
```
org.gradle.configuration-cache=true
org.gradle.configuration-cache.problems=fail
```



I fixed them in https://github.com/cashapp/paparazzi/pull/2362/commits/f46734d4b00299050a1e85a7f50de8965cf64d58, but at the same time I'm wondering why we have all these `by lazy`? Removing them would allow us to not expose `Variant`, `UnitTest`, `Project` as vals of `AndroidVariantSources`, making it CC-leak safe.
Also, I feel like `ArtifactCollection` -> `FileCollection` are somewhat lazy APIs, but I might be wrong.

colinmarsch:
Could we add a regression test here for the #2286 case? I think the empty anchor AAR issue isn't actually resolved by this change.

SimonMarquis:
@colinmarsch I can, but I'm not able to reproduce the issue on master with this test `build.gradle` file inside `paparazzi-gradle-plugin`:

```
plugins {
  id 'com.android.library'
  id 'app.cash.paparazzi'
}

android {
  namespace = 'app.cash.paparazzi.plugin.test'
  compileSdk = libs.versions.compileSdk.get() as int
  defaultConfig {
    minSdk = libs.versions.minSdk.get() as int
  }
  compileOptions {
    sourceCompatibility = libs.versions.javaTarget.get()
  }
}

dependencies {
  implementation 'androidx.lifecycle:lifecycle-viewmodel-android:2.10.0'
  implementation 'androidx.datastore:datastore-preferences-android:1.1.7'
  implementation 'com.squareup.okhttp3:okhttp:5.1.0'
}
```

I tested the 3 mentioned libraries (although there was no complete artifact coordinate provided) and the tests just works.

PR #2364, opened 2026-06-02
Add Compose anchored draggable snapshot fixture

colinmarsch:
## Why
Add regression coverage for Compose state that depends on anchors assigned during layout. This catches cases where Paparazzi captures before an `AnchoredDraggableState` has settled to its initial anchored offset. The 16.2.3 layoutlib bump PR is showing a failure here where the toggle is in the wrong position in the generated snapshot due to compose timing changes in that PR, adding this test to avoid regressions.

## What
- Add a small Compose fixture using `AnchoredDraggableState(initialValue = 1)`
- Assign draggable anchors from `onSizeChanged`
- Record a golden that verifies the right-side anchored position under `RenderingMode.SHRINK`


saket (approved):

jrodbx (commented):
nice

PR #2367, opened 2026-06-09
Rewrite syntheticWindowInsets fixture as pure Compose

colinmarsch:
## Why

The existing `syntheticWindowInsets` fixture is View-based: a custom `View` reacts to dispatched insets in `onDraw`. That shape doesn't exercise the path real Compose apps use — dispatching synthetic insets at the root and consuming them via `WindowInsets` padding inside a Compose hierarchy — so it failed to catch a regression in the layoutlib 16.2.3 upgrade (#2317).

Under layoutlib 16.2.3, the render hierarchy is hosted inside a real window decor (`DecorView` → `LinearLayout` → content frame). When a test dispatches synthetic window insets against the root view, the decor consumes them: the status bar inset becomes top padding and the navigation bar inset shrinks the decor's height, so the content frame shrinks mid-snapshot (1080×2340 → 1080×2037 on Pixel 5 with 62dp/48dp bars). Old layoutlib only updated insets state without resizing anything. Downstream consumers hit this as snapshot dimensions changing during tests that fake system UI via dispatched insets.

Rendering mode matters here: `SHRINK` sizes the canvas from content measurement and bypasses the decor's insets consumption entirely, which hides the bug — so this fixture deliberately runs in `NORMAL` mode, matching how downstream insets-dependent screens are snapshotted.

## What

- Replace the View-based `InsetAwareView` fixture with small Compose pieces: a `SyntheticSystemBarInsets` helper that dispatches status/navigation bar insets to the root view, an `InsetAwareScreen` that consumes the navigation bar insets via `WindowInsets.navigationBars.asPaddingValues()`, and `StatusBar`/`NavigationBar` chrome composables drawn inline.
- Run on `DeviceConfig.PIXEL_5` in `NORMAL` rendering mode.
- Update the golden accordingly.

Passes against master; fails against #2317 (images differ by 14.26%) until the insets/decor behavior there is resolved.

## Before / after under #2317

| Expected (master) | Actual (#2317 merged) |
| --- | --- |
|  |  |

On master, the status bar chrome sits at the top of the window and the bottom content ("Bottom content must remain above nav") is padded above the navigation bar at the bottom edge of the device. With #2317 merged, the decor consumes the dispatched insets: everything shifts down by the status bar inset, the content frame shrinks by the navigation bar inset, so the nav bar floats above the device's bottom edge and the bottom content disappears beneath it.

Paparazzi's delta output from the failing verify run (expected | diff | actual):


paparazzi-gradle-plugin/src/test/projects/compose/src/test/snapshots/images/app.cash.paparazzi.plugin.test_ComposeTest_syntheticWindowInsets.png
geoff-powell, line 1: nit: Why is this not the full size of the device window?

geoff-powell:
This LGTM but can we have a better visualization of the issue before and after?

colinmarsch:
> This LGTM but can we have a better visualization of the issue before and after?

I updated the PR description to include the snapshot diffs that I'm seeing as failures for this test case on PR #2317

paparazzi-gradle-plugin/src/test/projects/compose/src/test/snapshots/images/app.cash.paparazzi.plugin.test_ComposeTest_syntheticWindowInsets.png
colinmarsch: This is the full size of the device window but I changed the device used here to a pixel 5 to more closely match the issue we are seeing

saket (approved):

PR #2368, opened 2026-06-09
Add popup-window system bars Compose fixture

colinmarsch:
## Why

The `syntheticWindowInsets` fixture covers system bar chrome drawn inline in the host window, but another common pattern is hosting chrome in dedicated windows added via `WindowManager.addView`, sized to cover the host window. That path goes through layoutlib's window-manager implementation — which changed substantially in the 16.2.3 upgrade (#2317) and has no existing insets fixture coverage. Under #2317 it fails differently than the inline fixture: the popup window gets offset by the inset-shrunken content frame, so the bars render mispositioned (status bar pushed below the title, nav bar pushed off-screen) rather than the content merely compressing.

## What

- Add a `syntheticWindowInsetsWithPopupOverlay` snapshot test with screen content identical to `syntheticWindowInsets`, so the two goldens isolate the windowing difference. The chrome is hosted in a stock Compose `Popup`, which goes through `WindowManager.addView` via `PopupLayout`.
- Add the expected golden (pixel-identical to the inline fixture's golden on current layoutlib — by design).

Both bars share a single `Popup` rather than one each: when a composition hosts two popup windows, layoutlib drops the first window from the render and throws an NPE in `WindowManagerImpl.removeView` during teardown.

Passes against master; fails against #2317 (images differ by 13.49%) with a distinct failure signature from the inline fixture.


geoff-powell (approved):

paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeTest.kt
```diff
@@ -222,6 +260,57 @@ class ComposeTest {
     }
   }
 
+  /**
+   * Hoists [content] into a separate window added via WindowManager.addView, sized to match the
+   * host view on every layout pass — the pattern commonly used to render system bar chrome or
+   * other overlays above app content in a dedicated window.
+   */
+  @Composable
+  private fun OverlayWindow(content: @Composable () -> Unit) {
```
saket: Can you check whether we can use `Popup()` or `Dialog()` instead? We hand-wrote this overlay for Cash App to render pixel perfect system bars, which may not be necessary here.

saket (approved):

paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeTest.kt
colinmarsch: Updated this to just use a `Popup`, this was a good idea to generalize this test case more!

PR #2371, opened 2026-06-15
Honour FLAG_DIM_BEHIND

### LINKED ISSUE #2370: Paparazzi Compose doesn't honour `FLAG_DIM_BEHIND` for `Dialog`
## Problem

A Compose `Dialog` can clear the platform dialog window's `FLAG_DIM_BEHIND` flag from inside the
dialog content:

```kotlin
val dialogWindow = (LocalView.current.parent as? DialogWindowProvider)?.window
DisposableEffect(dialogWindow) {
  dialogWindow?.clearFlags(FLAG_DIM_BEHIND)
  onDispose {}
}
```

On Android, this should keep the dialog as a real window while removing the dimmed background behind
it. In Paparazzi, the rendered image still appears to include the dialog dim layer after the flag is
cleared.

## Repro

```kotlin
package app.cash.paparazzi.sample

import android.view.WindowManager.LayoutParams.FLAG_DIM_BEHIND
import androidx.compose.foundation.background
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.layout.size
import androidx.compose.runtime.DisposableEffect
import androidx.compose.ui.platform.LocalView
import androidx.compose.ui.unit.dp
import androidx.compose.ui.window.Dialog
import androidx.compose.ui.window.DialogWindowProvider
import app.cash.paparazzi.DeviceConfig
import app.cash.paparazzi.Snapshot
import app.cash.paparazzi.SnapshotHandler
import com.android.ide.common.rendering.api.SessionParams.RenderingMode.NORMAL
import com.google.common.truth.Truth.assertThat
import java.awt.image.BufferedImage
import org.junit.Rule
import org.junit.Test

class ComposeDialogDimFlagTest {
  @get:Rule
  val paparazzi = Paparazzi(
    deviceConfig = DeviceConfig.PIXEL_5.copy(softButtons = false),
    renderingMode = NORMAL,
    snapshotHandler = PixelAssertingSnapshotHandler { image ->
      val topLeftPixel = java.awt.Color(image.getRGB(8, 8), true)

      assertThat(topLeftPixel.red).isEqualTo(255)
      assertThat(topLeftPixel.green).isEqualTo(255)
      assertThat(topLeftPixel.blue).isEqualTo(255)
    },
  )

  @Test
  fun clearingFlagDimBehindRemovesDialogDimLayer() {
    paparazzi.snapshot {
      DialogWithoutDimRepro()
    }
  }
}

@Composable
private fun DialogWithoutDimRepro() {
  Box(Modifier.fillMaxSize().background(Color.White)) {
    Dialog(onDismissRequest = {}) {
      val dialogWindow = (LocalView.current.parent as? DialogWindowProvider)?.window
      DisposableEffect(dialogWindow) {
        dialogWindow?.clearFlags(FLAG_DIM_BEHIND)
        onDispose {}
      }

      Box(Modifier.size(120.dp).background(Color.Red))
    }
  }
}

private class PixelAssertingSnapshotHandler(
  private val assertImage: (BufferedImage) -> Unit,
) : SnapshotHandler {
  override fun newFrameHandler(
    snapshot: Snapshot,
    frameCount: Int,
    fps: Int,
    return object : SnapshotHandler.FrameHandler {
        assertImage(image)
      }

      override fun close() = Unit
    }
  }

  override fun close() = Unit
}
```

Expected result:

- The test passes. The root composable is white, and clearing `FLAG_DIM_BEHIND` should leave pixels
  outside the red dialog content white.

Actual result:

- The test fails because the sampled background pixel is dimmed instead of white.

## Notes

Paparazzi wraps composable snapshots in a `ComposeView`, adds that host view to the layoutlib root,
renders a `RenderSessionImpl`, and captures `bridgeRenderSession.image`. The likely issue is in the
separate window/overlay path used by Compose `Dialog`: the dim layer appears to be retained in the
final layoutlib image even after the dialog window flag is cleared from the dialog content.

oldergod:
Fixes #2370

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
@@ -19,6 +19,8 @@ import android.animation.AnimationHandler
 import android.content.Context
 import android.content.res.Resources
 import android.graphics.Bitmap
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
 import android.os.Handler
 import android.os.Handler_Delegate
 import android.util.AttributeSet
@@ -32,6 +34,10 @@ import android.view.View
 import android.view.View.NO_ID
 import android.view.ViewGroup
 import android.view.ViewGroup.LayoutParams
+import android.view.Window
+import android.view.WindowManager
+import android.view.WindowManager.LayoutParams.FLAG_DIM_BEHIND
+import android.view.WindowManagerImpl
 import androidx.activity.setViewTreeOnBackPressedDispatcherOwner
 import androidx.annotation.LayoutRes
 import androidx.compose.runtime.Composable
@@ -335,11 +341,7 @@ public class PaparazziSdk @JvmOverloads constructor(
 
         // If we have pendingTasks run recomposer to ensure we get the correct frame.
         var hasPendingWork = false
-        withTime(nowNanos) {
-          val result = renderSession.render(true)
-          if (result.status == ERROR_UNKNOWN) {
-            throw result.exception
-          }
+        renderFrame(nowNanos) {
           if (hasComposeRuntime && recomposer != null) {
             // If we have pending tasks, we need to trigger it within the context of the first frame.
             if (frame == 0 && (recomposer as Recomposer).hasPendingWork) {
@@ -349,12 +351,7 @@ public class PaparazziSdk @JvmOverloads constructor(
         }
 
         if (hasPendingWork) {
-          withTime(nowNanos) {
-            val result = renderSession.render(true)
-            if (result.status == ERROR_UNKNOWN) {
-              throw result.exception
-            }
-          }
+          renderFrame(nowNanos)
 
           val recomposerInstance = recomposer as Recomposer
           if (recomposerInstance.hasPendingWork) {
@@ -362,6 +359,10 @@ public class PaparazziSdk @JvmOverloads constructor(
               "Pending work detected. This may cause unexpected results in your generated snapshots. ${recomposerInstance.changeCount}"
             )
           }
+        }
+
+        if (syncLayoutlibWindowManagerRootViewBackground()) {
+          renderFrame(nowNanos)
         }
 
         val image = bridgeRenderSession.image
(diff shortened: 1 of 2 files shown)


geoff-powell:
Can you provide a snapshot of what this should look like when the flag is enabled?

oldergod:
Dialog naturally comes with a dimmed background around it like
image

Using the flag prevents this dimming.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -398,6 +399,63 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+    withTime(timeNanos) {
+      val result = renderSession.render(true)
+      if (result.status == ERROR_UNKNOWN) {
+        throw result.exception
+      }
+      block()
+    }
+  }
+
+  private fun syncLayoutlibWindowManagerRootViewBackground(): Boolean {
```
jrodbx, line 412: i feel like this is something that belongs upstream in layoutlib and not in paparazzi.

would need to dig in more, but i also have a question on the underlying issue and commented here: https://github.com/cashapp/paparazzi/issues/2370#issuecomment-4721172883

oldergod: Dug into the layoutlib side. The root cause is upstream, agreed: `WindowManagerImpl.addView` applies the dim once, from the flags at add time (`mCurrentRootView.setBackgroundColor(Color.argb(params.dimAmount, 0, 0, 0))` when `FLAG_DIM_BEHIND` is set), and `updateViewLayout` only re-applies position, elevation, and gravity. Nothing in layoutlib re-reads `FLAG_DIM_BEHIND` or `dimAmount` after the window is added. So a `clearFlags` after the window exists can never take effect there, and the frames in #2370 confirm it never self-corrects.

So the durable fix belongs in layoutlib, and an upstream report is worth filing. But Paparazzi consumes layoutlib as a prebuilt, so an upstream fix only reaches users a release train later. This patch has the same shape as the other layoutlib gaps `PaparazziSdk` already works around: it mirrors `addView`'s own dim formula, it lives in one private function, and it deletes cleanly once layoutlib honors attribute updates. My preference is to ship the shim now and remove it behind an upstream fix. Happy to link an upstream issue from a code comment here if you want that breadcrumb.

PR #2372, opened 2026-06-16
fix: exclude blank text from merged accessibility labels

### LINKED ISSUE #2331: AccessibilityRenderExtension includes empty Compose text in merged accessibility labels
**Description**
`AccessibilityRenderExtension` appears to include blank Compose text nodes when building accessibility overlay labels for merged semantics nodes.

For a merged Compose container with visible text plus an intentionally empty `Text("")` child used only for layout spacing, the accessibility snapshot can render a label like:

```text
$356.21,
```

instead of:

```text
$356.21
```

That makes the snapshot look like there is an accessibility issue even though the empty text does not represent useful user-facing semantics. Downstream consumers can work around this by adding `Modifier.clearAndSetSemantics {}` to the empty text node, but that seems like snapshot-harness cleanup rather than something app code should need for TalkBack behavior.

**Steps to Reproduce**
Use `AccessibilityRenderExtension` against a Compose hierarchy like:

```kotlin
Column(
  modifier = Modifier.semantics(mergeDescendants = true) {},
) {
  Text("$356.21")
  Text("")
}
```

Render it with Paparazzi accessibility snapshots.

**Expected behavior**
The accessibility overlay should ignore blank accessibility text when constructing merged labels, so the rendered label is:

```text
$356.21
```

**Actual behavior**
The blank text contributes an empty entry to the joined label, producing a trailing separator:

```text
$356.21,
```

**Potential cause**
In `AccessibilityRenderExtension`, merged Compose semantics are built by walking unmerged nodes and joining each node's `accessibilityText()` result. Blank text can survive as an empty string and then participate in `joinToString(", ")`.

A possible fix would be to treat blank accessibility text the same as absent accessibility text before joining merged child labels, or have the text construction path return `null` for blank-only text.

**Additional information:**
- Paparazzi Version: observed downstream with `2.0.0-internal15`
- OS: macOS
- Compile SDK: not yet reduced to a standalone sample
- Gradle Version: not yet reduced to a standalone sample
- Android Gradle Plugin Version: not yet reduced to a standalone sample

**Screenshots**
No public screenshot attached yet; the observed downstream snapshot diff was only the accessibility overlay label changing from `$356.21,` to `$356.21`.

geoff-powell:
## Summary

Fixes #2331 — `AccessibilityRenderExtension` was including empty/blank Compose `Text("")` nodes in merged accessibility labels, producing trailing commas (e.g., `"Hello, "` instead of `"Hello"`).

## Changes

**Root cause:** In `AccessibilityElementCollector.kt`, `.mapNotNull { it.accessibilityText() }` didn't filter out blank strings. `Text("")` returns `""` from `accessibilityText()` (not null), so it participated in `joinToString(", ")`.

**Fix:** Added `.takeIf { text -> text.isNotBlank() }` to exclude empty and whitespace-only text from merged labels:

```kotlin
// Before
.mapNotNull { it.accessibilityText() }

// After
.mapNotNull { it.accessibilityText()?.takeIf { text -> text.isNotBlank() } }
```

**Tests added:**
- `verify blank text excluded from merged accessibility labels` — `Text("")` inside a merged semantics container
- `verify whitespace-only text excluded from merged accessibility labels` — `Text("   ")` inside a merged semantics container

## Files changed

- `paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityElementCollector.kt` — one-line fix
- `AccessibilityRenderingTest.kt` — two new test methods + golden snapshots


oldergod (approved):

jrodbx (approved):
Can we dress up the PR commit message + body to be more concise and follow the "conventions" of the repo (one example: no "fix:" prefix)?  AI assistance is great, but I'd like this repo to still read for humans 😉

geoff-powell:
> Can we dress up the PR commit message + body to be more concise and follow the "conventions" of the repo (one example: no "fix:" prefix)? AI assistance is great, but I'd like this repo to still read for humans 😉

Haha yep, sorry. This is a commit message format leak from another project. I can clean it up 👍🏽

PR #2375, opened 2026-06-26
Upgrade LayoutLib to 16.2.3 with frame timing fix

### LINKED ISSUE #2373: Rendering a Compose Dialog crashes on teardown (WindowManagerImpl.removeView NPE) with layoutlib 16.2.1
## Description

Snapshotting any `androidx.compose.ui.window.Dialog` crashes during teardown. The snapshot itself renders, but when Paparazzi moves the lifecycle to `DESTROYED`, Compose's `AndroidDialog` `DisposableEffect.onDispose` calls `Dialog.dismiss()`, and layoutlib's `WindowManagerImpl.removeView` throws an NPE on a null `childrenList`.

## Minimal repro

```kotlin
@RunWith(TestParameterInjector::class)
class DialogTeardownTest {

    @Test
    fun dialog() {
        paparazzi.snapshot {
            Dialog(onDismissRequest = {}) { Text("hello") }
        }
    }
}
```

## Stacktrace

```
java.lang.NullPointerException: Cannot invoke "java.util.ArrayList.size()" because "childrenList" is null
    at android.view.WindowManagerImpl.removeView(WindowManagerImpl.java:209)
    at android.view.WindowManagerImpl.removeViewImmediate(WindowManagerImpl.java:245)
    at android.app.Dialog.dismissDialog(Dialog.java:404)
    at android.app.Dialog.dismiss(Dialog.java:386)
    at androidx.compose.ui.window.AndroidDialog_androidKt$Dialog$1$1$invoke$$inlined$onDispose$1.dispose(Effects.kt:630)
    at androidx.compose.runtime.DisposableEffectImpl.onForgotten(Effects.kt:91)
    at androidx.compose.runtime.CompositionImpl.dispose(Composition.kt:945)
    ...
    at app.cash.paparazzi.PaparazziSdk.takeSnapshots(PaparazziSdk.kt:378)
```

## Environment

- Paparazzi 2.0.0-alpha05
- layoutlib 16.2.1 (`com.android.tools.layoutlib:layoutlib:16.2.1`)
- Compose UI 1.11.2
- Kotlin 2.3.21, AGP 9.2.1, Gradle 9.5.1

geoff-powell:
## Summary

- Upgrades LayoutLib from 16.2.1 to 16.2.3
- Fixes frame timing mechanism in `PaparazziSdk.kt` to work correctly with LayoutLib 16.2.3's new `getNanosTime()` formula
- Updates AccessibilityRenderExtension to use `WindowManagerGlobal` for popup root view discovery
- Regenerates golden images for rendering differences in the new LayoutLib version

## Description

LayoutLib 16.2.3 changed how `SessionInteractiveData.getNanosTime()` computes time, which broke Compose frame callbacks that rely on non-zero frame times (causing "0 isn't a real frame time!" crashes in `AndroidUiFrameClock`).

### Frame Timing Fix (`PaparazziSdk.kt`)
Implements a hybrid choreographer execution approach:
1. **Manual callback execution first** — directly calls `choreographerCallbacks.execute(frameNanos, log)` with a guaranteed non-zero `frameNanos`, ensuring Compose's FrameCallback always receives a valid time
2. **Zeroed `sChoreographerTime` before `doFrame`** — prevents re-posted callbacks from double-firing during doFrame's internal callback execution, while still allowing HWUI native signaling for ripples and view animations

### Accessibility Extension
Updated `AccessibilityRenderExtension` to use `WindowManagerGlobal.findPopupRootView()` instead of `WindowManagerImpl.currentRootView` for more reliable popup window discovery.

### Golden Updates
Regenerated snapshot goldens affected by LayoutLib rendering changes:
- `syntheticWindowInsets.png`
- `syntheticWindowInsetsWithPopupOverlay.png`
- `anchoredDraggableAnchorsFromSize.png`
- Updated `EXPECTED_CURRENT_V_SCROLL_WIDTH` (514 → 562)

## Type of Change
- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [x] Architecture Enhancements


paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeRenderingModeSizingTest.kt
```diff
@@ -28,7 +28,7 @@ import org.junit.Test
 import java.awt.image.BufferedImage
 import kotlin.math.min
 
-private const val EXPECTED_CURRENT_V_SCROLL_WIDTH = 514
```
geoff-powell, line 31: this is the correct size now (was incorrect before)

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -392,30 +400,119 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+
+    try {
+      val window = decorView.javaClass.getFieldReflectively("mWindow").get(decorView)
+      window.javaClass
+        .getMethod("setDecorFitsSystemWindows", Boolean::class.javaPrimitiveType)
+        .invoke(window, false)
+    } catch (ignored: Throwable) {
+      // Older layoutlib versions don't wrap content in a PhoneWindow; nothing to disable.
+    }
+  }
```
geoff-powell: nit: Going to @Ignore the inset tests as they are specific to cash. @colinmarsch you can use this workaround to get insets to correctly delegate.

paparazzi/src/test/java/app/cash/paparazzi/PaparazziTest.kt
```diff
@@ -59,8 +59,6 @@ class PaparazziTest {
 
   @Test
   fun resetsAnimationHandler() {
-    assertThat(AnimationHandler.sAnimatorHandler.get()).isNull()
```
geoff-powell, line 62: AnimationHandler.sAnimatorHandler is non-null due to layout lib configuration that occurs in the `Paparazzi.prepare()` step

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityElementCollector.kt
```diff
@@ -73,21 +73,23 @@ internal class AccessibilityElementCollector {
...
-      locationOnScreen[1] += paddingTop
-      val orderedSemanticsNodes = viewRoot.semanticsOwner.rootSemanticsNode.orderSemanticsNodeGroup()
-      orderedSemanticsNodes.forEach {
-        it.processAccessibleChildren(
-          processElement = processElement,
-          locationOnScreen = locationOnScreen,
-          viewBounds = bounds,
-          unmergedNodes = unmergedNodes
-        )
+      val viewRoot = getChildAt(0) as? ViewRootForTest
```
geoff-powell, line 73: nit: PopupLayout from compose doesn't not implement `ViewRootForTest` interface

geoff-powell: Actually upon initial layout (PopoutLayout) doesn't have any children until a single measure/layout pass.

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/snapshots/images/app.cash.paparazzi.plugin.test_AccessibilityRenderingTest_alertDialogMaterial3.png
colinmarsch, line 1: Ideally this dialog should be only within the left half of the snapshot, but since this is an existing regression I think this is fine for this PR

paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeRenderingModeSizingTest.kt
colinmarsch: Do you know why the size changed here? I'd like to understand what the bug was here if the previous state was wrong.

sample/src/test/snapshots/videos/app.cash.paparazzi.sample_LottieTest_lottie_lottie_logo.png
colinmarsch, line 1: Are there still slight frame timing changes in these apng snapshots? Just curious why they needed to be re-recorded as they look visually identical to me.

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
colinmarsch: This sounds good to me, I'm aligned with this change

paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeRenderingModeSizingTest.kt
geoff-powell: If you looks at other snapshots with the same device. The generated snapshot is`562 x 1000`.
https://github.com/cashapp/paparazzi/pull/2375/changes#diff-53e69d09e34ebb410b2767ebc7aa07cf5dab52be39f62c5c68e43b8d25e5e201

paparazzi-gradle-plugin/src/test/projects/accessibility-rendering/src/test/snapshots/images/app.cash.paparazzi.plugin.test_AccessibilityRenderingTest_alertDialogMaterial3.png
geoff-powell: Yes, this is existing window issue. There was a fix in the other external layoutlib 16.2.3 PR but I was trying to make the minimal amount of changes with the least amount of regressions.

sample/src/test/snapshots/videos/app.cash.paparazzi.sample_LottieTest_lottie_lottie_logo.png
geoff-powell: The frame diff is like a pixel off. Most likely due to the frametime changes.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -50,39 +45,46 @@ public class AccessibilityRenderExtension : RenderExtension {
...
-            FrameLayout.LayoutParams(contentView.measuredWidth, MATCH_PARENT, Gravity.START)
+        val accessibleWindowRoots = WindowManagerGlobal.getInstance().windowViews.reversed().associate {
+          val id = it.hashCode()
+          val coreView = it.findComposeViewAdapterChild()
+          overlayDrawables.getOrPut(id) {
+            AccessibilityOverlayDrawable()
+          }.apply {
+            coreView.foreground = this
+          }
+          id to coreView
```
geoff-powell, line 57: Tracking accessibility information by each window view (instead of relying on a single view to drive all information.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -50,39 +45,46 @@ public class AccessibilityRenderExtension : RenderExtension {
...
     }
   }
 }
 
-private fun View.findRootView(): View {
-  var parent = parent
-  while (parent != null) {
-    if (parent is ComposeViewAdapter) {
-      return parent
+private fun View.findComposeViewAdapterChild(): View {
```
geoff-powell, line 78: finds children instead of a upward lookup

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -392,12 +398,37 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+   *
+   * We set the frame directly rather than calling [RenderSessionImpl.measure] because that would run
+   * an extra traversal/scroll pass whose process-global side effects (shared `Looper`/animation
+   * state) leak into later snapshots on the same thread. No-ops on layoutlib versions that don't
+   * expose `ViewRootImpl_Accessor.updateFrame`.
+   */
+  private fun sizeShrinkWindowFrameToDevice(contentView: View) {
+    val displayMetrics = contentView.context.resources.displayMetrics
+    ViewRootImpl_Accessor.updateFrame(contentView.viewRootImpl, displayMetrics.widthPixels, displayMetrics.heightPixels)
+  }
```
geoff-powell, line 431: AI Explaination

## What the fix does
- The anchoredDraggable fixture (ComposeTest.anchoredDraggableAnchorsFromSize, PR #2364) is a regression test for a real layoutlib 16.2.3 behavior change. In PaparazziSdk.kt:330-337, after viewGroup.addView(modifiedView), the fix calls sizeShrinkWindowFrameToDevice only when renderingMode == SHRINK:
- ViewRootImpl_Accessor.updateFrame(contentView.viewRootImpl, displayMetrics.widthPixels, displayMetrics.heightPixels)
I decompiled android.view.ViewRootImpl_Accessor from layoutlib-16.2.3.jar and confirmed updateFrame(vri, w, h) sets mWinFrame, mTmpFrames.frame, and mTmpFrames.displayFrame all to (0,0,w,h). So it restores the window frame to the full device size before the first render().
## What AI verified is correct
1. Root cause is real. layoutlib 16.2.3's inflate() eagerly measures the (empty) content and collapses the ViewRootImpl frame. Compose's first frame-clock pass reads that stale 0x0 frame, so onSizeChanged-derived state (the anchors) resolves at 0x0 first.
2. The SHRINK-only gate is correct. I decompiled the RenderingMode enum: only SHRINK uses SizeAction.SHRINK in both dimensions. V_SCROLL/H_SCROLL/FULL_EXPAND use EXPAND (grow from device size, never collapse), and NORMAL keeps device size. So no other mode can collapse to 0x0. The scope is right.
3. The API exists in the pinned layoutlib, and the updateFrame signature matches exactly.
4. Placement is correct — after addView, before the frame render loop, and it avoids RenderSessionImpl.measure() (the KDoc correctly explains that would leak process-global Looper/animation side effects into later snapshots).

geoff-powell:
## Frame-timing implementation summary after discovery

<details>
<summary><b>Click to expand — implementation summary</b></summary>

### What changed

**1. Per-frame layoutlib elapsed time (Google standalone-render approach)**

`withTime` now sets the render session's elapsed-frame time **per frame**, matching Google's deviceless layoutlib harness (`RenderTestBase` / `standalone-render`):

```kotlin
Choreographer_Delegate.sChoreographerTime = frameNanos
renderSession.setElapsedFrameTimeNanos(frameNanos)
```

Layoutlib 16.2.3's `RenderSessionImpl#render` divides `mElapsedFrameTimeNanos` by `1_000_000` into `AnimatedVectorDrawable`'s native animator (`sFrameTime`), so native animated-vector timing now advances in lockstep with the Choreographer clock for nonzero snapshot offsets (previously frozen at `0`).

**2. Removed Choreographer private-field reflection**

The old implementation reflected into private `Choreographer` fields (`mCallbacksRunning`, `mLastFrameTimeNanos`) to dispatch exactly one animation callback per frame. That reflection is gone, replaced by the public delegate:

```kotlin
Choreographer_Delegate.doCallbacks(
  Choreographer.getInstance(),
  Choreographer.CALLBACK_ANIMATION,
  currentTimeNanos
)
Choreographer_Delegate.sChoreographerTime = 0
Choreographer_Delegate.doFrame(currentTimeNanos)
```

Verified via disassembly of layoutlib 16.2.3: `Choreographer_Delegate.doCallbacks(choreographer, type, frameTimeNanos)` sets `mCallbacksRunning = true`, runs the aggregate `ChoreographerCallbacks.execute(nanoTime, log)` queue (`CALLBACK_ANIMATION` / `CALLBACK_TRAVERSAL`), then clears the flag — the exact work the reflection did. The `sChoreographerTime = 0` + `doFrame` two-step is retained to signal HWUI while preventing a re-posted callback from double-firing within the same frame.

### Why the reflection was safe to remove

- Layoutlib's `render()` sets `sFrameTime` but does **not** dispatch Choreographer/animation callbacks — Paparazzi must drive them.
- `doCallbacks` is public and guards `mCallbacksRunning` internally, so no private-field access is needed.
- Exactly-once-per-frame dispatch is asserted by `PaparazziTest.animationEvents` (precise per-frame `onAnimationUpdate` elapsed values, no double fire) and `animationCallbacksForStaticSnapshots`.

### Verification

| Check | Result |
|---|---|
| `:paparazzi:test` (full unit suite) | ✅ green |
| `animationEvents`, `animationCallbacksForStaticSnapshots` | ✅ exactly-once dispatch preserved |
| `layoutlibAnimatedVectorFrameTimeUsesSnapshotOffset` (RED→GREEN) | ✅ `expected 300` (was `0`) |
| `:sample:verifyPaparazziDebug` — KeypadView spin, Lottie, AVD, Compose | ✅ byte-identical goldens |
| `spotlessCheck` + `git diff --check` | ✅ clean |

### New regression fixtures (pixel-level proof)

- **`AnimatedVectorDrawableTest.rotatesAcrossFrames`** — an AVD rotating a full turn over 1s at 8fps; the 9-frame filmstrip shows distinct angles (0°→360°), proving native animation is no longer frozen.
- **`ComposeAnimationTest.slidesAcrossFrames`** — a box sliding via `animateDpAsState`, proving the Compose clock path is undisturbed.

### Net production diff (`PaparazziSdk.kt`)

Smaller and reflection-free while preserving pixel-for-pixel output.

</details>

geoff-powell:
## Investigation -- Per-frame Layoutlib elapsed time

<details>
<summary><b>Click to expand — spike document</b></summary>

# Spike 001 — Per-frame Layoutlib elapsed time (Google standalone-render approach)

## Question

Given a Paparazzi snapshot or GIF frame at `offsetMillis`, when Layoutlib renders
that frame, does native animation (`AnimatedVectorDrawable`) receive the same
elapsed time in milliseconds — and do the rendered **pixels** actually advance?

## Why this matters

Google's deviceless Layoutlib screenshot harness (`RenderTestBase`, and the
`standalone-render` library at
`mirror-goog-studio-main:standalone-render/`) drives animation timing by calling
`RenderSession.setElapsedFrameTimeNanos(frameTimeNanos)` before each render. The
Layoutlib 16.2.3 bridge divides that value by 1_000_000 (nanos → millis) and
stores it into `AnimatedVectorDrawable_VectorDrawableAnimatorUI_Delegate.sFrameTime`
(verified by disassembling `RenderSessionImpl`:
`getfield mElapsedFrameTimeNanos; ldc2_w 1000000; ldiv; putstatic sFrameTime`).

Paparazzi already advances the Compose/Choreographer clock for each frame, but it
initialized Layoutlib elapsed-frame time to zero **only once** when constructing
the render session. Native animated-vector timing therefore stayed at zero for
snapshots taken at a nonzero offset — native vector animations were frozen.

## Change (matches Google's per-render pattern)

Set the render session's elapsed-frame time **per frame**, inside `withTime` —
the single place every render pass flows through (primary render, the pending-work
Compose re-render, and the non-Compose path):

```kotlin
private fun withTime(timeNanos: Long, block: () -> Unit) {
  val frameNanos = timeNanos
  System_Delegate.setNanosTime(0L)
  Choreographer_Delegate.sChoreographerTime = frameNanos
  // Matches Google's RenderTestBase: set elapsed-frame time before each render.
  renderSession.setElapsedFrameTimeNanos(frameNanos)
  ...
}
```

`createRenderSession` still seeds `setElapsedFrameTimeNanos(0L)`; per-frame timing
now owns the value thereafter. This is deliberately narrow: it leaves the existing
Compose frame-clock workaround untouched and supplies Layoutlib's public per-frame
animation input, applied uniformly rather than at a single render call (as the
initial spike draft did).

## Follow-up: remove Choreographer private-field reflection

Layoutlib 16.2.3's `RenderSessionImpl#render` sets
`sFrameTime = mElapsedFrameTimeNanos / 1_000_000` but does **not** dispatch the
Choreographer animation callbacks — Paparazzi drives those in `withTime`. The old
implementation reflected into private `Choreographer` fields (`mCallbacksRunning`,
`mLastFrameTimeNanos`) to run exactly one animation dispatch per frame and avoid a
re-posted callback firing twice.

That reflection is unnecessary. `android.view.Choreographer_Delegate` exposes a
**public** `doCallbacks(Choreographer, callbackType, frameTimeNanos)` that already
does precisely this: it sets `mCallbacksRunning = true`, runs the aggregate
`ChoreographerCallbacks.execute(nanoTime, log)` queue (`CALLBACK_ANIMATION` /
`CALLBACK_TRAVERSAL`), then clears the flag. The reflection block was replaced with:

```kotlin
Choreographer_Delegate.doCallbacks(
  Choreographer.getInstance(),
  Choreographer.CALLBACK_ANIMATION,
  currentTimeNanos
)
Choreographer_Delegate.sChoreographerTime = 0
Choreographer_Delegate.doFrame(currentTimeNanos)
```

Same single-dispatch behaviour, no `getDeclaredField`/`isAccessible` reflection.
The exactly-once-per-frame contract is proven by `PaparazziTest.animationEvents`
(asserts precise per-frame `onAnimationUpdate` elapsed values with no double fire)
and `animationCallbacksForStaticSnapshots`.

## Evidence

### RED

With the per-frame call removed, the regression test failed:

```text
PaparazziTest.layoutlibAnimatedVectorFrameTimeUsesSnapshotOffset
expected: 300
but was : 0
```

The test snapshots at `offsetMillis = 300` and asserts Layoutlib's animated
vector frame-time delegate is `300` milliseconds. HWUI logs also showed
`IntendedVsync=0` before the fix vs `IntendedVsync=300000000` after.

### GREEN — delegate value (unit)

`PaparazziTest.layoutlibAnimatedVectorFrameTimeUsesSnapshotOffset` passes.

### GREEN — rendered pixels (integration goldens)

Two sample fixtures capture GIFs (recorded, then re-verified in `verify` mode):

- **`AnimatedVectorDrawableTest.rotatesAcrossFrames`** — an AVD that rotates a
  full turn over 1s, captured at 8fps. The 9-frame filmstrip shows the shape at
  distinct angles per frame (0°, 45°, 90°, … 360°) — native animation is no
  longer frozen. HWUI `IntendedVsync` advanced 125ms per frame.
- **`ComposeAnimationTest.slidesAcrossFrames`** — a box sliding via
  `animateDpAsState`, proving the Compose clock path still advances per frame and
  is undisturbed by the new layoutlib timing.

`:paparazzi:test` (full unit suite), `:sample:verifyPaparazziDebug` for both new
tests, `spotlessCheck`, and `git diff --check` all pass.

## Verdict: VALIDATED

### What worked

- Layoutlib receives the requested elapsed time for a nonzero snapshot offset,
  applied per-frame exactly where Google's harness applies it.
- Native `AnimatedVectorDrawable` pixels genuinely animate across frames.
- The Compose timing path remains green — the change augments, not replaces it.
- The Choreographer private-field reflection was removed in favour of the public
  `Choreographer_Delegate.doCallbacks`, with all view/native/Compose animation
  goldens and the timing unit tests unchanged (byte-identical pixels).

### Recommendation for the real build

Keep the per-frame `setElapsedFrameTimeNanos(frameNanos)` call and the
reflection-free `doCallbacks` dispatch in `withTime`, and keep both golden
fixtures (`AnimatedVectorDrawableTest`, `ComposeAnimationTest`) as regression
coverage for native-animation and Compose-animation timing.

</details>

geoff-powell:
Layoutlib 16.2.3 fixes #2373 as well

paparazzi-gradle-plugin/src/test/projects/verify-animation-gif/src/test/snapshots/videos/app.cash.paparazzi.plugin.test_AnimatedVectorDrawableTest_rotatesAcrossFrames_rotation.png
colinmarsch, line 1: Do we need both of these new test cases? Is the reasoning for both just to cover Compose and xml based animations?

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -50,39 +44,46 @@ public class AccessibilityRenderExtension : RenderExtension {
...
-        val rootView = contentView.findRootView()
-        rootView.foreground = overlayDrawable
-
-        // The root of the view hierarchy is rendered at full width.
-        // We need to restrict it when taking accessibility snapshots.
-        val windowManagerRootView = (windowManager as WindowManagerImpl).currentRootView
-        if (windowManagerRootView != null) {
-          windowManagerRootView.layoutParams =
-            FrameLayout.LayoutParams(contentView.measuredWidth, MATCH_PARENT, Gravity.START)
+        val accessibleWindowRoots = WindowManagerGlobal.getInstance().windowViews.reversed().associate {
```
colinmarsch, line 49: Why is this reversed?

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -398,23 +404,67 @@ public class PaparazziSdk @JvmOverloads constructor(
...
+      // layoutlib 16.2.3's Choreographer#doFrame dispatches the animation callbacks itself, but a
+      // re-posted callback (dueTime = uptimeMillis()) becomes due again within the same frame and
+      // fires a second time. To keep exactly one dispatch per frame (matching pre-16.2.3 behavior),
+      // dispatch the animation callbacks once here via the public Choreographer_Delegate.doCallbacks
+      // (which guards mCallbacksRunning and runs the aggregate ChoreographerCallbacks queue) - the
+      // same work the previous private-field reflection did, without reflection - then tick doFrame
+      // with sChoreographerTime zeroed so its internal dispatch finds the re-posted callbacks
+      // not-yet-due and skips them (while still signalling the native HWUI layer so ripples and view
+      // animations work).
+      Choreographer_Delegate.doCallbacks(
```
colinmarsch, line 459: This is a lot simpler now, this is great!

sample/src/test/java/app/cash/paparazzi/sample/ComposeDialogShrinkTest.kt
```diff
@@ -26,6 +27,10 @@ class ComposeDialogShrinkTest {
     renderingMode = RenderingMode.SHRINK
   )
 
+  @Ignore(
+    "Within previews the dialog is not rendered correctly, so we " +
```
colinmarsch, line 31: Had Paparazzi diverged from this preview behaviour previously? Just trying to understand why we need to ignore this now but it was passing beforehand?

paparazzi-gradle-plugin/src/test/projects/verify-animation-gif/src/test/snapshots/videos/app.cash.paparazzi.plugin.test_AnimatedVectorDrawableTest_rotatesAcrossFrames_rotation.png
geoff-powell: This was to cover frame timing for both AnimatedVectorDrawables and compose.
Added this test using 16.2.1 to ensure my frame timing updates didn't effect animations.
Lottie is also a test that would catch any frame time issues.

geoff-powell: Happy to remove these as they might be verbose

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
geoff-powell: This is reversed so the window view on top of the stack is traversed first

sample/src/test/java/app/cash/paparazzi/sample/ComposeDialogShrinkTest.kt
geoff-powell: Thats a good question. Something I can try with an older version of AD. My assumption here is that it is related to the `AnchoredDraggable` issue where the measure pass is done eagerly when the view is added to the render session view group. 

Current AS renders the same thing, so it is a LayoutLib issue
Image

Workaround being to not use SHRINK with only AlertDialog compsable

paparazzi-gradle-plugin/src/test/projects/verify-animation-gif/src/test/snapshots/videos/app.cash.paparazzi.plugin.test_AnimatedVectorDrawableTest_rotatesAcrossFrames_rotation.png
colinmarsch: Maybe we can reduce this if the existing gif tests provide enough unique coverage

paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -437,14 +487,6 @@ public class PaparazziSdk @JvmOverloads constructor(
...
       constructor.isAccessible = true
       val bridgeSession = constructor.newInstance(renderSession, result) as BridgeRenderSession
-      val viewGroup = bridgeSession.rootViews[0].viewObject as ViewGroup
-      // Workaround since layoutlib's [DisplayManagerGlobal] is missing [registerForRefreshRateChanges].
-      // This method is called by [Display.getRefreshRate] if [mRefreshRateChangesRegistered] is true.
-      // Remove once an updated layoutlib contains this upstream fix: https://android-review.googlesource.com/c/platform/frameworks/layoutlib/+/3876099
-      Display::class.java.getDeclaredField("mRefreshRateChangesRegistered").apply {
-        isAccessible = true
-        set(viewGroup.display, true)
-      }
```
geoff-powell, line 447: layoutlib bump doesn't require workaround any more 🎉

sample/src/test/java/app/cash/paparazzi/sample/ComposeDialogShrinkTest.kt
colinmarsch: I'd be curious to know if this worked on older versions of AS preview

geoff-powell: Good call, I can check this now on AS Panda and Otter

geoff-powell: Looks like AS Panda repros the same issue and AS Otter correctly renders the dialog.
image

Should be reported in Layoutlib issue tracker.

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
geoff-powell, line 49: @colinmarsch and @jrodbx After looking into piping `bridgeRenderSession.systemRootViews` value, I opted to use the current impl so the RenderExtension interface wouldn't need significant change.

colinmarsch: In the near term this seems fine, maybe this is something we can take on in future releases.

sample/src/test/java/app/cash/paparazzi/sample/ComposeDialogShrinkTest.kt
colinmarsch: Could we report this to layoutlib? It would good to have this fixed upstream and we can benefit from it.

colinmarsch (approved):

paparazzi-gradle-plugin/src/test/projects/compose/src/test/java/app/cash/paparazzi/plugin/test/ComposeRenderingModeSizingTest.kt
colinmarsch: I think we should take another look at this `V_SCROLL` case. In the case where we have long content (which happens in some snapshots we take internally at Cash App), it looks like the height gets cut off at the height of the viewport instead of including the entire height of the content. I have a draft PR [here](https://github.com/cashapp/paparazzi/pull/2413/changes) with a hacky change that addresses the issue but it doesn't feel mergeable to me, just good for raising the concern.

<details>
<summary>Image</summary>
image
</details>

geoff-powell: hmm, ya a better reproducible example here is helpful. Taking a look at your PR. Thanks for taking a look on cash side.

geoff-powell: interesting, so it must have to do with frame timing. My fix doesn't cover this use case.

geoff-powell: Fixed in https://github.com/cashapp/paparazzi/pull/2375/changes/e13665905578b7415c421f5d7ed24f2a4f9bfc15

colinmarsch:
Just wanted to provide an update on a few issues I was looking into here that were popping up in Cash App's test suite:
1. It seems the latest commit here for the `V_SCROLL` issue didn't fully resolve the issue mentioned in [this thread](https://github.com/cashapp/paparazzi/pull/2375#discussion_r3560161201). I have a fix that does resolve the issue [here](https://github.com/cashapp/paparazzi/pull/2422) that is ready for review and could be merged into this PR.
2. The regression in the ignored dialog test causes a large failure rate among the tests we have for dialogs and sheets in Cash App. [This PR](https://github.com/cashapp/paparazzi/pull/2423) is an attempt to resolve that issue fully, however it's not in final state yet (still running in some issues to track down in Cash App, hoping to get this identified tomorrow, it might be a cash side fix alongside the PR I reference here).
3. There seems to be a regression to drawing accessibility element bounding boxes from this change. I have a fix for this in [this ready for review PR](https://github.com/cashapp/paparazzi/pull/2424) that could be merged into this PR or merged separately.

cc @geoff-powell @jrodbx

gradle.properties
```diff
@@ -30,3 +30,6 @@ app.cash.paparazzi.internal=true
 
 # Dokka
 org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
```
colinmarsch: Do we need this change in this PR?

colinmarsch (approved):
One more comment, otherwise the latest state looks good to me!

gradle.properties
geoff-powell: oops nope, let me revert that.

PR #2377, opened 2026-07-03
Fix `cleanRecord` race condition between snapshot recording and deletion under parallel builds

FilippoVigani:
## Why
Running `cleanRecordPaparazzi<Variant>` across many modules with `--parallel` can leave some modules with an empty `snapshots/images/` directory even though their tests passed. The goldens get deleted right after they're recorded.

The only ordering edge is between two lifecycle tasks that have no actions themselves:

```
recordPaparazzi<Variant> mustRunAfter deletePaparazziSnapshots
```

The real work happens one level down, in `test<Variant>UnitTest` (writes goldens) and `delete<Variant>PaparazziSnapshots` (deletes them), and `mustRunAfter` doesn't propagate to the tasks each side depends on. So the real deleter and the real writer have no ordering between them, and under `--parallel` the delete can land last. It stays hidden locally because the delete task has no dependencies and usually runs early.

```mermaid
flowchart TD
    cleanRecord["cleanRecordPaparazzi&lt;Variant&gt;"]
    record["recordPaparazzi&lt;Variant&gt;"]
    deleteAnchor["deletePaparazziSnapshots"]
    test["test&lt;Variant&gt;UnitTest<br/>writes goldens"]
    delete["delete&lt;Variant&gt;PaparazziSnapshots<br/>Delete task"]

    cleanRecord -->|dependsOn| deleteAnchor
    cleanRecord -->|dependsOn| record
    record -->|dependsOn| test
    deleteAnchor -->|dependsOn| delete
    record -.->|mustRunAfter| deleteAnchor
    delete -.->|mustRunAfter · added by this PR| test

    classDef lifecycle fill:#e0e0e0,stroke:#999,color:#000
    classDef worker fill:#ffb347,stroke:#d17b00,color:#000
    class cleanRecord,record,deleteAnchor lifecycle
    class test,delete worker
```

## What
- Add `cleanRecordParallel` test, which reproduces the race. The fixture uses `shouldRunAfter` to nudge the delete to run after the test and avoid flakiness. Gradle drops that soft ordering once the fix's `mustRunAfter` is in place, so the test consistently fails without the fix and passes with it.
- Order the writer after the deleter directly: `testTaskProvider.configureEach { it.mustRunAfter(deleteVariantSnapshot) }`. `mustRunAfter` only applies when both tasks are in the graph, so `test`/`verify` runs are unaffected.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -309,6 +309,10 @@ public class PaparazziPlugin @Inject constructor(
 
       recordTaskProvider.configure { it.dependsOn(testTaskProvider) }
       verifyTaskProvider.configure { it.dependsOn(testTaskProvider) }
+      // Order the real writer after the real deleter when running with --parallel.
+      // The mustRunAfter between recordPaparazzi<Variant> and deletePaparazziSnapshots does not
+      // propagate to their dependencies, so without this we could erase freshly recorded snapshots.
+      testTaskProvider.configureEach { it.mustRunAfter(deleteVariantSnapshot) }
     }
   }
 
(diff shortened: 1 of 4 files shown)


oldergod (approved):

geoff-powell (approved):
Nice find here, LGTM

colinmarsch (approved):

colinmarsch:
Thanks for the change here! This seems like a nice small targeted change and I think we can merge it now to pull it forward to 2.0.0-alpha5.1 to get it out sooner. Does this sound good @geoff-powell @jrodbx? I'll merge it by EOD if there are no objections.

PR #2379, opened 2026-07-11
No more infinite loop when verifying single frame goldens

### LINKED ISSUE #2378: `verify` hangs forever (infinite loop in `ApngVerifier`) when a `gif()` golden contains a single frame
**Description**

A `Paparazzi.gif()` call whose parameters produce a **single frame** (`frameCount = (end - start) * fps / 1000 + 1 == 1`, e.g. `gif(view, start = 900, end = 1_000, fps = 1)`) records fine, but `verifyPaparazzi*` then hangs forever at 100% CPU when checking that golden.

Two things combine:

1. A single-frame recording is written without APNG animation metadata (no `fcTL` chunk), so when the golden is read back, `ApngReader.getFps()` falls through to its default and the golden reports **fps = 0**:

   ```kotlin
   fun getFps(): Int {
     return frameInfo?.delayDenominator?.toInt() ?: 0
   }
   ```

2. `ApngVerifier` computes the common frame rate with a subtraction-based GCD that never terminates when one operand is 0:

   ```kotlin
   init {
     currentGoldenFrame = pngReader.readNextFrame()
     commonFrameRate = leastCommonMultiple(fps, pngReader.getFps()) // lcm(1, 0)
     ...
   }

   private fun greatestCommonDenominator(first: Int, second: Int): Int {
     var first = first
     var second = second
     while (first != second) {
       if (first > second) {
         first -= second   // first -= 0 → loops forever
       } else {
         second -= first
       }
     }
     return first
   }
   ```

   `gcd(1, 0)`: `first = 1`, `second = 0` → `first > second` → `first -= 0` → `first` stays `1` forever. (`leastCommonMultiple` would also divide by zero for `lcm(0, 0)`, but the infinite loop is hit first.)

Thread dump of the hung test worker:

```
"Test worker" ... RUNNABLE
   at app.cash.paparazzi.internal.apng.ApngVerifier.greatestCommonDenominator(ApngVerifier.kt:200)
   at app.cash.paparazzi.internal.apng.ApngVerifier.leastCommonMultiple(ApngVerifier.kt:192)
   at app.cash.paparazzi.internal.apng.ApngVerifier.<init>(ApngVerifier.kt:50)
   at app.cash.paparazzi.SnapshotVerifier$newFrameHandler$1.<init>(SnapshotVerifier.kt:53)
   at app.cash.paparazzi.SnapshotVerifier.newFrameHandler(SnapshotVerifier.kt:48)
   at app.cash.paparazzi.Paparazzi.createFrameHandler(Paparazzi.kt:170)
   at app.cash.paparazzi.Paparazzi.gif(Paparazzi.kt:152)
```

This is nasty in practice: record mode and plain unit-test runs never construct `ApngVerifier`, so everything passes locally — then CI verify hangs with no diagnostic output, burns a worker until the CI-level job timeout kills it (~35 minutes in our case), and the log points at whichever test last flushed output rather than the hung one.

**Steps to Reproduce**

Standalone failing test — any module with the Paparazzi plugin applied:

```kotlin
import android.widget.TextView
import org.junit.Rule
import org.junit.Test

class SingleFrameGifTest {

  @Test
  fun singleFrameGifHangsOnVerify() {
    val view = TextView(paparazzi.context).apply { text = "hello" }
    // frameCount = (end - start) * fps / 1000 + 1 = 100 * 1 / 1000 + 1 = 1 frame
    paparazzi.gif(view, start = 900, end = 1_000, fps = 1)
  }
}
```

1. `./gradlew :module:recordPaparazziDebug` → passes, golden is written (a plain PNG — no APNG animation chunks).
2. `./gradlew :module:verifyPaparazziDebug` → the test never completes; the test worker spins at 100% CPU.

**Expected behavior**

Verify terminates. Ideally one (or more) of:

- `getFps()` / `ApngVerifier` guards against a 0 frame rate — treat a golden with no `fcTL` as `fps = 1`, or fail fast with a clear error ("golden X is not an animated PNG").
- The GCD uses Euclid's modulo form, which terminates for 0 inputs: `tailrec fun gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)`.
- `gif()` fails fast (or warns) when the computed `frameCount` is 1, since a single-frame "gif" is almost certainly a mistake and `snapshot(offsetMillis = ...)` is the better tool.

**Additional information:**

- Paparazzi Version: 2.0.x snapshot build (the relevant code in `ApngVerifier.kt` / `ApngReader.kt` is unchanged on `main`)
- OS: reproduced on macOS 26.5 (arm64) and Linux CI
- Compile SDK: 37
- Gradle Version: 9.6.1
- Android Gradle Plugin Version: 9.2.0

**Screenshots**

N/A — the failure mode is a hang; the thread dump above is the most useful artifact.

oldergod:
Fixes https://github.com/cashapp/paparazzi/issues/2378

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
@@ -39,6 +39,7 @@ internal class ApngVerifier(
   private val blankFrame by lazy { createBlankFrame(pngReader.width, pngReader.height) }
 
   private var deltaWriter: ApngWriter? = null
+  private var goldenFps: Int = -1
   private var commonFrameRate: Int = -1
   private var actualDeltasPerFrame: Int = 1
   private var expectedDeltasPerFrame: Int = 1
@@ -46,11 +47,17 @@ internal class ApngVerifier(
   private var invalidFrames = 0
 
   init {
+    require(fps > 0) { "fps must be positive, was: $fps" }
     currentGoldenFrame = pngReader.readNextFrame()
-    commonFrameRate = leastCommonMultiple(fps, pngReader.getFps())
+    // A single-frame golden is written as a still PNG with no animation chunks, dropping the fps
+    // it was recorded with, so ApngReader reports an fps of 0. Fall back to the actual fps: any
+    // fps describes the same single-frame file, and 0 would make the frame-rate math below loop
+    // forever or divide by zero.
+    goldenFps = pngReader.getFps().takeIf { it > 0 } ?: fps
+    commonFrameRate = leastCommonMultiple(fps, goldenFps)
     actualDeltasPerFrame = commonFrameRate / fps
-    expectedDeltasPerFrame = commonFrameRate / pngReader.getFps()
-    if (frameCount != pngReader.frameCount || fps != pngReader.getFps()) {
+    expectedDeltasPerFrame = commonFrameRate / goldenFps
+    if (frameCount != pngReader.frameCount || fps != goldenFps) {
       deltaWriter = ApngWriter(deltaFilePath, commonFrameRate, fileSystem)
     }
   }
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
colinmarsch, line 44: This file gets opened even in the case the `require(fps > 0)` check in the `init` fails. Could we inline this to avoid extra work here?

colinmarsch (approved):
One comment, otherwise this looks good to me to land for the alpha5.2 milestone

paparazzi/src/main/java/app/cash/paparazzi/internal/apng/ApngVerifier.kt
oldergod: Good catch, thanks. The check now runs before the file opens: 0ac4406b8.

PR #2383, opened 2026-07-22
Prune hidden accessibility subtrees

### LINKED ISSUE #2380: AccessibilityRenderExtension includes descendants hidden with IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS
`AccessibilityRenderExtension` includes accessibility elements from a view subtree marked with `IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS`.

This is visible with modal UI such as a bottom sheet. Although the underlying content is hidden from accessibility services, Paparazzi still adds its semantic bounds and labels to the accessibility snapshot. The bounds can appear over the sheet as if those elements were accessible through it.

## Actual behavior

In this snapshot, the colored boxes near the bottom belong to controls behind the white sheet. Their labels are also included in the accessibility legend.

Image

## Expected behavior

When a view uses `IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS`, neither that view nor its descendants should appear in the accessibility snapshot.

The snapshot should include only the accessibility elements exposed while the modal sheet is displayed.

## Reproduction

For example:

```kotlin
@get:Rule
val paparazzi = Paparazzi(
)

@Test
fun hiddenAccessibilitySubtree() {
  val root = FrameLayout(paparazzi.context)

  val background = ComposeView(paparazzi.context).apply {
    setContent {
      Button(onClick = {}) {
        Text("Background action")
      }
    }
    importantForAccessibility =
      View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS
  }

  val modal = ComposeView(paparazzi.context).apply {
    setContent {
      Surface {
        Text("Modal content")
      }
    }
  }

  root.addView(background)
  root.addView(modal)

  paparazzi.snapshot(root)
}
```

## Suspected cause

`AccessibilityElementCollector.processAccessibleChildren` checks `isImportantForAccessibility` before adding a traditional view, but it continues traversing the view's Compose semantics and child views regardless:

https://github.com/cashapp/paparazzi/blob/master/paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityElementCollector.kt#L60-L99

The traversal may need to stop when the current view has `IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS`.

## Additional context

This appears related to the general bottom-sheet accessibility behavior covered by #2252, but the behavior here is deterministic and specifically concerns hidden descendant traversal.

The issue was observed with a Paparazzi 2.0-based build. The current collector implementation on `master` appears to follow the same traversal path.

arpitagarwal1301:
Fixes #2380

- Stop accessibility collection at views marked `IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS`.
- Preserve descendant traversal for plain `IMPORTANT_FOR_ACCESSIBILITY_NO`.
- Add snapshot coverage for hidden classic and Compose descendants.

## Verification

- `./gradlew :paparazzi-gradle-plugin:test --tests 'app.cash.paparazzi.gradle.PaparazziPluginTest.accessibilityRendering'`
- `./gradlew spotlessCheck :paparazzi:test`
- `git diff --check`

## Limitations

- The full cross-platform `check` matrix was not run locally; upstream CI remains authoritative.


geoff-powell (approved):

colinmarsch (approved):
Thanks for the fix here!

PR #2385, opened 2026-07-27
Cleanup main dispatcher for Paparazzi test runs

### LINKED ISSUE #2382: `Dispatchers.Main` remains bound to a previous test’s layoutlib Looper
🤖 Disclaimer: I used AI for drafting this bug report and creating its reproducer test cases.

**Description**

Paparazzi creates a new layoutlib main `Looper` for each test, but the process-global `Dispatchers.Main` remains bound to the first test’s Looper.

Code dispatched through `Dispatchers.Main.immediate` can therefore stop executing in later Paparazzi tests. This affects content composed during measurement, such as content inside `Scaffold`, `LazyColumn`, or another `SubcomposeLayout`.

**Steps to Reproduce**

```kotlin
import androidx.compose.foundation.background
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.size
import androidx.compose.material3.Scaffold
import androidx.compose.runtime.RememberObserver
import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.remember
import androidx.compose.runtime.setValue
import androidx.compose.ui.unit.dp
import kotlinx.coroutines.CoroutineScope
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.SupervisorJob
import kotlinx.coroutines.cancel
import kotlinx.coroutines.launch
import org.junit.Assert.assertEquals
import org.junit.FixMethodOrder
import org.junit.Rule
import org.junit.Test
import org.junit.runners.MethodSorters
import java.util.concurrent.atomic.AtomicInteger

@FixMethodOrder(MethodSorters.NAME_ASCENDING)
class PaparazziStaleMainDispatcherReproducerTest {

  @Test fun `a direct composition initializes Dispatchers Main`() {
    val executions = AtomicInteger()

    paparazzi.snapshot {
      DispatcherBackedContent(executions)
    }

    assertEquals(1, executions.get())
  }

  // The ordering is intentional: the first test initializes `Dispatchers.Main`, 
  // and the second test runs with a new layoutlib main Looper.
  @Test fun `b measure-time composition can dispatch to the stale main Looper`() {
    val executions = AtomicInteger()

    paparazzi.snapshot {
      Scaffold {
        DispatcherBackedContent(executions)
      }
    }

    assertEquals(1, executions.get()) // Fails: actual is 0.
  }
}

@Composable
private fun DispatcherBackedContent(executions: AtomicInteger) {
  var loaded by remember { mutableStateOf(false) }

  remember {
    object : RememberObserver {
      private var scope: CoroutineScope? = null

      override fun onRemembered() {
        scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
        scope!!.launch {
          executions.incrementAndGet()
          loaded = true
        }
      }

      override fun onForgotten() {
        scope?.cancel()
      }

      override fun onAbandoned() = Unit
    }
  }

  Box(
    Modifier
      .size(100.dp)
      .background(if (loaded) Color.Green else Color.Transparent)
  )
}
```

Output:

```text
First test:
mainDispatcher=1742056495
mainDispatchNeeded=false
mainLooper=621843976
requests=1

Second test:
mainDispatcher=1742056495
mainDispatchNeeded=true
mainLooper=1951735884
requests=0
```

`Dispatchers.Main` remains the same object, while `Looper.getMainLooper()` changes between tests. In the second test, `Dispatchers.Main.immediate` no longer recognizes the current thread as its main thread.

Direct composition still works because its dispatched work is queued before Paparazzi drains handler callbacks. `Scaffold` composes its body later during measurement, after that drain point, so the work never runs before capture.

Reversing the test order causes both tests to execute their dispatched work.

**Additional information:**
- Paparazzi: `2.0.0-alpha02`
- kotlinx-coroutines-android: `1.10.2`
- Compose UI/runtime: `1.8.0`
- Material 3: `1.3.2`
- AGP: `8.13.2`
- Gradle: `8.14.3`
- Compile SDK: `35`
- macOS, arm64

geoff-powell:
Fixes #2382

Cleans up main dispatcher to avoid leaking across tests

### DIFF
--- a/paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/CoroutineDelayMainTest.kt
+++ b/paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/CoroutineDelayMainTest.kt
@@ -1,18 +1,28 @@
 package app.cash.paparazzi.plugin.test
 
 import android.os.SystemClock
+import androidx.compose.material.Scaffold
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.RememberObserver
+import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.ComposeView
 import app.cash.paparazzi.Paparazzi
 import org.junit.Assert.assertEquals
+import org.junit.FixMethodOrder
 import org.junit.Rule
 import org.junit.Test
+import org.junit.runners.MethodSorters
+import java.util.concurrent.atomic.AtomicInteger
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.test.resetMain
 
 @OptIn(ExperimentalCoroutinesApi::class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 class CoroutineDelayMainTest {
   @get:Rule
   val paparazzi = Paparazzi()
@@ -20,6 +30,28 @@ class CoroutineDelayMainTest {
   init {
     // coroutines-test installs a TestMainDispatcher which is incompatible with ComposeUi
     Dispatchers.resetMain()
+  }
+
+  @Test fun aDirectCompositionInitializesMainDispatcher() {
+    val executions = AtomicInteger()
+
+    paparazzi.snapshot {
+      DispatcherBackedContent(executions)
+    }
+
+    assertEquals(1, executions.get())
+  }
+
+  @Test fun bScaffoldCompositionUsesMainDispatcher() {
+    val executions = AtomicInteger()
+
+    paparazzi.snapshot {
+      Scaffold {
+        DispatcherBackedContent(executions)
+      }
+    }
+
+    assertEquals(1, executions.get())
   }
 
   @Test fun delayUsesMainDispatcher() {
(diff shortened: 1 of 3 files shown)


paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
```diff
@@ -623,10 +625,6 @@ public class PaparazziSdk @JvmOverloads constructor(
 
     internal lateinit var sessionParamsBuilder: SessionParamsBuilder
 
-    private val MAIN_DISPATCHER by lazy {
-      Handler.getMain().asCoroutineDispatcher("Paparazzi-Main")
```
jrodbx, line 627: Intentional to inline?  Can/should Looper.getMainLooper be added here or is there a timing bug that inlining would avoid?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/CoroutineDelayMainTest.kt
```diff
@@ -1,18 +1,28 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.test.resetMain
 
 @OptIn(ExperimentalCoroutinesApi::class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
jrodbx, line 25: I think we can avoid this if onForgotten or onAbandoned cleans up state?

paparazzi-gradle-plugin/src/test/projects/coroutine-delay-main/src/test/java/app/cash/paparazzi/plugin/test/CoroutineDelayMainTest.kt
```diff
@@ -22,6 +32,28 @@ class CoroutineDelayMainTest {
...
+    val executions = AtomicInteger()
+
+    paparazzi.snapshot {
+      DispatcherBackedContent(executions)
+    }
+
+    assertEquals(1, executions.get())
+  }
+
+  @Test fun bScaffoldCompositionUsesMainDispatcher() {
```
jrodbx, line 45: Nit: method names have a and b suffix.  Rename to more human readable test case phrasing?

colinmarsch:
I think the test cases added here pass even without this change, they also appear slightly different than what was in the reporting issue. Could we confirm they provide the right regression coverage for this issue?

PR #2388, opened 2026-08-09
fix: Workaround WindowManagerImpl.removeView NPE during Dialog teardown (#2373)

### LINKED ISSUE #2373: Rendering a Compose Dialog crashes on teardown (WindowManagerImpl.removeView NPE) with layoutlib 16.2.1

robto09:
**Closes #2373**

**What's happening**

Rendering a Compose `Dialog` crashes during teardown with:

```
NullPointerException: Cannot invoke "java.util.ArrayList.size()" because "childrenList" is null
    at android.view.WindowManagerImpl.removeView(WindowManagerImpl.java:209)
```

That's a layoutlib 16.2.1 bug. `removeView` doesn't null-check the result of `ViewGroup.buildOrderedChildList()` before calling `size()` on it. When the dialog's view is removed from the window and one child is left behind, `buildOrderedChildList()` returns null and everything blows up.

**The fix**

I registered a `net.bytebuddy.asm.Advice` on `WindowManagerImpl.removeView` that swallows the NPE. By the time it's thrown, the view has already been removed. Only the `ViewRootImpl` child reassignment fails, so ignoring it is safe.

Side note: I first tried a `MethodDelegation` interceptor with `@SuperCall`, but ByteBuddy can't bind that here since `removeView` implements an interface method and there's no super implementation to delegate to. `@Advice` wraps the method body in place, which sidesteps that entirely.

While doing this I also restructured `InterceptorRegistrar` a bit: delegations and advice for the same class are now grouped into a single redefine pass, because two separate redefines of one class undo each other.

**Test**

Added `WindowManagerRemoveViewNpeTest`, which reproduces the exact layoutlib state: add two views to the window manager, remove the last one. It fails with the same NPE without the fix and passes with it.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
@@ -55,7 +55,7 @@ import app.cash.paparazzi.internal.PaparazziSavedStateRegistryOwner
 import app.cash.paparazzi.internal.Renderer
 import app.cash.paparazzi.internal.SessionParamsBuilder
 import app.cash.paparazzi.internal.interceptors.EditModeInterceptor
-import app.cash.paparazzi.internal.interceptors.WindowManagerImplRemoveViewAdvice
+import app.cash.paparazzi.internal.interceptors.WindowManagerImplRemoveViewAdviceAccessor
 import app.cash.paparazzi.internal.parsers.LayoutPullParser
 import com.android.ide.common.rendering.api.RenderSession
 import com.android.ide.common.rendering.api.Result
@@ -582,7 +582,7 @@ public class PaparazziSdk @JvmOverloads constructor(
     InterceptorRegistrar.addAdvice(
       "android.view.WindowManagerImpl",
       "removeView",
-      WindowManagerImplRemoveViewAdvice::class.java
+      WindowManagerImplRemoveViewAdviceAccessor.adviceClass
     )
   }
 
--- a/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/PaparazziSdk.kt
@@ -55,6 +55,7 @@ import app.cash.paparazzi.internal.PaparazziSavedStateRegistryOwner
 import app.cash.paparazzi.internal.Renderer
 import app.cash.paparazzi.internal.SessionParamsBuilder
 import app.cash.paparazzi.internal.interceptors.EditModeInterceptor
+import app.cash.paparazzi.internal.interceptors.WindowManagerImplRemoveViewAdvice
 import app.cash.paparazzi.internal.parsers.LayoutPullParser
 import com.android.ide.common.rendering.api.RenderSession
 import com.android.ide.common.rendering.api.Result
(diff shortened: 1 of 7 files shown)


geoff-powell:
Closing this PR as I have verified your test:
```
@Test
  fun removeViewDoesNotCrashWhenOneChildRemains() {
    // Render once so the render root is attached; views added to the window manager then
    // have a ViewRootImpl, which is required for the buggy code path in removeView to run.
    paparazzi.snapshot(TextView(paparazzi.context))

    val windowManager = paparazzi.context.getSystemService(Context.WINDOW_SERVICE) as WindowManagerImpl

    val first = TextView(paparazzi.context)
    val last = TextView(paparazzi.context)
    val layoutParams = WindowManager.LayoutParams(
      WindowManager.LayoutParams.MATCH_PARENT,
      WindowManager.LayoutParams.MATCH_PARENT
    )

    windowManager.addView(first, layoutParams)
    windowManager.addView(last, layoutParams)

    // Removing the last added view leaves exactly one child in the window root view, which
    // makes ViewGroup.buildOrderedChildList() return null in the following removeView
    // reentrant path. Without the WindowManagerImplRemoveViewAdvice workaround this throws:
    //   NullPointerException: Cannot invoke "java.util.ArrayList.size()" because
    //   "childrenList" is null
    windowManager.removeView(last)

    // Clean up the remaining window.
    windowManager.removeView(first)
  }
}
```

Runs successfully on Layoutlib 16.2.3: https://github.com/cashapp/paparazzi/pull/2375

PR #2393, opened 2026-08-12
Add touch target size validation to accessibility snapshots

### LINKED ISSUE #2257: Add touch target size validation to accessibility snapshots
`AccessibilityRenderExtension` currently collects accessibility text and renders element bounds, but it does not validate minimum touch target size for interactive elements. Because of that, snapshots can appear accessibility-complete while still containing controls smaller than recommended guidance (48x48dp per Material / WCAG).

## Proposed enhancement

- Add touch target size validation for interactive elements in the accessibility collection/rendering pipeline.
- Flag any interactive element whose rendered bounds are below a configurable minimum size (default 48dp).
- Add visual warning styling for flagged elements in the overlay (for example: red border and/or warning badge/icon).
- Append a legend warning tag to flagged entries: <touch-target-too-small> (consistent with existing tags like <disabled> and <editable>).

## Suggested API

- Add configuration on AccessibilityRenderExtension for minimum touch target size with a default of 48dp.
- Keep default behaviour backward-compatible so existing accessibility snapshots still run without migration work.

robto09:
Fixes #2257

## What this does

The accessibility extension collects element text and draws bounds, but it never checked whether interactive elements were big enough to tap. I've added touch target validation so snapshots catch controls that are smaller than the recommended 48x48dp from Material and WCAG guidance.

Interactive elements (clickable, toggleable, editable, adjustable, or with custom actions / link annotations) are now checked against the minimum size. If either rendered dimension comes up short, the element gets a red border in the overlay and a `<touch-target>` tag in the legend, matching the existing `<on-click>` / `<toggleable>` style tags.

The minimum size is configurable through a new `minTouchTargetSizeDp` parameter that defaults to `48f`, so existing `AccessibilityRenderExtension()` call sites keep working unchanged.

## What changed

- **`AccessibilityRenderExtension`**: new `minTouchTargetSizeDp: Float = 48f` parameter, passed through to the collector.
- **`AccessibilityElementCollector`**: interactivity detection for classic Views (node info flags, `AbsSeekBar`, labeled actions) and Compose (`OnClick`, `OnLongClick`, `ToggleableState`, `IsEditable`, `SetProgress`, `CustomActions`, link annotations).
- **`TouchTargetValidator`** (new): a small single-purpose class that owns the threshold math and flag assignment, plus focused unit tests. I extracted it so the collector stays a collector and doesn't turn into a god class.
- **`AccessibilityElement`**: internal `isInteractive` / `isTouchTargetTooSmall` fields, defaulting to `false`.
- **`AccessibilityOverlayDrawable`**: draws a red border around flagged elements.
- **`AccessibilityOverlayDetailsView`**: appends `<touch-target>` to flagged legend entries.
- **`RenderSettings`**: `TOUCH_TARGET_SIZE_DP = 48f` and `WARNING_BORDER_COLOR` constants.
- **Tests**: two new golden tests (default 48dp and a custom 20dp threshold), four validator unit tests, and eight re-recorded fixture goldens in the accessibility-rendering project.
- **Docs**: `docs/accessibility.md` now covers the touch target validation section and the `<touch-target>` label.

## Things worth knowing

- **Either dimension counts.** A target is flagged when width *or* height is below the minimum. A 421x44dp slider is just as hard to tap as a 40x40 button, so I check both.
- **I measure rendered bounds, not the expanded touch target.** Components like Material 3 `TextButton`s (40dp) or `Checkbox`es (24dp) get flagged based on what's actually drawn, which is the whole point of snapshot validation. If your UI intentionally renders below 48dp, you can lower `minTouchTargetSizeDp`.
- **The detection heuristics are based on what layoutlib actually reports.** The compile-time Android stub (API 16) doesn't expose `ACTION_SET_PROGRESS` or `ACTION_CUSTOM_ACTION_START`, and layoutlib's node info includes spurious actions on every view, so I verified flags, widget types, and labeled actions empirically against real rendered hierarchies.

## Verification

- [x] `:paparazzi:test` passes, 119 tests, 0 failures (including the 9 accessibility and validator tests)
- [x] `:paparazzi:build` green, including `apiCheck` and `spotless`
- [x] accessibility-rendering fixture (`verifyPaparazziDebug`) passes
- [x] sample `ComposeA11yTest` passes
- [x] `:paparazzi:apiDump` produces no diff against the committed API


geoff-powell:
LayoutLib already supports accessibility validation so this is a lot of work that is largely figured out by LL team. We have plans to look into providing accessibility metadata in json format instead of in snapshots as it works today with the RenderExtension. (see [here](https://github.com/cashapp/paparazzi/pull/2268))

PR #2395, opened 2026-08-17
fix: make overwriteOnMaxPercentDifference APNG-aware

FrancoisBlavoet:
## Problem

When `paparazzi.test.record.overwriteOnMaxPercentDifference=true` is set and an animated golden already exists, re-recording an animation that changed only in a later frame silently skips the update — the golden file is left stale.

Root cause: the overwrite check calls `ImageIO.read(goldenFile)` and `ImageIO.read(snapshotFile)` to pixel-diff the two snapshots, but `ImageIO.read()` decodes only the **first frame** of an APNG. Any change confined to a frame beyond the first is invisible to this comparison, so the threshold is never exceeded and the golden is never written.

## Fix

For animated snapshots (`fps != -1`), replace the `ImageIO.read()`-based pixel comparison with a byte-level file equality check (`Files.equal()`). Since Paparazzi's APNG serialization is deterministic, an unchanged animation re-serializes to a bit-identical file (no spurious git diffs or unnecessary writes), while any frame change produces different bytes and correctly triggers a golden update.

Static (single-frame) goldens are unaffected and keep the existing pixel-diff comparison.

## Test

Added `videoOverwritesOnRecordWithFlagWhenOnlyLaterFrameChanges`: records an animation with `frame1=anyImage, frame2=anyImage`, then re-records with `frame1=anyImage, frame2=otherImage` (first frame unchanged, only the second frame differs) and asserts the golden is updated. This test fails before the fix and passes after.

### DIFF
--- a/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
+++ b/paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
@@ -18,10 +18,12 @@ package app.cash.paparazzi
 import app.cash.paparazzi.SnapshotHandler.FrameHandler
 import app.cash.paparazzi.internal.ImageUtils
 import app.cash.paparazzi.internal.PaparazziJson
+import app.cash.paparazzi.internal.apng.ApngReader
 import app.cash.paparazzi.internal.apng.ApngWriter
 import com.google.common.base.CharMatcher
 import com.google.common.io.Files
 import okio.BufferedSink
+import okio.FileSystem
 import okio.HashingSink
 import okio.Path.Companion.toPath
 import okio.blackholeSink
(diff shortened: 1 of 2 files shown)


paparazzi/src/main/java/app/cash/paparazzi/HtmlReportWriter.kt
```diff
@@ -131,6 +137,25 @@ public class HtmlReportWriter @JvmOverloads constructor(
     }
   }
 
+  private fun apngExceedsThreshold(snapshotFile: File, goldenFile: File): Boolean {
+    ApngReader(FileSystem.SYSTEM.openReadOnly(goldenFile.path.toPath())).use { goldenReader ->
+      ApngReader(FileSystem.SYSTEM.openReadOnly(snapshotFile.path.toPath())).use { snapshotReader ->
+        if (goldenReader.frameCount != snapshotReader.frameCount ||
+          goldenReader.getFps() != snapshotReader.getFps()
```
colinmarsch, line 144: `getFps()` is called before either reader consumes its first `fcTL`, so both calls return 0. As a result, changing only an animation’s FPS does not overwrite the golden. I was able to reproduce this with an FPS-only test that fails on both the base and this PR. Could we compare FPS after reading the first frame and add coverage for this case?

colinmarsch (approved):
One comment otherwise this looks good to me to merge in our alpha5.2 release upcoming for next week

PR #2413, opened 2026-09-02
Fix Compose V_SCROLL rendering with Layoutlib 16.2.3

colinmarsch:
## Summary

- stack this exploratory fix on #2375 and limit the behavior to `V_SCROLL` sessions containing a Compose host
- give Compose an initial render/layout/frame cycle before the final capture so below-fold content is drawn with Layoutlib 16.2.3
- add a bottom-edge pixel assertion to the Compose `V_SCROLL` fixture so a correctly sized but blank result fails

## Testing

- `./gradlew :paparazzi-gradle-plugin:test --tests "app.cash.paparazzi.gradle.PaparazziPluginTest.compose" spotlessCheck --console=plain`
- published as `2.0.0-SNAPSHOT` to Maven Local and exercised against the affected Cash Android investing `V_SCROLL` tests
  - restored full below-fold content
  - normal-text stock-details snapshots matched exactly
  - large-text stock details retained a 1.959566% delta
  - entity snapshots retained 4.98% to 11.07% deltas, so the downstream result is improved but not golden-identical

This is a review-only draft intended to make the proposed approach easier to inspect alongside #2375.


colinmarsch:
🤖 Closing this because it has been superseded by Geoff adding the related V_SCROLL and Compose regression fix to #2375.

PR #2418, opened 2026-09-04
Add agent guidance for Layoutlib references

colinmarsch:
## Summary

- add root-level agent guidance for consulting Layoutlib and Android framework internals
- direct agents to match upstream source to the Layoutlib version pinned by Paparazzi
- add a Claude entry point that links to the shared guidance

This change should diagnosing changes in layoutlib sources easier during agentic workflows in the paparazzi repo.

## Testing

Not run (documentation-only change).


geoff-powell (approved):
lol If only Claude would just use the AGENTS.md file without the pointer

PR #2422, opened 2026-09-06
Fix Compose V_SCROLL cutoff after follow-up layout

colinmarsch:
## Root cause

Layoutlib 16.2.3 measures `V_SCROLL` hierarchies twice on every render: first at the device height, then with an unbounded vertical axis. PR #2375's pre-frame measurement gives Compose the correct full height initially, but a follow-up layout after pre-draw can cause a bounded root to retain the intermediate device height. Layoutlib keeps the expanded image surface, so only one viewport is painted and the rest of the snapshot is blank.

This change pins the Compose root to its initial unbounded height while rendering, invalidates Layoutlib's cached render dimensions before each render to avoid cumulative expansion, and restores the caller's original layout height afterward.

## Regression test

Adds a self-contained golden test with long scrollable Compose content and a text-measurement-driven follow-up layout. Without the fix, rows after the first viewport and the final marker are blank; with the fix, the complete snapshot is rendered.

## Verification

- `./gradlew :paparazzi-gradle-plugin:test --tests app.cash.paparazzi.gradle.PaparazziPluginTest.compose --rerun-tasks --console=plain`
- Cash Android through Maven Local: affected `AccountViewTest` large-scaling cases pass


geoff-powell:
Closing for [commit](https://github.com/cashapp/paparazzi/pull/2375/commits/f5e4fcf79e7e6b189f3362da724c28ed47f4b89a) in #2375

PR #2423, opened 2026-09-06
Fix screenshot sizing for separate application windows

colinmarsch:
Layoutlib 16.2.3 renders dialogs and application overlays in separate windows, but SHRINK calculates its render target from the main hierarchy. Dialog-only content can fail with a 0×0 image, and overlays can copy a small nested host's bounds and clip their contents in NORMAL mode.

Prepare a device-sized render target in SHRINK when there is exactly one additional visible application window and system UI is disabled. Resize windows with explicit width and height directly, preserving content-sized dialogs and nested host wrappers. Accessibility windows use the left content pane's width and alignment so they do not cover the details pane. Restore modified window parameters between frames and during cleanup, and crop only when SHRINK's main hierarchy is empty.

Re-enable the existing ComposeDialogShrinkTest and refresh its 880×652 golden with small shadow/edge differences. Based on [#2375](https://github.com/cashapp/paparazzi/pull/2375).

Validation of the simplified sizing condition:
- Existing ComposeDialogShrinkTest snapshot verification, Kotlin formatting, and compilation pass.
- All 40 Cash Android OverdraftCoverageSheetViewTest cases pass against unchanged goldens using Maven Local.
- The three regular EarnerUpsell sheet preview variants pass. Its accessibility rendering is byte-for-byte identical to the prior pane-overflow fix, retaining the existing 3.29% golden difference.


geoff-powell:
So this bug is a layoutlib regression and not paparazzi. I would probably avoid a hacky work around here.

colinmarsch:
> So this bug is a layoutlib regression and not paparazzi. I would probably avoid a hacky work around here.

Ok, good point. I'll try to draft an issue tracker ticket for this issue for you to review before I submit it. I see it seems to still be a problem in layoulib 17 from your most recent PR as well.

geoff-powell:
> > So this bug is a layoutlib regression and not paparazzi. I would probably avoid a hacky work around here.
> 
> Ok, good point. I'll try to draft an issue tracker ticket for this issue for you to review before I submit it. I see it seems to still be a problem in layoulib 17 from your most recent PR as well.

Yep, still an issue in the latest AS

PR #2424, opened 2026-09-07
Fix stale accessibility bounds after Compose placement

colinmarsch:
Accessibility rectangles can remain at their initial positions when Compose moves content after pre-draw. In a static results screen, this leaves the results' boxes over the header and search field. Invalidate the existing accessibility foregrounds each frame and collect their bounds when they draw, after the content has been placed. Keep window discovery and legend collection on their existing global-layout path so transient popup roots are not newly included.

Adds one static snapshot test with a measured header and a results layer, plus its PNG golden. The test fails without the fix (3.436222% image difference) and passes with it. All 24 existing accessibility integration fixtures, the three core accessibility snapshot tests, and Kotlin formatting checks pass locally. No existing goldens were changed.

Based on [PR #2375](https://github.com/cashapp/paparazzi/pull/2375), targeting `geoff/upgrade-layoutlib-16.2.3`.


paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
```diff
@@ -64,7 +70,7 @@ public class AccessibilityRenderExtension : RenderExtension {
               rootView = view
             )
 
-            overlayDrawables[id]?.updateElements(elements)
```
geoff-powell, line 67: nit: Do you need this invalidation since it is overwritten in pre-draw callback?

geoff-powell (approved):

paparazzi/src/main/java/app/cash/paparazzi/accessibility/AccessibilityRenderExtension.kt
colinmarsch: Updated!

colinmarsch:
@geoff-powell would merging this into your PR #2375 be ok or would you prefer this to land separately afterwards? I'm ok either way.

PR #2434, opened 2026-09-11
Don't carry generated source producers into the snapshot dir

### LINKED ISSUE #2374: recordPaparazzi fails at configuration-cache store when KSP is applied to a KMP module (kspAndroidHostTest "property 'all'" serialization error)
```kotlin
// PaparazziPlugin.kt (2.0.0-alpha05), line ~475
private fun Project.snapshotDir(testVariant: UnitTest): Provider<Directory> {
  val sources = testVariant.sources.kotlin?.all   // <-- AGP provider "all"
    ?: testVariant.sources.java?.all
    ?: error("No Kotlin or Java sources on ${testVariant.name}")
  val projectDirectory = layout.projectDirectory
  return sources.map { dirs ->
    val sourceSetRoot = dirs.firstOrNull()?.asFile?.parentFile
      ?: error("No source dirs registered for ${testVariant.name}")
    projectDirectory.dir(sourceSetRoot.path).dir("snapshots")
  }
}
```

In a KMP module with KSP applied, `testVariant.sources.kotlin.all` (the provider whose display name
is `property 'all'` in the error) carries a **producer-task dependency on `kspAndroidHostTest`**,
because the KSP-generated host-test sources are part of that aggregate.

That mapped provider is then registered on the host-test `Test` task two different ways, gated by
`presentWhen` (`condition.filter { it }.flatMap { this }`):

```kotlin
// line ~285  — active during a VERIFY run
test.inputs.dir(snapshotOutputDir.presentWhen(isVerifyRun))
  .optional()

// line ~290  — active during a RECORD run
test.outputs.dir(snapshotOutputDir.presentWhen(isRecordRun))
  .withPropertyName("paparazzi.snapshots.output.dir")
  .optional()
```

The asymmetry between record and verify is **input vs output handling at configuration-cache store**:

- **record** → `test.outputs.dir(...)` is the active registration. The configuration cache resolves
  **output** file properties **eagerly** at store time
  (`TaskNodeCodec.writeRegisteredPropertiesOf` → `DeferredUtil.unpackOrNull`), which forces the
  `kspAndroidHostTest`-mapped provider to be resolved before that task has run → hard
  `InvalidUserCodeException`, build fails.
- **verify** → `test.inputs.dir(...)` is the active registration. Input file collections are stored
  **lazily** (as a recipe, resolved at execution), so the producer task is never visited at store
  time → no error, the cache stores cleanly.

So the trigger is the combination: KSP contributes a task-output-mapped source provider, and the
record path registers it (via `snapshotDir()`) as an `outputs.dir` whose value the configuration
cache must resolve eagerly.

**Stacktrace (store, trimmed):**

```
org.gradle.api.InvalidUserCodeException: Querying the mapped value of property 'all' before task ':lib:kspAndroidHostTest' has completed is not supported
  at org.gradle.api.internal.provider.TransformBackedProvider.beforeRead(TransformBackedProvider.java:95)
  at org.gradle.api.internal.provider.TransformBackedProvider.calculateOwnValue(TransformBackedProvider.java:80)
  at org.gradle.api.internal.provider.FlatMapProvider.calculateOwnValue(FlatMapProvider.java:53)        # presentWhen's flatMap
  at org.gradle.util.internal.DeferredUtil.unpackOrNull(DeferredUtil.java:49)
  at org.gradle.internal.serialize.codecs.core.TaskNodeCodecKt.writeRegisteredPropertiesOf(TaskNodeCodec.kt:393)
  at org.gradle.internal.serialize.codecs.core.TaskNodeCodec.writeTask(TaskNodeCodec.kt:104)
  at org.gradle.internal.serialize.codecs.core.WorkNodeCodec.writeGroupedNodes(WorkNodeCodec.kt:327)
  at org.gradle.internal.cc.impl.ConfigurationCacheState.writeWorkGraphOf(ConfigurationCacheState.kt:510)
  ...
```

**Suggested fix:** the snapshot directory path is statically derivable from the source-set layout;
it does not need to be a provider mapped over a task output. Options: (1) compute the snapshot dir
without going through `testVariant.sources.kotlin.all`, so the resulting provider carries no
producer-task dependency; (2) avoid registering the resulting provider as `outputs.dir` in a way
that forces eager resolution at store time; (3) stopgap — mark the record path
`notCompatibleWithConfigurationCache(...)`, mirroring the existing handling of `PrepareResourcesTask`.

**Workaround (consumers):** run `recordPaparazzi` with `--no-configuration-cache`; `verifyPaparazzi`
and regular host-test runs can stay on the configuration cache.

</details>

**Screenshots**

N/A — this is a Gradle configuration/build failure, not a rendering issue.

Repo with reproduce:
[paparazzi-cc-repro.zip](https://github.com/user-attachments/files/29220594/paparazzi-cc-repro.zip)

Bortlesboat:
Fixes #2374.

`snapshotDir()` builds the snapshot directory by mapping `sources.kotlin.all`. When the test source set has generated directories (KSP, or anything added with `addGeneratedSourceDirectory`), that provider carries the generating task as a producer, and so does anything mapped from it. During a record run the result is registered as a task output, and the configuration cache resolves outputs when it stores the entry, which fails:

```
InvalidUserCodeException: Querying the mapped value of list(interface org.gradle.api.file.Directory, item(flatmap(provider(task 'sourceGenAndroidMainHostTest', class SourceGenTask))) + flatmap(provider(?))) before task ':sourceGenAndroidMainHostTest' has completed is not supported
```

The snapshot directory only depends on where the source set lives, not on what the generating task writes, so this computes it with `providerFactory.provider { }` instead of `map`. It uses the same sources and the same "parent of the first source dir" rule, so snapshot locations don't change. The lambda only captures the variant name and the project directory.

I tried `sources.kotlin.static` first, but that list is empty for the KMP library plugin's `androidHostTest`, so it broke `kotlinMultiplatformPluginWithNewAndroidLibraryPlugin`.

Two new fixtures reproduce it without KSP on the fixture classpath, both registering a generated Kotlin dir on the unit test through the variant API:
- `configuration-cache-generated-test-sources`: Android library, `recordPaparazziDebug --configuration-cache`
- `configuration-cache-generated-test-sources-multiplatform`: KMP library plugin, `recordPaparazziAndroidMain --configuration-cache`

Both fail on master with the error above and pass with this change.

Locally I also ran the existing plugin tests that touch the snapshot directory or the configuration cache (`kotlinMultiplatformPluginWithNewAndroidLibraryPlugin`, `configurationCache*`, `record*`, `cleanRecord`, `deleteSnapshots`, `verifyMissingGolden`, `rerun*`, `customBuildDir`, `customReportDir`, `snapshotReport`, and a few others), plus `spotlessCheck`. All pass.

### DIFF
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -474,11 +474,17 @@ public class PaparazziPlugin @Inject constructor(
       ?: error("No Kotlin or Java sources on ${testVariant.name}")
     val testVariantName = testVariant.name
     val projectDirectory = layout.projectDirectory
+    val buildDirectory = layout.buildDirectory
     // Only the location of the source directories matters here. Mapping `sources` would also carry
     // the tasks that produce generated source directories (eg. KSP) into the record task's outputs,
     // which can't be resolved when the configuration cache entry is stored.
     return providerFactory.provider {
-      val sourceSetRoot = sources.get().firstOrNull()?.asFile?.parentFile
+      val sourceDirs = sources.get().map { it.asFile }
+      // Generated source directories can come first (eg. on KMP), and snapshots shouldn't end up in
+      // the build directory.
+      val buildDir = buildDirectory.get().asFile
+      val sourceSetRoot = (sourceDirs.firstOrNull { !it.startsWith(buildDir) } ?: sourceDirs.firstOrNull())
+        ?.parentFile
         ?: error("No source dirs registered for $testVariantName")
       projectDirectory.dir(sourceSetRoot.path).dir("snapshots")
     }
--- a/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
+++ b/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
@@ -472,10 +472,14 @@ public class PaparazziPlugin @Inject constructor(
     val sources = testVariant.sources.kotlin?.all
       ?: testVariant.sources.java?.all
       ?: error("No Kotlin or Java sources on ${testVariant.name}")
+    val testVariantName = testVariant.name
     val projectDirectory = layout.projectDirectory
-    return sources.map { dirs ->
-      val sourceSetRoot = dirs.firstOrNull()?.asFile?.parentFile
-        ?: error("No source dirs registered for ${testVariant.name}")
+    // Only the location of the source directories matters here. Mapping `sources` would also carry
+    // the tasks that produce generated source directories (eg. KSP) into the record task's outputs,
+    // which can't be resolved when the configuration cache entry is stored.
+    return providerFactory.provider {
+      val sourceSetRoot = sources.get().firstOrNull()?.asFile?.parentFile
+        ?: error("No source dirs registered for $testVariantName")
       projectDirectory.dir(sourceSetRoot.path).dir("snapshots")
     }
   }
--- a/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
+++ b/paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
@@ -390,6 +390,28 @@ class PaparazziPluginTest {
   }
 
   @Test
+  fun configurationCacheWorksWhenRecordingWithGeneratedTestSources() {
+    val fixtureRoot = File("src/test/projects/configuration-cache-generated-test-sources")
+    fixtureRoot.resolve("src/test/snapshots").registerForDeletionOnExit()
+
+    // Generated test sources (eg. from KSP) must not be resolved when storing the record task's
+    // outputs in the configuration cache. https://github.com/cashapp/paparazzi/issues/2374
+    gradleRunner
+      .withArguments("recordPaparazziDebug", "--configuration-cache", "--stacktrace")
+      .runFixture(fixtureRoot) { build() }
+  }
+
+  @Test
+  fun configurationCacheWorksWhenRecordingWithGeneratedTestSourcesInMultiplatform() {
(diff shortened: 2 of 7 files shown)


paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/PaparazziPlugin.kt
```diff
@@ -472,10 +472,14 @@ public class PaparazziPlugin @Inject constructor(
...
+    val testVariantName = testVariant.name
     val projectDirectory = layout.projectDirectory
-    return sources.map { dirs ->
-      val sourceSetRoot = dirs.firstOrNull()?.asFile?.parentFile
-        ?: error("No source dirs registered for ${testVariant.name}")
+    // Only the location of the source directories matters here. Mapping `sources` would also carry
+    // the tasks that produce generated source directories (eg. KSP) into the record task's outputs,
+    // which can't be resolved when the configuration cache entry is stored.
+    return providerFactory.provider {
+      val sourceSetRoot = sources.get().firstOrNull()?.asFile?.parentFile
```
colinmarsch: It looks like the existing `firstOrNull()` behaviour here can select the generated KMP source directory. I reproduced this by recording a snapshot from the new multiplatform fixture, which wrote the golden under build/generated/sourceGen/snapshots/images/ instead of src/androidHostTest/snapshots/images/. Since the generated path is disposable, could we derive the snapshot directory from a stable source root or exclude generated build directories?

Bortlesboat: Good catch, thanks. Fixed in c5c99a7: it now picks the first source directory outside the build directory, falling back to the first one. Both record fixtures now take a snapshot and check that the golden lands under `src/test/snapshots` / `src/androidHostTest/snapshots`. The multiplatform check fails without the change.

colinmarsch (approved):
This change looks good to me now! Just one small nit, do you think we could update the test to run the fixture twice and assert configuration-cache reuse?

I think we can release this with the alpha5.2 release.

Bortlesboat:
Added in 741309dca. Both record tests now run the task a second time and check that the configuration cache entry gets reused. The first run also checks that the entry was stored, so a discarded entry can't pass quietly.

PR #2299, opened 2026-03-05
Add paparazzi-junit4-reporting for Gradle test report image attachments

joshfriend:
Closes https://github.com/cashapp/paparazzi/issues/2263 

Introduces a new `paparazzi-junit4-reporting` artifact: a JUnit Platform engine wrapper (`paparazzi-vintage`) that intercepts `EngineExecutionListener` to call `fileEntryPublished()` for every snapshot PNG, surfacing images in Gradle 9.4+ test reports. The Paparazzi plugin auto-applies this new dependency and excludes the `junit-vintage` engine.

Issue #2182, opened 2025-12-04
Gradle internal API usage breaks Paparazzi on Gradle 9.3-rc-1


TWiStErRob:
**Description**

https://github.com/cashapp/paparazzi/blob/20f4936afc370b10cf08d689f0844441b0e606f5/paparazzi-gradle-plugin/src/main/java/app/cash/paparazzi/gradle/reporting/ClassTestResults.kt#L18
```
Caused by: java.lang.NoSuchMethodError: 'java.lang.String org.gradle.internal.FileUtils.toSafeFileName(java.lang.String)'
	at app.cash.paparazzi.gradle.reporting.ClassTestResults.<init>(ClassTestResults.kt:18)
	at app.cash.paparazzi.gradle.reporting.PackageTestResults.addClass(PackageTestResults.kt:38)
	at app.cash.paparazzi.gradle.reporting.AllTestResults.addTestClass(AllTestResults.kt:36)
	at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.loadModelFromProvider$lambda$1(PaparazziTestReporter.kt:46)
	at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.loadModelFromProvider$lambda$2(PaparazziTestReporter.kt:45)
	at org.gradle.api.internal.tasks.testing.report.generic.TestTreeModelResultsProvider.visitClasses(TestTreeModelResultsProvider.java:297)
	at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.loadModelFromProvider(PaparazziTestReporter.kt:45)
	at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.generateReport(PaparazziTestReporter.kt:37)
	at org.gradle.api.tasks.testing.AbstractTestTask$CustomTestReportingGenerator.lambda$generate$0(AbstractTestTask.java:794)
	at org.gradle.api.internal.tasks.testing.report.generic.TestTreeModelResultsProvider.useResultsFrom(TestTreeModelResultsProvider.java:72)
	at org.gradle.api.tasks.testing.AbstractTestTask$CustomTestReportingGenerator.generate(AbstractTestTask.java:792)
	at org.gradle.api.internal.tasks.testing.MultiTestReportGenerator.generate(MultiTestReportGenerator.java:46)
	at org.gradle.api.internal.tasks.testing.DefaultRootTestEventReporter.close(DefaultRootTestEventReporter.java:100)
[...]

Issue #2111, opened 2025-09-22
Gradle 9.1.0: PaparazziTestReporter - Could not generate test report


eygraber:
When a Paparazzi test fails on Gradle 9.1.0 the following exception is thrown:

```
org.gradle.api.GradleException: Could not generate test report to '/home/eli/workspace/app/screens/case-team/build/reports/tests/testDevDebugUnitTest'.
        at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.generateFiles(PaparazziTestReporter.kt:118)
        at app.cash.paparazzi.gradle.reporting.PaparazziTestReporter.generateReport(PaparazziTestReporter.kt:38)
        at org.gradle.api.tasks.testing.AbstractTestTask.createReporting(AbstractTestTask.java:617)
        at org.gradle.api.tasks.testing.AbstractTestTask.executeTests(AbstractTestTask.java:525)
        at org.gradle.api.tasks.testing.Test.executeTests(Test.java:714)
        at com.android.build.gradle.tasks.factory.AndroidUnitTest.executeTests(AndroidUnitTest.java:172)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:125)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:51)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:29)
        at org.gradle.api.internal.tasks.execution.TaskExecution$3.run(TaskExecution.java:252)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:166)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
[...]

Issue #1745, opened 2024-12-13
Gradle 8.12 internal API change, ClassNotFoundException: org.gradle.api.internal.tasks.testing.report.TestReporter


TWiStErRob:
**Description**
Gradle 8.12 cannot be used to build with Paparazzi.
See [Build Scan](https://scans.gradle.com/s/w4whl47rpgm42/failure#1) for exception.

[PaparazziPlugin.kt](https://github.com/TWiStErRob/net.twisterrob.sun/blob/main/gradle/plugins/src/main/kotlin/net/twisterrob/sun/plugins/PaparazziPlugin.kt)

**Steps to Reproduce**

See https://github.com/TWiStErRob/net.twisterrob.sun/pull/492

**Expected behavior**
Minor update, so just works.

**Additional information:**
[Toml](https://github.com/TWiStErRob/net.twisterrob.sun/blob/main/gradle/libs.versions.toml)
- Paparazzi Version: 1.3.5
- OS: any
- Compile SDK: N/A
- Gradle Version: 8.12-rc-1 (expected to stay the same in 8.12)
- Android Gradle Plugin Version: 8.7.3

TWiStErRob:
Confirmed fixed.
https://github.com/TWiStErRob/net.twisterrob.sun/pull/493 verified screenshots. Still failing, but that's due to Gradle deprecations.

PR #2436, opened 2026-09-14
Add reportType=native to surface diffs in Gradle's test report

### LINKED ISSUE #2263: Support attaching snapshot images to Gradle test results
https://docs.gradle.org/9.4.0-rc-1/release-notes.html#additional-test-data-capture

This new feature seems relevant and useful to paparazzi:

Image

jrodbx:
Introduces the `app.cash.paparazzi.reportType` Gradle property. It defaults to `legacy`, which is today's behavior unchanged: Paparazzi replaces Gradle's test reporter with its own and embeds diff images in the generated page.

Setting it to `native` leaves Gradle's report generator in place and instead publishes each snapshot diff as a FileEntry attachment on the test that produced it, so diffs appear in the attachments tab of Gradle's own report. This requires Gradle 9.4 or later, enforced at plugin-apply time.

Native mode routes JUnit 4 tests through a wrapping JUnit Platform engine, published as a new paparazzi-junit-platform artifact so that JUnit Platform stays out of the always-loaded paparazzi runtime. The two sides share no typed link: the engine discovers diffs by filename convention from paparazzi.failures.dir at executionFinished. The plugin adds the artifact to the test runtime classpath and excludes the real junit-vintage engine, which would otherwise discover the same tests a second time.

Not installing a custom TestReporter also keeps native mode clear of the Gradle internal reporting APIs that have repeatedly broken across releases (#1745, #2182, #2111).

Accessibility artifacts are not migrated yet. AccessibilityRenderExtension still bakes its overlay into the snapshot in both modes; publishing those as attachments is the follow-up that lets legacy be removed.

Obsoletes https://github.com/cashapp/paparazzi/pull/2299.  Closes https://github.com/cashapp/paparazzi/issues/2263.


geoff-powell (approved):

paparazzi-junit-platform/src/main/java/app/cash/paparazzi/junitplatform/PaparazziExecutionListener.kt
```diff
@@ -0,0 +1,105 @@
...
+    val prefix = "delta-${packageName}_${simpleClassName}_$methodName"
+
+    failureDir.listFiles()
+      ?.filter { file ->
+        val name = file.name
+        if (!name.startsWith(prefix) || !name.endsWith(".png")) return@filter false
+        // Ensure prefix is followed by '.' (no label) or '_' (label suffix) so
+        // a method like `testThing` doesn't accidentally match `testThingExtra`.
+        val boundary = name.getOrNull(prefix.length)
+        boundary == '.' || boundary == '_'
```
colinmarsch, line 91: Not sure I understand these boundary values fully? `_` is valid in test method names and `[` is used in parameterized tests but seem to be rejected?

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
```diff
@@ -794,6 +794,34 @@ class PaparazziPluginTest {
...
+
+    val deltaName = "delta-app.cash.paparazzi.plugin.test_VerifyTest_verify.png"
+    val delta = File(fixtureRoot, "build/paparazzi/failures/debug/$deltaName")
+    assertThat(delta.exists()).isTrue()
+
+    // Assert against the rendered HTML, not the binary event log: that log also contains
+    // the failure message's copy of the delta path, so matching it would pass either way.
+    val testReportDir = File(fixtureRoot, "build/reports/tests/testDebugUnitTest")
+    val attachmentReferenced = testReportDir.walkTopDown()
+      .any { it.isFile && it.extension == "html" && it.readText().contains(deltaName) }
```
colinmarsch, line 821: Should we check more specifically here that this is in the attachments or is an ``? As is this could pass if the filename appears in failure text.

colinmarsch (approved):
This looks good to me! Just left a couple comments

paparazzi-junit-platform/src/main/java/app/cash/paparazzi/junitplatform/PaparazziExecutionListener.kt
jrodbx: Good catch on `[` (confirmed with a test). 

Parameterized writes `delta-…_verify[0].png`, but Vintage's `MethodSource` reports the bare verify. 

Before: delta on disk, no attachment. 
After: it's in the attachments tab.

See test `verifyFailureNativeReportParameterized`.

Re: `_`, I kept as-is.  `…_verify_baseline_accessibility.png` is identical for method `verify baseline accessibility` and for `verify baseline` with label accessibility, since `toFileName` uses `_` for both. Over-matching beats silently dropping an image; something to fix in the future.

paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
jrodbx: Fixed.  Now matching `<img src="…delta-….png"`, which only the attachments markup produces.

jrodbx:
odd, this auto-merged despite a failed Windows CI build and I was relying on that to add/fix some stuff.  fast follow incoming.

colinmarsch:
> odd, this auto-merged despite a failed Windows CI build and I was relying on that to add/fix some stuff. fast follow incoming.

Ah, I saw that failure and it looked like a dep download flake so I re-ran it and I guess it passed.

PR #2450, opened 2026-09-16
Support JUnit 5 in native mode, via per-framework artifacts

jrodbx:
Native mode now attaches snapshot diffs for JUnit 5 tests as well as JUnit 4, by adding a Jupiter engine wrapper alongside the Vintage one.

The wrappers ship as separate artifacts because engine registration is per jar. ServiceLoader instantiates every registered engine and EngineFilter only applies afterwards, so one artifact registering both would classload both engines no matter what is excluded. `paparazzi-junit-platform` now holds the shared listener and the one function that rewires an ExecutionRequest; `paparazzi-junit-vintage` and `paparazzi-junit-jupiter` each hold one engine with its own registration and engine dependency. A JUnit 4 project never resolves the Jupiter engine.

That rewire now forwards the whole ExecutionRequest. The 3-arg factory substitutes a disabled CancellationToken and an absent store, which Vintage tolerates only because Gradle never arms the token, and which Jupiter rejects outright.

`app.cash.paparazzi.nativeReportFrameworks` picks the wrappers, defaulting to `junit4` and accepting a comma-separated set so one test task can run both. Declaring only one framework leaves the other engine unexcluded, so those tests still run, just without attachments.

Each wrapper also supplies its own key for matching a diff to a test, which resolves the boundary-character question from #2436. Vintage uses the descriptor's display name, which is the JUnit 4 Description's method name and already carries Parameterized's suffix, so `[` no longer has to be accepted as a prefix boundary. Jupiter keeps MethodSource, because its display names are author-controlled via `@DisplayName`. A method name containing `_` is still indistinguishable from a labelled snapshot of a shorter name; that one is inherent to `Snapshot.toFileName`.

Attachment assertions moved into one helper that checks per report page: exactly one page carries the ``, and that page names the owning test. Presence somewhere in the report no longer counts. Fixtures cover plain JUnit 4, parameterized JUnit 4, Jupiter, and both frameworks in a single test task.


paparazzi-junit-platform/src/main/java/app/cash/paparazzi/junitplatform/PaparazziExecutionListener.kt
```diff
@@ -77,18 +78,17 @@ internal class PaparazziExecutionListener(
...
 
     failureDir.listFiles()
       ?.filter { file ->
         val name = file.name
         if (!name.startsWith(prefix) || !name.endsWith(".png")) return@filter false
-        // '.' unlabelled, '_' before a label, '[' before a parameterized suffix that
-        // MethodSource omits. A '_' in the method name reads as a label, so it over-matches.
+        // '.' when unlabelled, '_' before a label; a '_' inside the method name over-matches.
         val boundary = name.getOrNull(prefix.length)
-        boundary == '.' || boundary == '_' || boundary == '['
```
colinmarsch, line 91: The `[` boundary isn't needed?

colinmarsch (approved):
Looks good!

paparazzi-junit-platform/src/main/java/app/cash/paparazzi/junitplatform/PaparazziExecutionListener.kt
jrodbx: Correct!  It's not needed anymore now that each engine supplies its own key.

Before, both went through `MethodSource`, which reports the bare `verify` while `Parameterized` writes `verify[0].png`, so `[` had to count as a boundary.  Now, Vintage uses the descriptor's display name, which already has the `[0]`, so the boundary is `.` again. `verifyFailureNativeReportParameterized` still covers that case.

Meanwhile, Jupiter stays on `MethodSource`, since Jupiter display names can be renamed with `@DisplayName`. It's also the name the fixtures hand to `TestName`, so a `@ParameterizedTest` only gets a bracket in the filename if the author puts one there.

`_` is still there for the labelled snapshot case from last time.

PR #2452, opened 2026-09-18
Deliver path system properties from doFirst, not an argument provider

jrodbx:
#2381 kept five absolute paths out of the test task's cache key with a CommandLineArgumentProvider and an @Internal map. Four other path properties here already do it by being set from doFirst, after inputs are snapshotted, so use that for all nine.

The path strings never needed tracking: resources.json is covered by inputs.file, the report dir by outputs.dir, and the project and artifact-cache roots by the path-sensitive resource and asset dir inputs. #1874 stays fixed, covered by cacheableRelocatable.


colinmarsch (approved):
This change looks good to me for consistency!

PR #2456, opened 2026-09-21
Wire clean-record to the variant's own snapshot deleter

jrodbx:
#2377 ordered `test<Variant>UnitTest` after `delete<Variant>PaparazziSnapshots`, and explained why the older constraint did nothing: the `mustRunAfter` between `recordPaparazzi<Variant>` and `deletePaparazziSnapshots` orders two action-less tasks, so it never reaches their dependencies.

That dead constraint stayed. So did `cleanRecordPaparazzi<Variant>`'s dependency on the aggregate `deletePaparazziSnapshots`, which pulls in every variant's deleter. They all clear the same `src/test/snapshots`, so with unit tests on more than one build type, another variant's deleter can run after this variant's test task and erase what it just recorded.

Instead, let's depend on `delete<Variant>PaparazziSnapshots`, drop the dead constraint, and move the ordering into the `configureEach` block that already wires the test task.


paparazzi-gradle-plugin/src/test/java/app/cash/paparazzi/gradle/PaparazziPluginTest.kt
colinmarsch, line 1063: Can we add a multi variant cleanRecord test for more coverage here?

geoff-powell:
hot take / question: is there ever a future where buildFlavors generate screenshots in different folders? Currently they don't and this is a limitation of the plugin. Different variants overwrite the same snapshots

geoff-powell (approved):

geoff-powell:
This is a good callout as I ran into this the other day and have a note to log an issue for. Thanks for tackling!!

